PR #36, opened 2015-08-13
Updated getLog

raynichc:
getLog can now search for ip and endDate now works properly.


rossdotparker:
Looks great, thanks Ray.

PR #45, opened 2015-10-27
Fixed visual bug in second column.

raynichc:
Fixed a bug where "Giver" and "Role" appeared on the same line.


rossdotparker:
Well caught, thanks for the contribution!

PR #46, opened 2015-12-03
Patch 11

raynichc:


rossdotparker:
Thanks for this Ray. Looks good, but needs to be committed to the v11 branch, not Master. Can you recommit to v11, and then remove this commit? Cheers : )

rossdotparker:
Actually, no need to remove, as I just closed it. Just recommit to v11.

PR #51, opened 2016-03-17
Made the Student Name in the Minor Links a link.

raynichc:


rossdotparker:
Ray,

I like this solution, and am grateful you took time to code and push it, but I am not sure it is a problem entirely worth solving. It DOES make the minor links look more consistent, but if you click on your name, you only see things that you already know. It also adds processor and database overhead to page loads, and complexity (albeit only a little) to the code base. What do you think?

Ross

raynichc:
Hmm I didn't think about that, but I don't think it's too much of a problem. I don't think this will slow down or overload the system by too much and the amount it does is probably negligible when considering the rest of the content on the page.

rossdotparker:
OK, you convinced me : )

PR #52, opened 2016-03-21
Fixed Password Fields

raynichc:
The installer form will no longer allow you to proceed if the passwords do not match.


rossdotparker:
Thanks for this Ray. Did not need this push request in the end, as I had committed the code you emailed to me already.

PR #53, opened 2016-04-12
System: added Macanese Pataca as currency

SKuipers:
Added MOP$  https://en.wikipedia.org/wiki/Macanese_pataca


rossdotparker:
Love the fact that you even updated the change log. That's the kind of pull request I like : ) Thanks!

PR #55, opened 2016-04-16
Install.php STRICT MODE

folesen:
Changed error message to inform the user whether they are in strict
mode or not.


rossdotparker:
Nope, nope, nope!

PR #62, opened 2016-04-21
Completed Activity Attendance feature

SKuipers:
Finished up the activity attendance from last Saturday. Added a new action 'Enter Activity Attendance'. Taking attendance works on a checkbox grid, with some added javascript to make it more intuitive, and should also be easy to bulk-add attendance information. Added an Export to Excel, which makes a decent printable version of the data as well.

I've added a double scrollbar to make the grid easier to view when there's a large number of rows & columns, with the student list pinned on the left column. I'm also testing this feature locally for Markbook improvements, to make viewing columns easier (less clicking newer/older). To test: create an activity with time slots, enroll students, and set a fairly long date or term range.

The original 'Attendance by Activity' report was moved to 'Printable Attendance Sheet' with the ability to select how many columns. It also adds printer page-breaks for large lists of students.

The new 'Attendance by Activity' report is a view-only version of the 'Enter Activity Attendance' action.


SKuipers:
Added a quick fix because returnProcess had changed since I last looked at it ;) Also tested a bunch more locally after merging.

rossdotparker:
OK, all looks great. Love the interface style :+1: I have made a couple of tweaks to remove some minor issues to do with warnings and <?, which you can see in these two commits: https://github.com/GibbonEdu/core/commit/b78d58044971b5f5b556dd9cbc7b09fba17af963 and https://github.com/GibbonEdu/core/commit/22dd61f0c0241e7a9622df0a099bba4d89e13d37. Thanks very much for your work on this!

SKuipers:
Thanks!

PR #66, opened 2016-05-13
Markbook v12

SKuipers:
Hi Ross,

Here's the markbook update (finally!). There's some visual improvements to the markbook display. There's also options to add weightings and calculations to the grades, which won't be useful for your school, but brings the system in line with what other grading systems can do and (unfortunately) what our school needs.

I've tested it a bunch and I'll also enlist our tech coaches and early adopters to help thoroughly test all the markbook changes. I've added Markbook Settings, defaulted to off, for each of the added features so that school's won't see a major change in the markbook unless they enable the settings. 

Changelog:

- Added support for raw attainment values which can automatically calculate the actual value if the grade scale is percent based.
- Added configurable weightings for markbook types, along with cumulative and final grade calculation
- Added mark book column filters and functionality to view and calculate marks by term
- Improved the display, order and pagination of markbook columns

Other stuff:
- The manage weightings page lets teachers set a weighted value for certain markbook types, and determine when the weighting should be applied: cumulatively (eg: quizzes 30%, assignments: 20%, projects: 50%), or just only the final grade (such as final exams, final projects).
- Added a date field to markbook columns (separate from the go-live date) to help determine which term they belong to, and for column ordering.
- Added drag-and-drop column ordering. Maybe a little frivolous but it looks neat ;)
- Added extra columns to display cumulative and final averages
- Session variables keep the same class & filters selected as teachers browse through the system
- Double-scroll helps with longer lists. Auto-scrolls to the right where the newest columns are.

The markbook view is organized into two main classes that handle a lot of the data and calculations. There's comments and stuff but be sure to give a shout if I can help explain or clarify anything.

The current v12 branch has been merged in so it should be all up-to-date.

Thanks!!


rossdotparker:
Sandra, going to commit this, as all looks good.

One thing, you have gone in a change lines 356 and 359 in CHANGEDB.php, but these have already been run on some installations. I will leave the changes there, but can you add matching ALTER TABLE calls to the end of v12 in that file to fix the issue for sites who have already passed 356 in the change record? These fields are actually somewhat redundent in core modules, but best to keep all installs the same.

My general rule is not to change any PHP statments in CHANGEDB.php once that file has been committed into the branch, as you never know who has already run in. Thanks!

PR #67, opened 2016-05-13
V12.0.00 export from Page

crayner:
Excel export now works through the class in all circumstances, but some areas not tested due to lack of data.


rossdotparker:
I notice that this request includes the themes Darkness, Time Forest and Seafarer. Could you remove those and pull request again? I will shut this first request. Thanks!

PR #77, opened 2016-06-21
Markbook: Updated use of empty to improve backwards compatibility

SKuipers:
Updated a few instances of the empty() function to conform to pre-PHP 5.5

>     Note:
>     Prior to PHP 5.5, empty() only supports variables; anything else will result in a parse error. In other words, the following will not work: empty(trim($name)). Instead, use trim($name) == false.

from http://us3.php.net/empty


rossdotparker:
With the requirements bumped up to PHP 5.5, we will not need this.

PR #82, opened 2016-08-22
Master

crayner:
update of gitignore ONLY


rossdotparker:
Craig, we are currently not merging into the Master branch, as this will happen only when v13 is complete, and we approach release. Do you want to push these changes to the v13 branch instead?

crayner:
Yes, this was a version 12 issue...   No problem

PR #84, opened 2016-09-09
tracking Settings Bug

crayner:
Warning: Invalid argument supplied for foreach() in /home/soft/public_html/Gibbon/modules/School Admin/trackingSettings.php on line 89


crayner:
This bug is showing on the SOftaculous Demo.   This  should fix the issue, but I have not tested.  v13-Backwrads does NOT have this biug.

PR #85, opened 2016-09-10
Composer/Vendor Preparation

crayner:
This pull generates the vendor directory for v13.  It also adds composer to the root directory ready to autopopulate the vendor directory with packages.
The composer program generates a standard autoload program, so some minor changes to the core classes to repoint to the new autoloader. 
Removed Gibbon autoloader


crayner:
Hi Ross, It does not matter as the new version of this file will overwrite completely this solution..

PR #91, opened 2016-09-19
Bootstrap Mark 1 Issue 2

crayner:
Menu/finder bug fix
Chrome Anchor bug fix
Login script repair when called from page needing authentication but user not authenticated.
More migrations of System Admin.
Password Reset script added and table added but not yet in use, but linked in files used for Security Login, so checks made.


crayner:
OK,  Close this so you can do a whole Module.  System Admin Completely Migraed

PR #93, opened 2016-09-21
Debug. Minor changes only

crayner:
Bigger Notification ICONS
Module Category - Remove ENUM (no sort)
System Setting - Rename Primary to Default 
System Setting, Third Party - Remove Submit All
Botton of Page Anchor

Scripts can now be pushed to bottom of page, as perceived speed increase.


rossdotparker:
Craig, I have committed a couple of changes, and so your CHANGELOG.php is a little out of sync. Can you use the most recent, and then include ALTER TABLE `gibbonModule` CHANGE `category` `category` VARCHAR(10) CHARACTER SET utf8 COLLATE utf8_unicode_ci NOT NULL;end on line 512. Otherwise, some systems might miss part of the upgrade. Other than that, it looks good.

crayner:
Closed to allow another pull

PR #115, opened 2016-10-04
Debug Like

crayner:
last pull was 6be714d686ba19e59b047ff5ebe00b1224d19d7a
4th Oct 2016
Debug: Like Count


crayner:
Still another mistake

PR #126, opened 2016-10-27
Planner: Hide attendance reasons until dropdown changed

SKuipers:


SKuipers:
Updated to fix Planner student attendance pulling in most recent status

PR #168, opened 2017-02-08
Fixed Croatian, Ukrainian language names, added Myanmar language and code

VicMyFriend:
Should I include a check for 'value already present' before inserting new values in the table, to avoid duplicates?


rossdotparker:
Usually I don't check for existing records, but it might not be a bad idea to do so. Do you want to check this forum page and see if you can get it to work the way you want? http://stackoverflow.com/questions/3164505/mysql-insert-record-if-not-exists-in-table

SKuipers:
Can be closed, changes included in https://github.com/GibbonEdu/core/pull/199

PR #171, opened 2017-02-17
Form Class OO improvements

SKuipers:
Form Class changes:

Improved extensibility, particularly with module developers and the future core in mind:
- The Form class has two replaceable components: FormFactory and FormRenderer
- FormFactory handles all object creation methods (eg: addTextField). Module developers will be able to add their own objects easily by extending the existing factory or providing a new one.
- FormRenderer handles turning all Form, Rows and Elements into HTML. Module developers can provide a new renderer to change how the form is displayed without needing to edit or change how the rest of the Form is built.
- The static Form::create function automatically creates a default FormFactory and FormRenderer for each Form. This is the preferred method to create new forms throughout Gibbon, rather than calling new Form.
- Applied some SOLID principals to keep things modern & future-friendly: Class dependencies are based in Interfaces wherever possible, dependencies are injected (no globals), classes have a single responsibility, etc.

Improved file structure:
- Moved namespace from Library/Forms to Gibbon/Forms
- Organized classes into logical sub-namespaces

Interfaces:
- OutputtableInterface: Defines any piece of the form that is ultimately called by the Renderer to be output. Anything that implements OutputtableInterface can be added to a row, making it easier to add custom content to forms.
- ValidatableInterface: Defines any piece of the form that has validation (at this point, only Inputs)
- FormRendererInterface & FormFactoryInterface: Defines essential methods for developers creating a new renderer/factory

Traits:
- BasicAttributesTrait: Common getter/setter methods for basic HTML attributes: class and id, used in components that are translated into HTML
- InputAttributesTrait: Common getter/setter methods for input attributes: name, type, value, required
- MultipleOptionsTrait: Common methods for reading in multiple options from a variety of sources, used in checkbox groups, radio, select


modules/School Admin/activitySettings.php
```diff
@@ -33,234 +35,64 @@
...
-						<option <?php if ($row['value'] == 'Term') { echo 'selected '; } ?>value="Term"><?php echo __($guid, 'Term') ?></option>
-					</select>
-				</td>
-			</tr>
-			<script type="text/javascript">
-				$(document).ready(function(){
-					<?php if ($row['value'] == 'Date') { ?>
-						$("#maxPerTermRow").css("display","none");
-					<?php } ?>
+    $form = Form::create('financeSettings', $_SESSION[$guid]['absoluteURL'].'/modules/'.$_SESSION[$guid]['module'].'/activitySettingsProcess.php');
```
rossdotparker: This should be activitySettings, right?

SKuipers: Yes it should be.

PR #179, opened 2017-02-23
Fixed Student History css and Markbook column sequence number

SKuipers:


SKuipers:
[Markbook: New columns getting added in the middle of the set.](https://trello.com/c/EzYzkcUh/163-markbook-new-columns-getting-added-in-the-middle-of-the-set)

PR #187, opened 2017-03-02
Students: Fixed file uploads failing in applicationFormProcess

SKuipers:
If cookies are disabled in a user's browser the session variables like $_SESSION[$guid]['absolutePath'] are missing so the file uploads were silently failing.

This fix does two things: loads the system settings into $_SESSION if they don't exist, and does one more attempt at uploading using the base Gibbon path if the first attempt fails.


rossdotparker:
Fixed and tested. Looks good. I guess this won't be an issue on most pages, as you can't login without cookies.

PR #193, opened 2017-03-21
Staff: Fixed application uploads when applying to multiple posititions

SKuipers:
Looks like applicationFormProcess.php was moving the temp uploaded files for each position in the loop, which caused any references to those files after the first loop to fail. This change moves the processing of uploads to before the loop for each position, and collects an array of files that are then attached to each position once inside the loop.

### DIFF
--- a/modules/Staff/applicationFormProcess.php
+++ b/modules/Staff/applicationFormProcess.php
@@ -218,6 +218,49 @@ if ($proceed == false) {
             $partialFail = false;
             $ids = '';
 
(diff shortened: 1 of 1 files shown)


VicMyFriend:
Just tested - issue fixed! Uploaded files are properly associated with several Vacancies the candidate applied for and "Download" links open the proper files on the server.
THANK YOU!

SKuipers:
Awesome, glad to hear it's working. Thanks for helping test :)

PR #197, opened 2017-03-22
Added new "Enter Activity Attendance" sub-action

raynichc:
The new action allows attendance to be taken for activities by the Organiser or Assistants of that activity.


modules/Activities/activities_attendance.php
```diff
@@ -41,6 +41,8 @@
     echo __($guid, 'Choose Activity');
     echo '</h2>';
 
+    $highestAction = getHighestGroupedAction($guid, '/modules/Activities/activities_attendance.php', $connection2);
+    print $highestAction;
```
SKuipers: Great looking commit! Looks like you may have left a debugging output in here ;) If you push any new changes to your branch it'll automatically update the PR

raynichc: Whoops you right. I'll fix that now. Thanks!

modules/Activities/activities_attendanceProcess.php
```diff
@@ -45,6 +45,31 @@
...
+            $dataCheck = array("gibbonPersonID" => $gibbonPersonID, "gibbonActivityID" => $gibbonActivityID);
+            $sqlCheck = "SELECT role FROM gibbonActivityStaff WHERE gibbonActivityID=:gibbonActivityID AND gibbonPersonID=:gibbonPersonID";
+            $resultCheck = $connection2->prepare($sqlCheck);
+            $resultCheck->execute($dataCheck);
+
+            if ($resultCheck->rowCount() > 0) {
+                $row = $resultCheck->fetch();
+                if ($row["role"] != "Organiser" && $row["role"] != "Assistant") {
+                    $URL .= '&return=error0';
+                    header("Location: {$URL}");
```
rossdotparker: I'm pretty sure you need exit(); after this header, and the following two headers, otherwise your code will run on beyond the if statement.

raynichc: I've added the exits and checked the My Activity Attendance button. The check uses isActionAccessible instead of highestAction, so no problems.

PR #202, opened 2017-03-30
Attendance: Fixed attendance trends when counting end of day status

SKuipers:
Fixed some fun and interesting interactions with aggregated columns, apparently they will return arbitrary values for a row on non-grouped columns.

Changed the column to NULL as discussed, and added a non-unique index to help optimize the attendance log lookups (since its one of the biggest tables in the database and the graph query uses an inner join, without an index it can take several seconds to query). Its not a unique index so won't cause any errors when added.


CHANGEDB.php
```diff
@@ -739,4 +739,6 @@
 INSERT INTO `gibbonPermission` (`permissionID` ,`gibbonRoleID` ,`gibbonActionID`) VALUES (NULL , '1', (SELECT gibbonActionID FROM gibbonAction JOIN gibbonModule ON (gibbonAction.gibbonModuleID=gibbonModule.gibbonModuleID) WHERE gibbonModule.name='System Admin' AND gibbonAction.name='Notification Settings'));end
 INSERT INTO `gibbonNotificationEvent` (`event`, `moduleName`, `actionName`, `type`, `scopes`, `active`) VALUES ('Daily Behaviour Summary', 'Behaviour', 'Find Behaviour Patterns', 'CLI', 'All', 'Y');end
 INSERT INTO `gibbonNotificationEvent` (`event`, `moduleName`, `actionName`, `type`, `scopes`, `active`) VALUES ('New Negative Record', 'Behaviour', 'View Behaviour Records_all', 'Core', 'All,gibbonPersonIDStudent,gibbonYearGroupID', 'Y');end
+ALTER TABLE `gibbonAttendanceLogPerson` CHANGE `gibbonAttendanceCodeID` `gibbonAttendanceCodeID` INT(3) UNSIGNED ZEROFILL NULL;end
```
SKuipers: Will do. It'll only affect logs prior to the addition of the gibbonAttendanceCodes table, the field will never be blank for any new records. In the end I guess it could save a few KB for older data sets.

SKuipers:
Updated CHANGEDB. Thanks!

PR #207, opened 2017-04-03
Install themes - correct_instrucions

leofreitas:


rossdotparker:
Wow, how had we not spotted this one before. Thanks for the fix! Going to Merge it now.

PR #224, opened 2017-05-02
System: Added notification events for Activity registration/withdraw

SKuipers:
Implemented notification events for activities. 

- The events are raised from the Student/Parent registration process: a staff member adding or removing enrolment won't send a notification (does this make sense? I think that was the aim of the initial forum request.) 
- Sent to the activity organizer(s) by default, plus anyone subscribed to the notification event
- They currently trigger if the activity is running (within it's program/term start and end dates), which seems like the most notification-worthy timeframe rather than spamming on every registration.

Feature request: https://ask.gibbonedu.org/discussion/comment/1751#Comment_1751


SKuipers:
[Ability to set an activities coordinator (for whole school), who is notified on certain events.](https://trello.com/c/OXa3g3Kr/9-ability-to-set-an-activities-coordinator-for-whole-school-who-is-notified-on-certain-events)

SKuipers:
[Alert activitiy leader (for an activity) when a new student joins, once activity starts](https://trello.com/c/XLluzON8/8-alert-activitiy-leader-for-an-activity-when-a-new-student-joins-once-activity-starts)

PR #238, opened 2017-05-20
System: Added ability to set mysql port through config.php

SKuipers:
Allows users to add $databasePort to their config file to specify a MySQL port number. Due to a PDO limitation, the port is only recognized when using `127.0.0.1` instead of localhost, or when using a specific hostname.


SKuipers:
[Ability to specify mysql port for connections](https://trello.com/c/tmdaKFPf/5-ability-to-specify-mysql-port-for-connections)

PR #245, opened 2017-05-24
Attendance: Updated date used to check enrolment for Take Attendance

SKuipers:
The take attendance pages currently check the enrolment for a student based on their start and end dates using the current date. This can cause the class lists to appear incorrect when taking attendance for a date in the past when a student has joined the class after that day. This PR updates the roll group/class queries to use the date entered rather than current date, which reflects the enrolment at that point in time.

### DIFF
--- a/modules/Attendance/attendance_take_byCourseClass.php
+++ b/modules/Attendance/attendance_take_byCourseClass.php
@@ -276,14 +276,14 @@ else {
 
 					//Show roll group grid
 					try {
-						$dataCourseClass=array("gibbonCourseClassID"=>$gibbonCourseClassID, 'date' => $currentDate, 'today' => date('Y-m-d') );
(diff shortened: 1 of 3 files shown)


VicMyFriend:
Thanks for fixing these! Class Attendance has been improved with Exceptions taken into account. But, we have a strange issue when with one specific Class not showing any students on Mondays, although they should be shown. I'm trying to fish out a problem by testing the //Show roll group grid query, no luck so far. What would be the best way to tackle it?

SKuipers:
Hmm, that does sound odd. Do you timetable each day separately, perhaps one class is missing a TT day-row, or the exceptions for that TT day-row were an accidental 'select all' (it happens). Feel free to start a forum thread if you don't figure it out 😄

PR #256, opened 2017-06-01
System: Potential jquery.chained.js fix

SKuipers:
I haven't been able to reproduce the 'too much recursion' error yet, but I've cut the each() call from the function and simplified it even more. I tested it with a mock set of selects and 10000 options to filter, it worked without any errors (albeit a little slowly).

Hopefully this does the trick, otherwise something really strange is going on :thinking:


SKuipers:
Tested in FF, Chrome, Safari Desktop & iOS (wasn't able to reproduce the original error though). The only other issuse that comes to mind could be the minification of the script itself, so if this fix doesn't work, try deleting `jquery.chained.min.js` and renaming `jquery.chained.js` to `jquery.chained.min.js` (then clear cache). That will run the un-minified version, and at the very least the error messages should be more helpful.

PR #264, opened 2017-06-07
Tests: Added more acceptance tests in School Admin

SKuipers:
Some tests for pages that didn't get OOified yet, but its probably good to get them into dev branch anyway (I assume we'll likely strip the all tests out for the release?)

This brings the test count up to 50 for v14, all tests passing with the old and new demo data.


rossdotparker:
Interesting thought to remove the tests before release. Have to think about that one! They are not large, so might not be a bad idea to keep them in...others might spot them and offer to contributes.

PR #270, opened 2017-06-11
System: Fixes for PHP notices

SKuipers:
PHP notice cleanup during release prep, a couple notable ones:
- Found and removed an odd form element from Activities Manage Enrolment (likely a copy-paste artifact, there's no input or submit button for that section).
- Re-connected a link to Add Medical Form from Student Profile (looks like medical forms may have been managed in User Admin at some point and then moved?)


rossdotparker:
Some nice catches here : ) Do you want to add relevant CHANGELOG.txt entries if the issue did not get created earlier in v14? I think these all pre-date v14, and so should go into the log for this version. Thanks again!

PR #284, opened 2017-06-20
Students: Updated required documents to read-only in Edit Application

SKuipers:
This PR restores the read-only behavior of documents in Edit Application Form. It will handle displaying only the most recent file for the odd instance with two uploads. Internal documents were added in 14, so it applies to them as well. 

I do wonder if this could cause some headaches for admissions, if they accidentally upload to the wrong type or need to change a file to a newer version? :thinking:


SKuipers:
Updated to handle file replacement (an in-between solution that fixes the initial issue without preventing additional uploads):
- Displays a link to the most recently uploaded attachment (if one exists)
- Allows uploading a new attachment to replace the existing one. Added description text 'Will overwrite existing attachment.' to indicate this and match other file-upload areas.
- Does a uniqueness in editProcess check to avoid inserting multiple files (for now, until the UI and processing can be added for multiple files per type.)

PR #285, opened 2017-06-21
Students: Fixed School History in profile to exclude upcoming years

SKuipers:
This one caused an unexpected surprise when the enrolment for upcoming years was visible even though the user role did not have access to future years 😧

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -32,6 +32,7 @@ v15.0.00
 
 	Tweaks & Bug Fixes
 		i18n: fixed erroneous Georgian language code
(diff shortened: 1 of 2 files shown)


rossdotparker:
Well caught! Never considered this little leak before. Ross.

SKuipers:
Yep, fairly obscure, though little was not among my choice of words for it
yesterday 😅

On 21 Jun 2017, at 9:35 PM, Ross Parker <notifications@github.com> wrote:

Well caught! Never considered this little leak before. Ross.

—
<https://github.com/GibbonEdu/core/pull/285#issuecomment-310080203>, or mute
the thread

Issue #291, opened 2017-08-05
Application Form submit error when not select family name by logged in user with family.


hanattaw:
The error shown "Your request failed because your inputs were invalid."

I've fixed this by manually select family name.

<img width="766" alt="screen shot 2560-08-05 at 14 44 47" src="https://user-images.githubusercontent.com/3141235/28993961-ab15048c-79ec-11e7-8914-dd06bda4df2a.png">

After searching in the code i see the "addRadio->checked($checked)" the checked method need to give the value but not not boolean.

SKuipers:
Thanks! Good catch, just PR'd a fix. Sounds like you're familiar with PHP, if you'd like to pull request any fixes for bugs you find feel free to give a shout with any dev questions 😃

rossdotparker:
@hanattaw good catch indeed, and thanks for pointing out the solution. Let me know if you want to be added to the team here on GitHub so you can put in pull requests yourself. And, you are welcome to join us on Slack (just let me know and I will invite you) where we do our dev chat.

This issue has been merged into v15 dev branch in commit https://github.com/GibbonEdu/core/commit/a7cd7f03ec097b32d4ecb1f1b1b364c0cb3b7ccc. Feel free to test and feedback. In the meantime I will close this ticket.

hanattaw:
@rossdotparker Thank you. Please add me to the team.

rossdotparker:
Done!

PR #292, opened 2017-08-06
Students: fixed default selected gibbonFamilyID in Application Form

SKuipers:
For issue #291. This fix updates the Student Application Form to always select the first gibbonFamilyID when one or more families are available to select (ensuring the value is never null).

This one must have snuck past us in pre-release testing (although I do recall testing the logged-in application form a few times). Currently there isn't an automated test for the logged-in family-exists variant of the form so I've made a note to write one.

We had previously discovered that LiveValidation doesn't work for Radio elements (because LV uses IDs and they have to be unique). This means currently there's no handling for required radio elements. Do you think it would make sense to update the Radio class to always select a default value if its a required field? It makes it possible for a user to accidentally leave the default selected, but that's likely better than submitting a form with nothing selected, as currently there isn't a way to flag it during LiveValidation 🤔

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -40,6 +40,7 @@ v15.0.00
 		Finance: set name of outgoing email to school name, not user's name
 		Library: added Optical Media as item type
 		Students: fixed School History section in Student Profile to exclude upcoming years
+		Students: fixed Application Form for logged in users not selecting a family by default (if one exists)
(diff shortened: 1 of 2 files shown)


rossdotparker:
Sandra, thanks for this. Just going to review it. In the meantime, yes, I think you approach to select a default value when required is a good hack, given the LV situation.

PR #300, opened 2017-08-16
School Admin: updated Manage School Years to prevent removing or unsetting the current School Year

SKuipers:
It's possible to accidentally set the current year to Past without setting a new Current year, which causes all further login attempts to fail. This PR aims to enforce the presence of a single Current year with the following changes:

- Current school year cannot be deleted or it's status changed
- Changing another year's status to Current will replace the current school year, and warn the user making the change:
screen shot 2017-08-16 at 3 22 22 pm
- Changes to the current year updates session variables, which enables the top-of-page warning if they're now logged into a non-current year
- Clarified the error message when there's no current year, on the off chance it's changed through the database or some other means

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -46,6 +46,7 @@ v15.0.00
 		Individual Needs: added ability to record educational assistants for a student
 		Library: added Optical Media as item type
 		Planner: updated attendance to only hide the reason and comment fields if Present is the default attendance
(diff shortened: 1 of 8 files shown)


SKuipers:
sparta

SKuipers:
Applied all the above logic to Add School Year as well (forgot it initially).

PR #311, opened 2017-08-21
User Admin: added ability to set username formats by role

SKuipers:
From the development roadmap: Namespaces for usernames
https://ask.gibbonedu.org/discussion/comment/1718#latest

- Adds a section to User Settings to define username formats by Role
    - Copies over the existing formats from settings into a table
    - Handles different formats for different roles, with a default for any roles undefined
    - Handles numeric usernames that increase each time they're generated
- Adds a 'Generate Username' button to Add User, which uses the filled-in data and primary role to generate a unique username
- Adds the username uniqueness check from public registration to the Add User page
- Updates the student and staff application forms to use the new functionality
- Updates the public registration check to use the new functionality


SKuipers:
[Namespaces for usernames](https://trello.com/c/xh7Ym7uS/41-namespaces-for-usernames)

PR #370, opened 2017-10-21
Fix column objects to support multiple buttons

jimbo8098:
Buttons did not identify properly before with IDs, this is now fixed but beware that addButton now requires another value at construction to supply the ID. 

When the column was running getOutput(), the ID was picked but the ID was a string value "NULL" and therefore the ID matches on all buttons inside that column object. This fix uses the ID assigned to the button instead.

### DIFF
--- a/CHANGEDB.php
+++ b/CHANGEDB.php
@@ -849,5 +849,5 @@ ALTER TABLE `gibbonRubricEntry` ADD INDEX(`contextDBTable`);end
 ALTER TABLE `gibbonRubricEntry` ADD INDEX(`contextDBTableID`);end
 ALTER TABLE `gibbonRubricRow` ADD INDEX(`gibbonRubricID`);end
 INSERT INTO `gibbonNotificationEvent` (`event`, `moduleName`, `actionName`, `type`, `scopes`, `active`) VALUES ('Weekly Attendance Summary ', 'Attendance', 'Attendance Summary by Date', 'CLI', 'All,gibbonYearGroupID', 'Y');end
-
+ALTER TABLE `gibbonActivityStaff` MODIFY `gibbonPersonID` INT(10) UNSIGNED ZEROFILL NOT NULL DEFAULT '0000000000';end
 ";
--- a/src/Forms/FormFactory.php
+++ b/src/Forms/FormFactory.php
@@ -164,9 +164,9 @@ class FormFactory implements FormFactoryInterface
         return new Input\MultiSelect($this, $name);
     }
 
-    public function createButton($label = 'Button', $onClick = '')
+    public function createButton($name, $label = 'Button', $onClick = '')
     {
-        return new Input\Button($label, $onClick);
+        return new Input\Button($name, $label, $onClick);
     }
 
     public function createCustomBlocks($name, OutputableInterface $block, \Gibbon\Session $session)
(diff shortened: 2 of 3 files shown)


SKuipers:
Thanks for the PR! The Button class is already in use in the codebase so we'd need to be careful changing the constructor signature. Perhaps the $name value be used for both value and name?

The `getName()` and `setName()` methods are part of the InputAttributesTrait so this class should have access to them, it appears the `$this->setName($name)`call was just missed when this class was created. Looks like the two methods you've added may be ovewriting the base class methods and can safely be removed.

Also hate to be a pain 😅  but we're aiming to use PSR-2 as the codestyle for the core. Could you update these changes to use soft tabs (4 spaces). More info here: http://www.php-fig.org/psr/psr-2/#24-indenting  There's often CS-Fixers that can be installed depending on your editor.

Many thanks!!

jimbo8098:
Sure, sounds fine. I'll amend the branch.

RE PSR-2, I'll configure vim next time I'm working on it.

On 24 Oct 2017 9:07 am, "Sandra Kuipers" <notifications@github.com> wrote:

Thanks for the PR! The Button class is already in use in the codebase so
we'd need to be careful changing the constructor signature. Perhaps the
$name value be used for both value and name?

The getName() and setName() methods are part of the InputAttributesTrait so
this class should have access to them, it appears the $this->setName($name)call
was just missed when this class was created. Looks like the two methods
you've added may be ovewriting the base class methods and can safely be
removed.

Also hate to be a pain 😅 but we're aiming to use PSR-2 as the codestyle
for the core. Could you update these changes to use soft tabs (4 spaces).
More info here: http://www.php-fig.org/psr/psr-2/#24-indenting There's
often CS-Fixers that can be installed depending on your editor.

Many thanks!!

—
<https://github.com/GibbonEdu/core/pull/370#issuecomment-338908989>, or mute
the thread

jimbo8098:
Ok I've set vim to use 4 spaces instead of tab chars now. 

I've made a solution to the button class which leaves the constructors for the class alone. When created through FormFactory, the ID is assigned through a 3rd, optional, variable. If no ID is provided, the label is used, as before. If you want to use multiple buttons of the same name, you'll need to use the third variable to set it.

Here's what you can do with the changes:

```
    $frm = Form::create('test',null,'get');

    $frm->addRow()->addButton('hello'); //Create a button with a label and no onclick
    $frm->addRow()->addButton('hello again',null); //Create a button with an imaginary onclick in place of null
    $frm->addRow()->addButton('hello again again',null,'fancyid'); //Create a button with an ID

    //This snippet will make one button since the button objects will use the name 'hello' in the array
    $frmRow = $frm->addRow();
    $frmRow->addButton('hello');
    $frmRow->addButton('hello');

    //This snippet, with buttons embedded in a column will output 3 buttons in said column. You could embed the buttons in a row too since they work in a similar fashion.
    $frmRow2 = $frm->addRow();
    $rowCol = $frmRow2->addColumn();
    $rowCol->addButton('hello',null,'hello1'); //Button with ID 'hello1'
    $rowCol->addButton('hello',null,'hello2'); //Button with ID 'hello2'
    $rowCol->addButton('hello',null,'hello3'); //You get the idea...
    echo $frm->getOutput();
```

jimbo8098:
Bah! No way to change the branch. It's the buttonfix branch. I'll raise a new PR.

PR #371, opened 2017-10-21
Add WebLink Class

jimbo8098:
Adds a WebLink class with support for embedded Input objects. This was mainly to allow buttons to work on the system as it is for now but you can set this to support images like before as long as an Image class is added with the Input class. Trivially, could extend to allow the OutputableInterface objects as well.


SKuipers:
Hi Jim! I see where you're going with the WebLink class, I wonder though if it requires a separate class definition, as a link isn't necessarily a type of form input :thinking: Perhaps this is something that the `addContent()` method can be used with the `append()`, `prepend()` and `wrap()` methods?

It looks like the purpose of the ElementCollection class is a duplication of the Column class. To keep things concise it may be best to focus on updating the Column class if there's functionality missing. If you're looking the place column elements inline, there's an `inline` css class that could be added to the column with `addClass('inline')`

Apologies for the sparse documentation! you've done some great detective work figuring things out so-far 😃 I wonder which forms you're working on OOifying? Seems like an above-average number of buttons and links.

Thanks!!

jimbo8098:
Hey!

On addContent, I'd need to check into it. Is that implemented in Form()? If
we were to add addURL to Input, we could allow that element to be embedded
in an A tag at getElement()?

I'll see about the addClass method, that might fix it.

I'm sure we'll  have better docs soon, no worries there. I updated the
sheets document with the ones I work on. One of the forms had 3 buttons in
one column.


On 24 Oct 2017 9:22 am, "Sandra Kuipers" <notifications@github.com> wrote:

Hi Jim! I see where you're going with the WebLink class, I wonder though if
it requires a separate class definition, as a link isn't necessarily a type
of form input 🤔 Perhaps this is something that the addContent() method can
be used with the append(), prepend() and wrap() methods?

It looks like the purpose of the ElementCollection class is a duplication
of the Column class. To keep things concise it may be best to focus on
updating the Column class if there's functionality missing. If you're
looking the place column elements inline, there's an inline css class that
could be added to the column with addClass('inline')

Apologies for the sparse documentation! you've done some great detective
work figuring things out so-far 😃 I wonder which forms you're working on
OOifying? Seems like an above-average number of buttons and links.

Thanks!!

—
<https://github.com/GibbonEdu/core/pull/371#issuecomment-338912862>, or mute
the thread

jimbo8098:
Yes, I think the addClass setting would be best here. 

Re WebLink though, I'm not sure that would be best. I think it could be done by prepending the <a> tag then, on the last object, appending. That or maybe using a column object to enclose it all, could prepend and append then but there's no append/prepend on that or the Row class (which is extends). 

I can see append and prepend are present in the Element class which Layout implements. It's exactly what we'd need to do this. Technically, I think we could just extend Element since it's technically just a div.

So in short I think WebLink has it's place as a way to embed content in an A tag without needing to use divs, appends and prepends.

SKuipers:
I certainly see where wrapping content in links could be made easier with an additional class, it's not quite a type of form input though, so the use-case is puzzling 🤔 I wonder if by chance you're ooifying one of the data tables rather than a form?

SKuipers:
Hi Jim, We've merged this into the core. Started running into enough cases of links used inside forms (especially the Markbook) where it does make sense to add a clean implementation as a class rather than relying on appends().

Thanks!!

PR #378, opened 2017-10-24
Button Fix

jimbo8098:
Following on from previous PR where I clearly displayed my superior git knowledge and totally borked the branch... Here is a branch with just one fix instead of like 3 fixes bundled in!

Post copied from last PR:

==============

Ok I've set vim to use 4 spaces instead of tab chars now.

I've made a solution to the button class which leaves the constructors for the class alone. When created through FormFactory, the ID is assigned through a 3rd, optional, variable. If no ID is provided, the label is used, as before. If you want to use multiple buttons of the same name, you'll need to use the third variable to set it.

Here's what you can do with the changes:

```
    $frm = Form::create('test',null,'get');

    $frm->addRow()->addButton('hello'); //Create a button with a label and no onclick
    $frm->addRow()->addButton('hello again',null); //Create a button with an imaginary onclick in place of null
    $frm->addRow()->addButton('hello again again',null,'fancyid'); //Create a button with an ID

    //This snippet will make one button since the button objects will use the name 'hello' in the array
    $frmRow = $frm->addRow();
    $frmRow->addButton('hello');
    $frmRow->addButton('hello');

    //This snippet, with buttons embedded in a column will output 3 buttons in said column. You could embed the buttons in a row too since they work in a similar fashion.
    $frmRow2 = $frm->addRow();
    $rowCol = $frmRow2->addColumn();
    $rowCol->addButton('hello',null,'hello1'); //Button with ID 'hello1'
    $rowCol->addButton('hello',null,'hello2'); //Button with ID 'hello2'
    $rowCol->addButton('hello',null,'hello3'); //You get the idea...
    echo $frm->getOutput();
```

### DIFF
--- a/src/Forms/FormFactory.php
+++ b/src/Forms/FormFactory.php
@@ -164,9 +164,18 @@ class FormFactory implements FormFactoryInterface
         return new Input\MultiSelect($this, $name);
     }
 
-    public function createButton($label = 'Button', $onClick = '')
+    public function createButton($label = 'Button', $onClick = '', $id = null)
     {
-        return new Input\Button($label, $onClick);
+        if($id == null)
+        {
+            return new Input\Button($label, $onClick);
+        }
+        else
+        {
+            $button = new Input\Button($label, $onClick);
+            $button->setName($id);
+            return $button;
+        }
     }
 
     public function createCustomBlocks($name, OutputableInterface $block, \Gibbon\Session $session)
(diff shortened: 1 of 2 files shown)


SKuipers:
Thanks! I think you've found a bug where the `setName` and `setID` calls were forgotten when creating the Button class 😄 it's possible that adding in `$this->setName($name);$this->setID($name);` to the constructor may do the trick. The methods are inherited from the base Input class via the InputAttributesTrait

SKuipers:
Thanks for the changes, looks good! I'd say it's ready to merge 👍

PR #391, opened 2017-11-09
User Admin: Manage Users OOification

SKuipers:
OOification and tests, with a couple updates to the Form classes:

- Date format is automatically added to Date fields (if it's not already there)
- Form classes can define a getLabelContext() method, which will append extra text to the label that's linked to that element. Helps with input types that need to add extra help-text depending on how they're configured.
- Added a couple utility classes to Theme css
- Added a seeFieldIsNotEmpty method to acceptance testing


rossdotparker:
It's a big one!

PR #397, opened 2017-11-10
Form Class: add a prefab Delete Form, implement in Manage Roles

SKuipers:
This adds a new PrefabFormFactory class for creating predefined forms in one method call. Helps reduce duplicate code for forms that are nearly identical throughout the system. The createDeleteForm method has option for confirmation text-entry, off by default.

I've implemented it in Manage Roles as an example. Both the delete and duplicate actions function as modal windows with a small tweak to the code (also as an example, easy to revert out if needed).

### DIFF
--- a/src/Forms/PrefabFormFactory.php
+++ b/src/Forms/PrefabFormFactory.php
@@ -1,0 +1,61 @@
+<?php
+/*
+Gibbon, Flexible & Open School System
+Copyright (C) 2010, Ross Parker
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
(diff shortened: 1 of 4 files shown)


src/Forms/PrefabFormFactory.php
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @version v14
+ * @since   v14
+ */
+class PrefabFormFactory
+{
+    public static function createDeleteForm($action, $confirmation = false)
+    {
+        $form = Form::create('deleteRecord', $action);
+        $form->addHiddenValue('address', $_GET['q']);
```
IDisposable: Why putting the value in address too, it'll be injected as `q` in line 38 anyway?

SKuipers: Hi Marc! A good question. Many of the deleteProcess pages rely on an address value available in POST, the added line here prevents breaking the existing behaviour in the code.

IDisposable: Cool, just wanted to make sure this was intentional and not just cargo cult.  I'm JUST starting into the codebase and PHP is not my native/best language... sorry for any questions.

SKuipers: Questions are most welcome! This class is part of our ongoing effort to modernize the core, focusing in v15 on moving the html forms over to a set of OO classes. Feel free to give a shout as you delve into the codebase, also be sure to check out the support forum: ask.gibbonedu.org

rossdotparker: Hi @IDisposable thanks for stopping in to comment. I'm not sure we've met before (I may be mistaken), so welcome to the Gibbon community : )

PR #403, opened 2017-11-14
Data Updater: Update Personal Data OOification 

SKuipers:
Test coverage + OOification. There's an automated test but this one could use some extra manual testing.

❤️ deleting all that code!


rossdotparker:
It's a monster!

PR #420, opened 2017-11-27
School Admin: add search to Manage Facilities page

SKuipers:
**New Feature**

We got up to four pages on the Manage Facilities list, needed to add a search 😄 

_Edit: the fix for the build is in PR #417_


rossdotparker:
Thanks for adding this feature : ) I followed this one up with a commit that adds a little UI consistency with other parts of the system, in terms of how headers are used. I think there are plenty that use the unadorned style...but I'll try and catch the ones that I can to get them to be adorned with headers.

PR #428, opened 2017-12-08
Form Class: move DeleteForm into its own class, remove PrefabFormFactory 

SKuipers:
Changed my mind 😅 Having predefined forms for common things like delete forms, bulk action, etc still makes a lot of sense, but I realized piling them all into one Factory class was going to get messy really quick.

This PR updates all the PrefabFormFactory instances to point to a DeleteForm. Luckily the feature is recent and only used in the core. Thanks in advance for bearing with me on this one 😁


rossdotparker:
This was a fun one to read, just for the comment! We live and learn, I get it : )) Keep up the excellent work!

PR #439, opened 2017-12-14
Form Class Changes

raynichc:
**New Feature**
Block specific labels.
MultiSelect search bar.

## Description
CustomBlocks has been changed to make labels block specific. change() now called when date is selected in Date. MultiSelect now holds sort after adding or removing and also has a search for the source select. Back-end functions have also been prepared for migration to external document.

## Motivation and Context
Block specific labels allows for per block control of labels. change() being called on date change allows for javascript onChange() function to be used for date field. MultiSelect search allows for certain items to be found quickly.

## How Has This Been Tested?
These changes have been tested in an isolated test environment and in the Trip Planner submit form.


rossdotparker:
@SKuipers I think you are in a better position to review these changes. Can you do that and then merge when you are satisfied? Thanks : )

SKuipers:
Hi Ray! I've merged your PR, it's looking great! Made one small tweak to the datepicker function scope. I think we'll get a chance to try out the custom blocks out in the core when the planner gets ooified 😄 In v16 it'd be a good next step to start working out how to generalize the embedded javascript in these classes and move it into the core.js file.

PR #441, opened 2017-12-16
My Activities

jimbo8098:
Finished OOification for My Activities page. I did spot some issues with the 3 options on the page but they refer to other pages on Gibbon. Happy to investigate if that's something caused by the commit.


rossdotparker:
Hi Jim, thanks for working on this one. I've just realised, after reviewing, accepting and testing that this never had a <form in it, and so should not have been converted at this stage! So, I've reverted it, and will close it shortly. I appreciate the code commit, but don't want to get ahead of ourselves, as the tables may well be done in a different manner in v16 or 17. Thanks for the effort!

PR #445, opened 2018-01-03
Revert "My Activities"

rossdotparker:
Reverts GibbonEdu/core#441


rossdotparker:
Hi Jim, after reviewing, merging and testing this one, I realise it had no <form elements in it, and so should never have been listed in the conversion tracker! Rather than proceed with this now, I'd rather leave it until we tackle all tables in the system (v16 or v17 I guess), so that we can retain consistency. Sorry your work was in vain...my bad! Going to close this one now. Ross.

PR #446, opened 2018-01-05
System: add error handling and prevent error messages in Production

SKuipers:
Two main purposes for this PR: to always prevent the display of error messages in Production (regardless of server config), and to improve the display of error messages in Development and Testing installs.

## Description
- Uses the built-in set_exception_handler and set_error_handler functions to define custom behaviour for error handling.
- In Production any fatal errors will fallback to the generic Gibbon error page. Any notices and warnings will not display.
- In Development and Testing, any code errors/warnings/notices are wrapped in an error div and the stack trace is formatted for improved readability.
- Fixes the visual glitch where sidebars don't render after a fatal error or exception.

## Motivation and Context
It's likely a good idea to ensure the display_errors setting is always off for Production, especially security-wise by preventing leaked info through error messages, as well as the odd 'broken' page caused by a PHP notice.

## Screenshots:
screen shot 2018-01-05 at 2 32 34 pm
screen shot 2018-01-05 at 3 16 59 pm
screen shot 2018-01-05 at 2 03 30 pm

### DIFF
--- a/error.php
+++ b/error.php
@@ -19,10 +19,10 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 ?>
 
 <h1>
-	<?php echo __($guid, 'Oh no!'); ?><br/>
+	<?php echo __('Oh no!'); ?><br/>
 </h1>
 <p>
-	<?php echo __($guid, 'Something has gone wrong: the Gibbons have escaped!') ?><br/>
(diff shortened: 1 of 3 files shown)


rossdotparker:
This looks amazing. Well overdue, and elegantly constructed. Thanks : ) I am going to merge and test now.

PR #450, opened 2018-01-08
Future Absence OOification

jimbo8098:
OOification of attendance_future_byPerson.php.

    

-  attendance_future_by_person.php now (almost) fully OO. Had to implement a JS-style fix for colspan attribute as it's not supported by current FormRenderer. A setAttribute function would be perfect but that is protected in code.
- attendance_future_byPersonProcess.php has been slightly amended to send the user back to the page they were on along with providing the success response.
- css/module.css slightly amended to include a float:right stylisation on the class table

Be aware:

- The "SET FUTURE ATTENDANCE" box doesn't disappear when the absenceType is changed. There's currently no OO way to do this unless this is all made in the same form which would, as far as I can see, make the page look much different to what it was before.


SKuipers:
Hi Jim. Looking good. There's a couple spots needing translation wrappers `__()` and a missing `isActionAccessible()` check (line 93 of the original file), which we can catch post-merge.

Good idea to return the state back with the submitted form. This can also be added on at the top when the $URL is constructed to apply to each redirect. And well spotted on the setAttribute method visibility, it is indeed deliberate 😄 it's protected to help keep a lid on the attributes exposed through the form API. In the future the forms are likely to be rendered as divs rather than tables (for mobile responsiveness); aside from elements inside the Table class, which have a new-ish TableCell class to help with those attributes.

Thanks for the OOification work! It was a pretty massive undertaking for us this version and every bit of help is hugely appreciated 😃

PR #453, opened 2018-01-30
System: update userAdmin_statusCheckAndFix CLI

SKuipers:
Fixes an issue where parents with multiple families could be unintentionally set as Left if the children in one family have left, but others are still active.


rossdotparker:
Thanks for looking into this. The solution looks really elegant. I think this is a mysql mode issue, but the SQL on line 88 is returning "'gibbon_dev_core.gibbonFamilyAdult.gibbonFamilyID' isn't in GROUP BY" on my dev setup.

SKuipers:
Good catch. I've updated the query to play nicely with ONLY_FULL_GROUP_BY.

PR #460, opened 2018-03-06
Form Class: markbook-related additions and changes

SKuipers:
The markbook OOification is done but the change set ended up pretty big so I've split it into two parts. The first is just the related Form API changes, and some Formal Assessment updates that overlap with the same functionality:

Adds the following form elements to DatabaseFormFactory, and implements the change where needed:
- SelectSchoolYearTerm
- SelectGradeScale
- SelectGradeScaleGrade
- SelectRubric

Adjusts the append/prepend handlin g in Element class to be more succinct and allow nested wrap()s.

Adds the auto-resize functionality found in the markbook to the TextArera class as a method. 

Adds a conditional method which can be chained into form elements. This improves the overall readability of a form that has lots of conditionals, and allows for easier implementation of forms that have many different optional components. See it in action in Part 2 of this PR ;)

### DIFF
--- a/modules/Formal Assessment/internalAssessment_manage_add.php
+++ b/modules/Formal Assessment/internalAssessment_manage_add.php
@@ -18,6 +18,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
 use Gibbon\Forms\Form;
+use Gibbon\Forms\DatabaseFormFactory;
 
 //Module includes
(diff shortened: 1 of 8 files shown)


SKuipers:
Ack! Travis is right, don’t merge this yet 😅 I’ll push a fix.

SKuipers:
Fixed 😁

PR #463, opened 2018-03-09
Markbook: final OOification

SKuipers:
OOified forms for Add, Edit Markbook Column, Edit Data and Edit Targets.

Other notable changes:
- Adds fromQueryChained to select as a helper for transforming an SQL result set into a chained query.
- Separates out getHookedUnits() and renderStudentSubmission() into functions for re-usability/readability
- Uses the if() chained methods in Edit Data to try to keep the form readable. There's still a lot going on there, I hope to tackle it with some gateway classes at some point to improve it further.


rossdotparker:
That was a huge job, and it looks great. Thanks : )

PR #487, opened 2018-04-21
Spelling

fvlasie:
when send -> when sending

Yay! I managed to figure out separate pull requests! :)


rossdotparker:
I ended up dealing with this one when I merged the PR request that preceeded it, so this became redundant. Thanks all the same : )

PR #493, opened 2018-04-25
New "today" highlight

fvlasie:
Also see next pull request with updates to main.css

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -73,7 +73,7 @@ v16.0.00
 		System Admin: fixed spelling error in Department settings
 		System Admin: fixed typo in Third Party Settings
 		Timetable: added This Week link to timetable date navigation in Timetable by Person
-		Timetable: added highlight to the current day column header text in Timetable by Person
+		Timetable: added highlight to the current day column in Timetable by Person
 		User Admin: fixed Child's Name list in Edit Family to include all students
 		User Admin: fixed bug preventing some family links from showing up in Manage Users
 		User Admin: removed Back link from child and adult delete modals in Manage Families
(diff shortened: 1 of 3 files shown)


SKuipers:
Hi @fvlasie!

Now that you've mastered setting up separate branches, it may be worth noting that it's okay (and preferable) to put all your related changes together in one branch 😄 This helps make it easier to read & review all the parts of the change on one PR. For example, a PR called "Add Today highlight to timetable" could contain all the file changes for that idea, and then if you had a separate idea or change in a different area of Gibbon then you would create a new branch for it.

Adding commits to the same branch will update an existing PR. In this case, it may help to merge your patch-8 and patch-9 branches into this patch-7 branch, which will combine the changes on one place.

Thanks!

fvlasie:
Sandra, thank you for the tip! I hope I handled that correctly.

rossdotparker:
Hi FV, I think it would be better to avoid the extra div, and just add the class direct to the th tag. In addition, I'd be tempted, for now, just to use this class to set the colour, and leave everything else inline, as this will all need a more complete review in due course. Want to have a play with it?

fvlasie:
Alright! I will see if the highlight can be applied without new elements.

fvlasie:
I think everything is in order this time!

rossdotparker:
This is a big improvement. I've found a way to make it a little more concise, which I've implemented and committed. I noticed that you'd also highlighted the <td> for Timetable by Facility, but not the <th>, so that's also included in the same commit:

https://github.com/GibbonEdu/core/commit/bf165d7ab64a143e54df0e30178b361b1f0c9656

You've made a huge step in contributing to the project, and have already made the system better. Thanks for your time : )

fvlasie:
My pleasure Ross!

PR #496, opened 2018-04-26
Timetable: fix days outside the term displaying the wrong name 

SKuipers:
The `$rowDay` was being set inside a loop, and then checked for `isset()`. Added an unset at the beginning of the loop to clear the value.

Also tweaks the recent This Week/Last Week changes to the translation strings, so the added symbols are outside the string.


rossdotparker:
Thanks for this Sandra. Two for the price of one...well caught on the translation/symbols.

PR #502, opened 2018-05-03
Activities: remove debug output from activity attendance export

SKuipers:
I'm going to guess you were debugging this script at some point 😄 

screen shot 2018-05-03 at 2 41 17 pm


rossdotparker:
Good catch : )

SKuipers:
Credit goes to our ECA coordinator _"Why is this page now blank when it worked before"_ 😉

PR #504, opened 2018-05-06
Added notify class code for Planner Add and Edit pages

fvlasie:
I decided it was best to leave the notification off by default so the teacher could turn it on only when he thought it necessary.

### DIFF
--- a/modules/Planner/planner_add.php
+++ b/modules/Planner/planner_add.php
@@ -1221,7 +1221,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/planner_add.php') 
 						</td>
 						<td class="right">
 							<input type="checkbox" name="notify" value="on">
-							<label for="notify">Notify all class participants</label>
+							<label for="notify"><?php echo __('Notify all class participants') ?></label>
 							<input type="submit" value="<?php echo __($guid, 'Submit'); ?>">
 						</td>
 					</tr>
--- a/modules/Planner/planner_add.php
+++ b/modules/Planner/planner_add.php
@@ -1220,6 +1220,8 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/planner_add.php') 
 							<span class="emphasis small">* <?php echo __($guid, 'denotes a required field'); ?></span>
 						</td>
 						<td class="right">
+							<input type="checkbox" name="notify" value="on">
+							<label for="notify">Notify all class participants</label>
 							<input type="submit" value="<?php echo __($guid, 'Submit'); ?>">
 						</td>
 					</tr>
--- a/modules/Planner/planner_addProcess.php
+++ b/modules/Planner/planner_addProcess.php
@@ -19,6 +19,10 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 include '../../gibbon.php';
 
+use Gibbon\Comms\NotificationSender;
+use Gibbon\Domain\System\NotificationGateway;
+
+
 $URL = $_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/'.getModuleName($_GET['address']).'/planner_add.php';
 
 if (isActionAccessible($guid, $connection2, '/modules/Planner/planner_add.php') == false) {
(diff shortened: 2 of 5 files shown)


SKuipers:
Hi FV, this is looking great! 

I see that you've used the `setNotification()` function, which we're phasing out in favor of NotificationEvent objects (which now that I look, it should be marked as deprecated, and there's a few more setNotifications left in the code than I thought 😅). One more part of the transitional nature of the current refactoring. 

If you have a look at line 701 of Students/applicationFormProcess.php this has a great example of the NotificationEvent in use, which is very similar to the setNotification function but can be re-used: you create the event, add any number of recipients, then send it. Be sure to add `use Gibbon\Comms\NotificationEvent;` at the top of the file too.

Another note: when running an SQL prepare/execute function, it helps to wrap it in a try {} catch {} so you can handle any exceptions thrown by the statement. In the context of this code, a catch statement like:
```
$URL .= "&return=warning1$params";
header("Location: {$URL}");
exit();
```
would let you return a warning back from the process page. This tells the user that their action was completed (since the primary action is to add/edit the planner) but that something else failed along the way.

I'll add a few other small comments as notes on the code, overall this looks well thought out and a handy feature to have 👍

modules/Planner/planner_add.php
```diff
@@ -1220,6 +1220,8 @@ function outcomeDisplayElements(number) {
 							<span class="emphasis small">* <?php echo __($guid, 'denotes a required field'); ?></span>
 						</td>
 						<td class="right">
+							<input type="checkbox" name="notify" value="on">
+							<label for="notify">Notify all class participants</label>
```
SKuipers: Wrapping the text inside this label with `__()` will tell the code that it's translatable. Generally, anytime you write a string that the user will read it's best to add __().

modules/Planner/planner_addProcess.php
```diff
@@ -295,6 +295,22 @@
...
+                //GET CHECK BOX VALUE
+                if (isset($_POST['notify'])) {
+                    //Create notification for all people in class except me
+                    $dataClassGroup = array('gibbonCourseClassID' => $gibbonCourseClassID);
+                    $sqlClassGroup = "SELECT * FROM gibbonCourseClassPerson INNER JOIN gibbonPerson ON gibbonCourseClassPerson.gibbonPersonID=gibbonPerson.gibbonPersonID WHERE gibbonCourseClassID=:gibbonCourseClassID AND status='Full' AND (dateStart IS NULL OR dateStart<='".date('Y-m-d')."') AND (dateEnd IS NULL  OR dateEnd>='".date('Y-m-d')."') AND (NOT role='Student - Left') AND (NOT role='Teacher - Left') ORDER BY role DESC, surname, preferredName";
+                    $resultClassGroup = $connection2->prepare($sqlClassGroup);
+                    $resultClassGroup->execute($dataClassGroup);
+                    while ($rowClassGroup = $resultClassGroup->fetch()) {
+                        if ($rowClassGroup['gibbonPersonID'] != $_SESSION[$guid]['gibbonPersonID']) {
+                            $notificationText = (__("Lesson plan “".$name."” has been created."));
```
SKuipers: Strings inside __() are pulled out of the code so that they can be shared and translated, so this is one case where they can't be broken up and concatenated the way we would a normal string. The way around this is to use `sprintf()` with a `%1$s` token to represent the variable you need to add to the string. In this case, formatting it would look similar to: `sprintf(__('Lesson plan %1$s has been created.'), $name);`

I do wish all this stuff was in the docs for you! but for now the best way is to learn-by-doing and we're happy to help with suggestions 😄

fvlasie:
Sorry Sandra, I cannot quite figure out how to get it to work with the NotificationEvent instead of setNotification.

fvlasie:
Thank you for your help!

rossdotparker:
Thanks for working on this FV and Sandra. The end result looks great. I found one minor bug in edit, which is now fixed in this commit:

https://github.com/GibbonEdu/core/commit/dbc1455be1ae6be839091e3b30bf3ba689b4aa6e

FV I reckon it would be more efficient to factor out the adding/editing user in SQL, rather than PHP. Want to see if you can make that happen?

PR #508, opened 2018-05-11
Sundays show following days for Sunday people

fvlasie:
Sundays show following days in Timetable regardless of which day of the week is the "first". Previously, Sundays only showed following days if Monday was the "first" day. Sunday "first" people were feeling a little excluded!


rossdotparker:
@fvlasie I think this would be better as:

` if ($q == '' && ($_SESSION[$guid]['firstDayOfTheWeek'] == 'Monday' || $_SESSION[$guid]['firstDayOfTheWeek'] == 'Sunday')) {`

This way it only impacts these two particular cases, rather than say, a school that starts its week on Friday. Thoughts?

PR #526, opened 2018-05-24
System: mark deprecated functions, update where applicable

SKuipers:
I've marked the following functions as deprecated, and updated the code if a drop-in replacement was available (which is most of them).
- setNotification
- ynExpander
- formatName
- addressFormat
- getAge
- formatPhone
- printUserPhoto
- getUserPhoto
- dateConvert
- dateConvertBack
- dateConvertToTimestamp
- printPagination
- ExportToExcel

Removed:
- setStringReplacementList() Not used since v13
- dump() Not used since v13
- dateConvertToTimestampGM Not used anywhere, core or additional modules.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -18,6 +18,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
 use Gibbon\Forms\Form;
+use Gibbon\Services\Format;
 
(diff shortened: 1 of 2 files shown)


rossdotparker:
But I love hiding stuff in functions.php!

PR #566, opened 2018-06-10
Attendance: fix student attendance links based on subpage name change

SKuipers:


rossdotparker:
Well caught! I obviously neglected to make this change when I (finally) updated the name of the sub-page.

PR #593, opened 2018-08-16
Staff: add a Mark as Left bulk action to Manage Staff 

SKuipers:
**New Feature**

## Description
Adds a bulk action form to the Manage Staff page, with a Mark as Left action. This will set any selected staff members status to Left and update their End Date with the date provided.

## Motivation and Context
Any time there's a changeover in staff, such as the end of year, it's helpful to be able to see the full list of staff members and check off any that have left in one action.

Also adds a row highlight based on user status to make checking the list of active staff easier.


SKuipers:
Btw, the build for Travis should stop failing after #591 is merged, so it may be handy to do that one first.

PR #595, opened 2018-08-21
Messenger: restricted Manage Groups view to display groups from the current school year only

rossdotparker:
**Bug Fix**
Brings consistency between Manage Groups and New Message

**How Has This Been Tested?**
Locally for now, production test will follow merge.

### DIFF
--- a/src/Domain/Messenger/GroupGateway.php
+++ b/src/Domain/Messenger/GroupGateway.php
@@ -42,7 +42,7 @@ class GroupGateway extends QueryableGateway
      * @param QueryCriteria $criteria
      * @return DataSet
      */
-    public function queryGroups(QueryCriteria $criteria, $gibbonPersonIDOwner = null)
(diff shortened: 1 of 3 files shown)


src/Domain/Messenger/GroupGateway.php
```diff
@@ -54,7 +54,9 @@ public function queryGroups(QueryCriteria $criteria, $gibbonPersonIDOwner = null
             ->leftJoin('gibbonGroupPerson', 'gibbonGroupPerson.gibbonGroupID=gibbonGroup.gibbonGroupID')
             ->leftJoin('gibbonPerson', 'gibbonPerson.gibbonPersonID=gibbonGroup.gibbonPersonIDOwner')
             ->groupBy(['gibbonGroup.gibbonGroupID']);
-
+        
+        $query->where('gibbonGroup.gibbonSchoolYearID = :gibbonSchoolYearID')
```
SKuipers: The `->where()` can be chained into the query creation in this instance, because there's no conditional on it. Just code-style though 🤓, the changes look good and test out perfectly.

Issue #608, opened 2018-09-03
Code style: Separating display logic from business logic


yookoala:
To put it simply, separate "what to display" from "how to display them" into different code chunks.

To truly fix #208, the framework need to have enough flexibility to easily do theming code without messing up the core programming logic. Then applying existing responsive framework (e.g. [Bootstrap](https://getbootstrap.com/)) should be easy. This will also give more flexibility for theming in the future.

Most existing files, especially `index.php` suffers from mixing business logic with display logic. Like,

```php
<html>
<title>
<?php
                echo $_SESSION[$guid]['organisationNameShort'].' - '.$_SESSION[$guid]['systemName'];
                if ($_SESSION[$guid]['address'] != '') {
                    if (strstr($_SESSION[$guid]['address'], '..') == false) {
                        if (getModuleName($_SESSION[$guid]['address']) != '') {
                            echo ' - '.__($guid, getModuleName($_SESSION[$guid]['address']));
                        }
                    }
                }
?>
</title>
```

It can be broken down into 2 chunks:
```php
$title = $_SESSION[$guid]['organisationNameShort'].' - '.$_SESSION[$guid]['systemName'];
if ($_SESSION[$guid]['address'] != '') {
    if (strstr($_SESSION[$guid]['address'], '..') == false) {
        if (getModuleName($_SESSION[$guid]['address']) != '') {
            $title .= ' - '.__($guid, getModuleName($_SESSION[$guid]['address']));
        }
    }
}

```

```php
<title><?php echo $title; ?></title>
```

The 2 chunks can exists on different script files for that specific purpose. The 2nd chunk can later be easily re-written with modern theme engines. There is CPU and memory cost, but the difference, when running on modern computers, is acceptable.

crayner:
hear hear

rossdotparker:
Hi Koala, we hear you loud and clear! This is a process that is currently underway, as part of the refactoring work that @SKuipers is leading. In short, we are pulling at certain threads within the system (e.g. forms first, then tables, then themes) and rewritting each strand in an object-oriented style that also separates out logic into gateway classes. Sandra is currently working on a diagramatic representation of the upcoming verisons and what will change, and that will be shared within the community.

Are you interested in assisting us with this work? If so, email support@gibbonedu.org and I can invite you to join our developer Slack workspace.

Thanks for taking the time to share your ideas, it is most appreciated.

Ross

SKuipers:
[...]

PR #616, opened 2018-09-06
WIP: Refactor fullscreen.php

yookoala:
## Description
Refactor fullscreen.php to separate business logic and display logic.

## Motivation and Context
As part of the effort in addressing #608.

## How Has This Been Tested?
1. Unit test pass.
2. Open a test site fullscreen.php address.

### DIFF
--- a/fullscreen.php
+++ b/fullscreen.php
@@ -27,111 +27,103 @@ if (empty($_SESSION[$guid]['systemSettingsSet'])) {
 }
 //If still false, show warning, otherwise display page
 if (empty($_SESSION[$guid]['systemSettingsSet'])) {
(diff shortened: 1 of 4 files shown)


yookoala:
@SKuipers: I've add 2 HTML helper class here, do you think it is the right way to do this?

SKuipers:
Looking good. I like that you've split the js/css loading logic out into some well-commented methods. Some of these functions overlap with the ongoing refactoring & roadmapped items, I'll share some of more of these with you can perhaps we can work the functions/methods into the same set of classes. Thanks!

yookoala:
> I'll share some of more of these with you can perhaps we can work the functions/methods into the same set of classes. Thanks!

Sure. I'd love to.

PR #632, opened 2018-09-21
System Admin: handle i18n files during Install and Update

SKuipers:
This PR builds on the i18n features added in #628 to address the issue of installing and updating language files.

Database:
- Adds two fields to the gibboni18n table: `installed` & `version`

Installation:
- Step 2 will download & install the language file for the System Default language, ensuring each new system has at least one language installed.
- Step 2 displays a message if the download failed, so the user can install language files manually.
- The final install step will also check for & set installed to Y any languages that may have been manually installed by adding them to the `i18n/` folder.

Update:
- The update process will find & set installed to Y any languages where the files already exist (populating the Manage Languages list). It will also set installed to N for any files that no longer exist.
- The Manage Languages page now has an Update action, available for any language where the version installed is less than the current Gibbon version. 
- The Manage Languages page also detects manually installed language files, same as the Update/Install process.

Forms:
- OOified the SelectSystemLanguage options, which appear in the installer step 1 before the database has been setup.
- OOified the SelectI18n options, which appear in login forms and on the preferences page: only installed & active languages will appear in the list.


rossdotparker:
This is really brilliant work Sandra. All seems to test out perfectly. What a huge leap forward for Gibbon in terms of ease of management and flexibility when dealing with languages. Thank you : D

PR #633, opened 2018-09-23
Fix to account for daylight savings when rendering the table in the Tie Day to Dates page

shemco:
Fix to account for daylight savings when rendering the table in the Tie Day to Dates page.

**Bug Fix | New Feature | Documentation**

Bug Fix

## Description

Changes the code in ttDates.php to do date math with strtotime() thus avoiding the assumption that php will treat all days as 86400 seconds long.

## Motivation and Context

The calendar on the Tie Days to Dates page does not generate properly in a locale with daylight savings.

## How Has This Been Tested?

Loaded Tie Days to Dates page with a full year schedule. 
Before:
bad-rendering

After:
good-rendering


SKuipers:
I wonder if this might also affect the timetable rendering. Are you able to check and see if that's the case? (I don't happen to be in an area that uses DST)

Perhaps hop forward on your timetable 11/04/2018 and see if it looks right. I know the functions in modules/Timetable/moduleFunctions.php are using 86400 as part of the date calculation too. 

Thanks!

shemco:
I checked around DST dates in my timetable and didn't see any problems. If I run into this elsewhere in the future I'll push a fix. 


Thanks.
~shemco

SKuipers:
Awesome, thanks!

PR #635, opened 2018-09-26
Staff: post-ooification fix for non-basic Staff Type in Manage Staff

SKuipers:
A quick post-ooification fix: when a role has been selected for the Staff Type, the numeric ID was being displayed in tables:

screen shot 2018-09-26 at 2 09 46 pm


rossdotparker:
I think this is more than a display error. I think that before the form was OOified it stored the name as a value, but post-OOification it started to store the ID. Hence, this fix only works with newly saved date. Can you verify that what I've described is true?

SKuipers:
Sure, that works. I can make the change.

SKuipers:
Reverted the previous table change & applied the above mentioned fix.

PR #636, opened 2018-09-26
Timetable: add a bookings button to TT in View Timetable by Facility

SKuipers:
This is a usability tweak for schools using the facility booking options. It adds a quick way to book a period of time relative to the school's timetable, by looking up a room in View Timetable by Facility and clicking a + button to create a booking. This pre-fills the Add Facility Booking page with the details of the selected space & times.

screen shot 2018-09-26 at 3 22 08 pm


rossdotparker:
Lovely feature : ) It tests out well, except that the + link opens in a new tab, which is something Gibbon does not generally do for regular interface pages. I wonder if it might be better to open in the same tab, but selectively redirect to the timetable with the success message? Any thoughts?

PR #643, opened 2018-09-29
Data Updater: remove the start date check from required updates logic

SKuipers:
This PR reverts the start date logic added in #564 which was causing some inconsistencies in Data Updater reports. The Student & Family Data Updater History reports already filter by student enrollment, which should be sufficient.

I've done some tests to compare the updater history reports to the view that parents see on the My Data Updates page and they appear to be consistent.

This change set also adds some database keys for the Data Updater, Family Adult/Child and Student Enrolment tables. These address the performance issues encountered with the Family Data Updater History, and will generally help speed up any queries involving these tables.

Ross: you'll want to delete the keys you've added before updating, otherwise you'll end up with some unnecessary extra keys on your table. The reason your original keys didn't help with performance was the order of the compound keys: School Year has to be second, because in this case it's not involved in the query, so your keys were not being used. 

I did a diff of the keys on my test database vs. the demo database, the main difference was the enrolment key which I've added here. It appears the likely culprit for slow query speed with your data was that you have more years worth of past updates being joined to the query.


rossdotparker:
Hi Sandra, thanks for this amazing, timely fix! Reviewed, merged and tested, and all looks fantastic.

I'd already removed my testing indexes yesterday, so all OK on that front. In the past I've most only used single indexes (except for some composite uniques), but I was following your lead this time...but with not enough understanding it seems. Lesson learned, thanks for sharing!

PR #645, opened 2018-10-02
Planner: Name clarification and pre-checked notifications

fvlasie:
Just a couple of minor changes prompted by feedback from professors. Namely, changing the label of the "Name" field to "Name of Lesson Plan". It seems there was some confusion as to what name to put in there! Also, the "Notify all class participants check box" is now checked by default. Some professors were forgetting to check and it seemed that an opt-out approach was less hazardous.

### DIFF
--- a/modules/Planner/planner_add.php
+++ b/modules/Planner/planner_add.php
@@ -311,7 +311,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/planner_add.php') 
 					</tr>
 					<tr>
 						<td>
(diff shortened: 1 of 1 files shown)


fvlasie:
Hello Ross,

"Lesson Name" sounds good! As for the checkbox, I will just add a note to my upgrade instructions so I don't forget to reimplement it.

rossdotparker:
Thanks for this FV. I've added the same change to the Edit view, and updated CHANGELOG.txt: https://github.com/GibbonEdu/core/commit/e030ca1f68bac09d03a653d91eca98836e2ad684

PR #609, opened 2018-09-04
WIP: Code style update

yookoala:
An attempt in addressing #608.

## Description
* Restyle code in `index.php` and `likes.php` to separate business logic from display logic.
* Add documentation to `index.php` variables.
* Also some development helper files (docker-compose.yml and .editorconfig).
* Code are gradually changed between commits. Should be easy to follow step-by-step to see what was changed and why.

## Motivation and Context
To make the code base easier for theming and thus responsive design later.

## How Has This Been Tested?
Browsing development copy of the site with the committing changes.

PR #649, opened 2018-10-03
System: add Page & AssetBundle classes, plus initial Module & Theme classes

SKuipers:
No changes to the core just yet, still laying some more of the groundwork 🏗 

This PR adds four classes which will be used with refactoring `index.php` (and others). It also adds the accompanying PHPUnit Tests for each class.

@yookoala If you have a chance, please do take a look and let me know what you think. This builds on some of our discussion so-far, and works into the DI container & routing stuff I have in-the-works. With these classes in place I think we're getting closer to merging your awesome work in #609 and #616.

The `AssetBundle` is heavily inspired by your HTMLHelpers in #616, as a reusable class to make handling & merging the styles & scripts between page, module & theme easier. Rather than adding those to functions.php (which was the original way of doing things), I've setup some very basic `Module` and `Theme` classes which can become the basis for making the extendable parts of Gibbon object-oriented (manifests, version files, etc).

Similar to frameworks like WordPress, the AssetBundle separates out registering an asset and adding it: the core/module/theme can pre-define a script by registering it, which isn't included until it's added by name. This can help us cut down on loading every asset into every page, and instead a script can add just the assets it needs.

The `Page` class sets up some of the templating groundwork, similar to the `fullscreen` refactoring, which pulls together errors, warnings, scripts, styles, etc. so they can all be output at the end of the page's runtime.


yookoala:
I like the overall design of `AssetBundle`. In unified scripts and stylesheets and managed them in the same manner. But I think it'd me more flexible have context and version grouped under one `$options` parameter. That would make it more extensible.

Also, you might want to cater for inline scripts and styles as well. The current design only works with URL.

Example,

```php
    /**
     * Register a named asset for later use. Name should be unique.
     *
     * @param string $name      Unique identifier for this asset.
     * @param string $src          URL, relative to the system absolutePath, or inline content depends on
     *                                          `type` in `$options`.
     * @param array $options    Options for rendering, includes these fields:
     *                                          - string `type`
     *                                                      'url' (default) for URL as $content.
     *                                                      'inline' for inline script or style as $content.
     *                                          - string `context`
     *                                                      context of the definition.
     *                                                      for scripts: context should be the output location, eg: 'head', 'foot'
     *                                                      for styles: context should be the media type, eg: 'all', 'screen', 
     *                                                      'print'.
     *                                          - string `version`
     *                                                      The version number is appended to the asset URL for
     *                                                      cache-busting.
     */
    public function register($name, $src, $options=[])
    {
        // apply default value to key(s) that is not specified
        $options += [
            'type' => 'url',
            'context' => null,
            'version' => null,
        ];
        // combine definitions with src
        $this->registeredAssets[$name] = [
            'src'     => $src,
        ] + $options;
    }
```

yookoala:
One more note. It's a bit non-intuitive to have a class named `Theme`.

In practice, we'd have 3 kinds of assets in page rendering:
1. core assets (e.g. message and error style and scripts); and
2. module assets (e.g. page specific scripts and stylesheets).
3. theme assets (e.g. bootstrap scripts and stylesheets);

And apparently, the `Page` object should collect all of them. Then a theme implementation should read the `Page` object and render it accordingly.

The question, should theme developer develop their own theme class? Or should they use the default theme class in their custom code?

SKuipers:
Thanks for the awesome feedback!!

I did notice your implementation for inline script/style tags. I wonder if it may help enforce good front-end practices to now allow them in the `AssetBundles` class. That way, it doesn't give Module or Theme designers the option to just inline things that should be a separate js/css file. 

For the core, there's methods on the Page object for adding Head/Foot/Sidebar code, because at this point in the refactoring there's some stuff that just needs to be a raw chunk of HTML. By not limiting those code chunks to  only script/style tags, it allows some extra flexibility: additional <meta> in the head, modal-related divs in the footer, module-specific sidebar code, etc. 

For `AssetBundles`, it makes sense to keep `$context` as a required field, because they're used to filter where the assets will be rendered. Otherwise there would be orphaned assets. 

With only two $options params, one required and one optional, it'd then keep things easier to read & understand to not split them out into a separate array. This way, the array of asset data returned by the class `[name => src, context, version]` matches the same order as the methods for adding them.

SKuipers:
Yep, these classes will come into play for developers. The Module and Theme classes represent the two ways to extend Gibbon. The ones here are just initial implementations, to help get up and running with templating. I see them as the eventual OO replacement for the manifest.php file: base Module and Theme classes would be extended by developers to add in their functionality.

I think rendering the Page would still be a job for the core and not a specific theme implementation. This is where a templating engine like Twig can come into play: the necessary data and logic of page rendering is part of the core, but the layout can be split into a template file that could optionally be customized by the theme.

yookoala:
* One potential `AssetBundle` extra `$options` value is `weight`, which would come in handy if the module wants to raise / lower the script execution order of a specific script. The value can be used combined with [usort()](http://php.net/usort) before actually rendering.

* Module developer would want a way to easily inject PHP variables or settings to styles or javascript (e.g. user changable background color, JS library call parameters). Inline styles or scripts is the easiest way to do so. And if not, we'd need to provide alternative way to expose them.

* It make sense to expose page rendering logic with overridable template files in theme.

* If Module and Theme were designed to be implemented by developer, should they be abstract class or interface instead of a class?

yookoala:
If `Module` and `Theme` are abstract class or interface, we can have a `ClassicModule` and `ClassicTheme` that implements them by adapting data in existing manifest file(s). That way, converting existing module or theme would be less painful.

SKuipers:
>   If Module and Theme are abstract class or interface, we can have a ClassicModule and ClassicTheme that implements them by adapting data in existing manifest file(s). That way, converting existing module or theme would be less painful.

Yep, the Module and Theme as abstract makes sense, I'll make the change. They're certainly going to need more work, this is jsut the initial step to get a bit closer to page rendering. An added layer to handle the classic modules using manifest.php is that I was thinking too 👍 

>   One potential `AssetBundle` extra `$options` value is `weight`, which would come in handy if the module wants to raise / lower the script execution order of a specific script. The value can be used combined with [usort()](http://php.net/usort) before actually rendering.

Hmm, I'll have to think on this one. On the one hand it's good to plan for future needs, but on the other hand it'd be adding in more complexity right out of the gate.

>   Module developer would want a way to easily inject PHP variables or settings to styles or javascript (e.g. user changable background color, JS library call parameters). Inline styles or scripts is the easiest way to do so. And if not, we'd need to provide alternative way to expose them.

The `Page` object would still be accessible via the DI container for modules, so they could inject raw code with the `addHeadExtra`, `addFootExtra` `addSidebarExtra` methods, and in these cases it doesn't constrain the code to _only_ being styles and scripts.

yookoala:
Exposing variables might need to be in specific order to make sense.

On one hand, the inline code might need to call JS libraries, which needed to be declared before. On the other hand, the inline code might need to be used by other user scripts that follows. In the most complicated case, inline code might need to be specifically run between script A and script B.

So it provides more flexibility if, eventually,
* inline assets are managed with URL based assets.
* assets can be sorted regardless of the order they are added to the bundle.

Simply "addHeadExtra", "addFootExtra" and / or "addSidebarExtra" might not be flexible enough.

SKuipers:
Hmm, those are some very good points. It's a level of complexity that doesn't currently exist, but may certainly be worth considering. I'll take a look and see about a clean way to work that in.

rossdotparker:
@SKuipers @yookoala thanks for all your hard work getting this PR ready. Reading the comment log here is really revealing and informative. Your work is very much appreciated!

SKuipers:
Thanks @yookoala for the great questions! I've incorporated your suggestions into the AssetBundle class 👍

PR #651, opened 2018-10-05
WP: Further refactor index

yookoala:
## Description
* Improve Gibbon\View\AssetBundle and Gibbon\View\Page class.
* Refactor index.php with AssetBundle and Page.

## Motivation and Context
* The target is to, in long term, rewrite the index page with Twig template.
* The intermediate target is to narrow down the variables to use in template.

## How Has This Been Tested?
* Test with local testing server setup.
* No error on my test. Please check if you find any problem with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,8 @@ os:
 # list any PHP version you want to test against
 php:
   - 7.0
+  - 7.1
+  - 7.2
 
 # optionally specify a list of environments, for example to test different RDBMS
 env:
--- a/fullscreen.php
+++ b/fullscreen.php
@@ -1,24 +1,32 @@
 <?php
-/*
-Gibbon, Flexible & Open School System
-Copyright (C) 2010, Ross Parker
-
-This program is free software: you can redistribute it and/or modify
-it under the terms of the GNU General Public License as published by
-the Free Software Foundation, either version 3 of the License, or
-(at your option) any later version.
-
-This program is distributed in the hope that it will be useful,
-but WITHOUT ANY WARRANTY; without even the implied warranty of
-MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
-GNU General Public License for more details.
-
-You should have received a copy of the GNU General Public License
-along with this program. If not, see <http://www.gnu.org/licenses/>.
-*/
+/**
+ * Gibbon, Flexible & Open School System
+ * Copyright (C) 2010, Ross Parker
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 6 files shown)


SKuipers:
Thanks! I'll now need some time to work these changes into the templating stuff I've also been working on. Since the index affects pretty much everything in Gibbon, I've been going over the changes and manually testing each piece quite carefully. Takes more time, but well worth it with the scope of code that's been modified. Our test coverage isn't high enough to catch these things automatically.

I'd like this temp branch to not diverge too far from the main one, so for now I'll try to incorporate some more of your changes, and focus on getting the changes we've already made working and merged back in.

I notice there's a number of code style changes in these commits. We're aiming for PSR2 as our standard (not there yet in a lot of legacy files, but its the goal for future changes). Some of the added white-spaces seem like they're maybe from a different standard?

yookoala:
I used phpcs to check the code against PSR-2 and docblock standards. If it is not PSR-2 compliance, I can change it.

Please tell me what to change.

SKuipers:
No worries. For example, I see that:

```php
uasort($usedAssets, function ($a, $b) {
    return $a['weight'] <=> $b['weight'];
});
```

is changed into:
```php
uasort(
    $usedAssets, function ($a, $b) {
        return $a['weight'] <=> $b['weight'];
    }
);
```

Which isn't something my PSR2 php-cs/php-cs-fixer is requiring, and to me doesn't necessarily make the code more readable 🤔 

The DocBlock changes are interesting. I'm not sure about tagging every method on a class with the @since version, I think it may just keep things less cumbersome to track that on the class itself. Same with the @package @author @license on the classes, as these are the same project-wide they likely don't need to be added individually to every class. It's personal preference, but I tend to favor simplicity in these things 😄

SKuipers:
Hi Koala. I think what I may do is merge this PR in, so that I can rebase my changes ontop of these and work out any conflicts, then take a pause on merging any more PRs for a bit. We're both working on the exact same set of files right now 😅  so I need some time to get my changes worked into yours, and thoroughly tested.

Huge thanks for the work you've done so-far!! I know the index was very tangled up with logic and presentational code, and you've put in an awesome amount of effort breaking it apart into smaller, more manageable pieces of logic 👍  I'll aim to get my changes rebased and pushed to this branch soon, and then ideally very soon after that we can PR this dev branch back into v17.0.00.

yookoala:
Sorry. A but busy in the weekend. I meant to revert some code style change. But since its merged now, I'd better do that later in some other PR (if I have time). Thanks for your work.

I'll update my phpcs config. I think I ought to remove some rules. I agree that some of those changes are not improving readability.

SKuipers:
No worries, I reverted a few of the code style bits as I was merging in some of my changes too. Thanks again!

PR #653, opened 2018-10-08
System: complete refactor of index.php and fullscreen.php

SKuipers:
Here is is!! 🎉 This PR represents a huge collaborative effort, **many thanks** go to @yookoala for digging through the index to break apart and clarify a lot of the nested logic.

Is it 100% done? Nope! But it's a big step forward, it's stable, and I wanted to keep the change set manageable enough to read & review. There's more work to be done for sure. I've added a few thoughts and TODO items in the index.php file as reminders.

Each piece of logic and each major change has been tested and double-checked in as many ways as I can think of. There are hopefully as few bugs as humanly possible 😅 but we will certainly want to keep testing it thoroughly with different platforms & schools. 

Ideally, since it's a refactor there shouldn't be any noticeable changes. However there are a few nice performance boosts:

- Page loads feel smoother. This is because the output buffering reduces the flash of content loading between pages, and loads the whole page as one chunk.
- Fullscreen loading times seem much faster, particularly for rubrics and the delete confirmations.

### Templates

This PR adds templates for rendering pages in Gibbon. Much of the HTML for each a page has been separated out into these template files, which use a simple markup syntax to render the content. 

Templates in the core can be quite easily customized by theme developers by including their own modified copy in the theme's `/templates` folder. This is the next step towards responsive design, one step among many in our refactoring efforts, and the next step from here will be to start tackling the css files.

### Index

This breaks apart much of the nested logic into smaller components. As we move to HTTP routing, many of these will become part of the middleware layer (redirects, session populating, etc.). Other parts like the dashboards and sidebar can be refactored into their own templates. Any time there's large chunks of mixed logic and html (particularly in functions.php), these will be good candidates for using templates. There's also queries that will need moved into Gateway classes, but to help keep the change footprint small I've decided to do these in a future PR.

### Page Class

The page class is much like a View in the MVC pattern, with some added capabilities to handle Gibbon's particular setup (eg: modules, themes, alerts). The page holds an internal `$content` array, and has methods to `write` a string, `writeFromFile` to include a PHP file in a protected scope and output-buffer the result, and `writeFromTemplate` which loads and renders a template file. Rather than scripts accessing the template engine directly, they can simply call `render` to output the view.

### Module & Theme Classes

This is the starting point of more refactoring to come. I'd like to take some time now to map out the functionality that developers will need when building these kinds of extensions in the future. In the interim, they've been somewhat simplified to just hold the scripts and styles they need.

### Other Notable Changes

- Adds the `action`, `module`, `theme` and `page` as shared services in the CoreServiceProvider.
- Adds an `addMultiple()` method to AssetBundle and updates some of the defaults.
- Moves the code for initializing javascript libraries to it's own file, and passes configuration from PHP to JS via a json encoded array (thanks Koala!)


yookoala:
One comment thou. IMO, weight should be an additional sorting factor other than the natural order when asset is added. So I don't think we should automatically calculate the weight [like this](https://github.com/GibbonEdu/core/pull/653/files#diff-910e71a4f5691b1619c79132ae21f859R68). All assets added without weight is considered weight 0.

Then any positive and negative weight numbers would have an absolute priority reference. If an asset has weight -1, it should be sorted before all normal assets. If an asset has weight 1, it should be sorted after all normal assets.

Auto-calculation makes module developer uncertain what weight to use.

SKuipers:
Ahh, yep, that's a very good point. I had discovered that uasort won't maintain the original order of the array after sorting, even of all the weights are 0, so I added that default weighting in. I'll take another look and see about a different way to handle the issue, so that the weights can be kept at 0 be default.

yookoala:
I found a PHP implementation of a uasort-like function that claims to keep the array order: 
http://php.net/manual/en/function.uasort.php#114535

SKuipers:
Thanks, that is quite the function 😅 I've tweaked the sorting behaviour to keep the array index and use it internally as a secondary uasort criteria, so that weight is considered first but can also default to 0.

PR #658, opened 2018-10-15
Messenger: add a read-only role category field for Quick Wall messages

SKuipers:
Handles the edge case where a user can post a Quick Wall message but doesn't have access to the `New Message_role` target. In this instance, the Role Category target will appear read-only and the targets will be retained via hidden fields:

screen shot 2018-10-15 at 1 41 46 pm


rossdotparker:
Brilliant, thanks : )

PR #668, opened 2018-10-25
Attendance: OOification of Daily Attendance table

yookoala:
**Refactor**

## Description
OOification of the 2 daily attendance table (by roll group, by course class) in attendance.php. Also refactor the breadcrumb and code style.

## Motivation and Context
* Follow PSR-2 standard.
* OOification of table render code.
* Refactor of breadcrumb after #661.

## How Has This Been Tested?
* Tested on local development environment.
* Tested with Codeception acceptance tests.

## Screenshots (if appropriate):
**Before refactor**

**After refactor**

### DIFF
--- a/modules/Attendance/attendance.php
+++ b/modules/Attendance/attendance.php
@@ -51,34 +51,32 @@ if (!isActionAccessible($guid, $connection2, '/modules/Attendance/attendance.php
 }
 
 // define attendance filter form, if user is permit to view it
-if (isActionAccessible($guid, $connection2, '/modules/Attendance/attendance.php')) {
-    $form = Form::create('action', $session->get('absoluteURL').'/index.php', 'get');
+$form = Form::create('action', $session->get('absoluteURL').'/index.php', 'get');
 
-    $form->setTitle(__('View Daily Attendance'));
-    $form->setFactory(DatabaseFormFactory::create($pdo));
-    $form->setClass('noIntBorder fullWidth');
+$form->setTitle(__('View Daily Attendance'));
+$form->setFactory(DatabaseFormFactory::create($pdo));
+$form->setClass('noIntBorder fullWidth');
 
-    $form->addHiddenValue('q', "/modules/".$session->get('module')."/attendance.php");
+$form->addHiddenValue('q', '/modules/'.$session->get('module').'/attendance.php');
 
+$row = $form->addRow();
+$row->addLabel('currentDate', __('Date'))->description($_SESSION[$guid]['i18n']['dateFormat'])->prepend(__('Format:'));
+$row->addDate('currentDate')->setValue(Format::date($currentDate))->isRequired();
+
+if (isActionAccessible($guid, $connection2, '/modules/Attendance/report_rollGroupsNotRegistered_byDate.php')) {
     $row = $form->addRow();
-    $row->addLabel('currentDate', __('Date'))->description($_SESSION[$guid]['i18n']['dateFormat'])->prepend(__('Format:'));
-    $row->addDate('currentDate')->setValue(Format::date($currentDate))->isRequired();
-
-    if (isActionAccessible($guid, $connection2, '/modules/Attendance/report_rollGroupsNotRegistered_byDate.php')) {
-        $row = $form->addRow();
-        $row->addLabel('gibbonPersonID', __('Staff'));
-        $row->addSelectStaff('gibbonPersonID')->selected($gibbonPersonID)->placeholder()->isRequired();
-    } else {
-        $form->addHiddenValue('gibbonPersonID', $session->get('gibbonPersonID'));
-    }
-
-    $row = $form->addRow();
-    $row->addFooter();
-    $row->addSearchSubmit($gibbon->session);
+    $row->addLabel('gibbonPersonID', __('Staff'));
+    $row->addSelectStaff('gibbonPersonID')->selected($gibbonPersonID)->placeholder()->isRequired();
+} else {
+    $form->addHiddenValue('gibbonPersonID', $session->get('gibbonPersonID'));
 }
 
+$row = $form->addRow();
+$row->addFooter();
+$row->addSearchSubmit($gibbon->session);
+
 // define attendance tables, if user is permit to view them
-if (isActionAccessible($guid, $connection2, '/modules/Attendance/attendance.php') && isset($_SESSION[$guid]["username"])) {
+if (isset($_SESSION[$guid]["username"])) {
     // generator of basic attendance table
     $getDailyAttendanceTable = function ($guid, $connection2, $rowID, $takeAttendanceURL) use ($session) {
 
--- a/modules/Attendance/attendance.php
+++ b/modules/Attendance/attendance.php
@@ -27,6 +27,9 @@ use Gibbon\Tables\Renderer\SimpleRenderer;
 //Module includes
 require_once __DIR__ . '/moduleFunctions.php';
 
+// get session object
+$session = $container->get('session');
+
 // rendering parameters
 $currentDate = (isset($_GET["currentDate"])==false) ? date("Y-m-d") : Format::dateConvert($_GET["currentDate"]);
 $today = date("Y-m-d");
@@ -34,7 +37,7 @@ $lastNSchoolDays = getLastNSchoolDays($guid, $connection2, $currentDate, 10, tru
 $accessNotRegistered = isActionAccessible($guid, $connection2, "/modules/Attendance/report_rollGroupsNotRegistered_byDate.php")
     && isActionAccessible($guid, $connection2, "/modules/Attendance/report_courseClassesNotRegistered_byDate.php");
 $gibbonPersonID = ($accessNotRegistered && isset($_GET['gibbonPersonID'])) ?
-    $_GET['gibbonPersonID'] : $_SESSION[$guid]["gibbonPersonID"];
+    $_GET['gibbonPersonID'] : $session->get('gibbonPersonID');
 
 // show access denied message, if needed
 if (!isActionAccessible($guid, $connection2, '/modules/Attendance/attendance.php')) {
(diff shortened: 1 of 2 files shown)


SKuipers:
Looking great! I like the use of clone to re-use the DataTable structure 👍

I'll test out the new GitHub suggest tools in the code review, looks like you can now accept suggested changes right in the UI.

yookoala:
@SKuipers: Sorry, but I couldn't see the code review message :-(

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,433 @@
...
+                //
+            }
+
+            if ($result->rowCount()>0) {
+                $count=0;
+
+                $attendanceByCourseClass = [];
+                while ($row = $result->fetch()) {
+                    // Skip unscheduled courses
+                    //if ( isset($ttHistory[$row['gibbonCourseClassID']]) == false || count($ttHistory[$row['gibbonCourseClassID']]) == 0) continue;
```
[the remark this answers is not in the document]
yookoala: Yup. Not sure if I should remove it or not. So I just left it here.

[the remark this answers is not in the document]
SKuipers: Ah okay, looks like it's dead code, feel free to take it out.

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,433 @@
...
+        ->width('40px');
+
+    if ($currentDate>$today) {
+        $page->addError(__("The specified date is in the future: it must be today or earlier."));
+    } elseif (isSchoolOpen($guid, $currentDate, $connection2)==false) {
+        $page->addError(__("School is closed on the specified date, and so attendance information cannot be recorded."));
+    } elseif (isActionAccessible($guid, $connection2, "/modules/Attendance/attendance_take_byRollGroup.php")) {
+
+        // Show My Form Groups
+        try {
```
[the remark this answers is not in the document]
SKuipers: Sure, `/modules/Roll Groups/rollGroups.php` has an example of creating the gateway from the container and using it for a basic select. The class is found in `src/Domain/School/RollGroupGateway.php. In most cases, the SQL would be moved over to the class as a basic select, and interacts with the database via the Connection interface. 

In the case of paginated data tables the QueryableGateway part come into play, which uses Aura SQL query builder to create the more complex queries to handle sorts, filters, pagination, limits, etc. An example of the same gateway being used in this case would be `/modules/School Admin/rollGroup_manage.php`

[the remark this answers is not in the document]
yookoala: I've read 07e2e7e2. But I'm not sure how I should proceed with Attendance module.

Should I add new method to the module gateway class? Or should I create new gateway class implementation in the module?

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,448 @@
...
 
-    $form->addHiddenValue('q', "/modules/".$_SESSION[$guid]['module']."/attendance.php");
+    $form->addHiddenValue('q', "/modules/".$session->get('module')."/attendance.php");
 
-	$row = $form->addRow();
-	    $row->addLabel('currentDate', __('Date'))->description($_SESSION[$guid]['i18n']['dateFormat'])->prepend(__('Format:'));
-	    $row->addDate('currentDate')->setValue(dateConvertBack($guid, $currentDate))->isRequired();
+    $row = $form->addRow();
+    $row->addLabel('currentDate', __('Date'))->description($_SESSION[$guid]['i18n']['dateFormat'])->prepend(__('Format:'));
+    $row->addDate('currentDate')->setValue(Format::dateConvert($currentDate))->isRequired();
```
SKuipers: Just tested out the branch and caught a couple more small things. The date picker UI uses the localized date, in this case just Format::date() does the trick.

```suggestion
    $row->addDate('currentDate')->setValue(Format::date($currentDate))->isRequired();
```

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,448 @@
...
+                $attendanceByRollGroup[] = $row;
+            }
+
+            // define DataTable
+            $attendanceByRollGroupTable = clone $dailyAttendanceTable;
+            $attendanceByRollGroupTable->setTitle(__('My Roll Group'));
+            if (isActionAccessible($guid, $connection2, "/modules/Attendance/attendance_take_byRollGroup.php")) {
+                $attendanceByRollGroupTable->addActionColumn()
+                    ->width('50px')
+                    ->setAction('takeAttendance')
```
SKuipers:
```suggestion
                    ->addAction('takeAttendance')
```

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,448 @@
...
+                    // grouping 10 days as a row
+                    $day = $row['recentHistory'][$i + $j];
+                    $link = '';
+                    $content = '';
+
+                    // default link and content
+                    if (!empty($day['currentDate']) && !empty($day['currentDayTimestamp'])) {
+                        // link and date content of a cell
+                        $link = './index.php?' . http_build_query([
+                            'q' => '/modules/Attendance/attendance_take_byCourseClass.php',
```
SKuipers: Hmm, this is also linking to `attendance_take_byCourseClass.php` for the Form Group/Roll Group, which results in an error 🤔 Should be `attendance_take_byRollGroup.php` for that case.

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,448 @@
...
+            $attendanceByRollGroupTable->setTitle(__('My Roll Group'));
+            if (isActionAccessible($guid, $connection2, "/modules/Attendance/attendance_take_byRollGroup.php")) {
+                $attendanceByRollGroupTable->addActionColumn()
+                    ->width('50px')
+                    ->setAction('takeAttendance')
+                    ->setLabel(__('Take Attendance'))
+                    ->setIcon('attendance')
+                    ->setURL(
+                        'index.php?' . http_build_query([
+                            'q' => '/modules/Attendance/attendance_take_byRollGroup.php',
```
SKuipers: The index.php is already added by the Action class, so just `/modules/Attendance/attendance_take_byRollGroup.php` works for this method, then you can use `addParam(key, value)` or `addParams(array)` to add the url params.

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,448 @@
...
-
-
-							print "</tr>" ;
-						}
-					print "</table>" ;
-				}
-			}
-		}
-	}
+<?php if (isset($form)) { ?>
```
SKuipers: These can go right in the php code, no need to escape out of the php block with `?>` and `<?php`

modules/Attendance/attendance.php
yookoala: Should I change all the `Format::dateConvert` to `Format::date`? Or just this one?

SKuipers: Ah, good catch, if they're using `$currentDate` then yep, change then all to `Format::date`

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,447 @@
...
+
+        // action column, if user has the permission
+        if (isActionAccessible($guid, $connection2, $takeAttendanceURL)) {
+            $dailyAttendanceTable->addActionColumn()
+                ->width('50px')
+                ->addAction('takeAttendance')
+                ->setLabel(__('Take Attendance'))
+                ->setIcon('attendance')
+                ->setURL($takeAttendanceURL)
+                ->addParam($rowID)
```
SKuipers: This is throwing an ArgumentCountError when I test it, looks like it's because `addActionColumn()` returns an ActionColumn object, and then in your chained methods, `addAction()` returns an Action. The `addParam()` without passing a value is a method on ActionColumn. Moving the addParam call above addAction fixes it. Splitting this chained call may help clarify what's going on too.

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,447 @@
...
+                        $row['today'] = 'not timetabled';
+                    }
+                    $row['in'] = ($resultLog->rowCount()<1)? "" : ($log["total"] - $log["absent"]);
+                    $row['out'] = $log["absent"];
+
+                    $attendanceByCourseClass[] = $row;
+                }
+
+                // define DataTable
+                $takeAttendanceURL = '/modules/Attendance/attendance_take_byRollGroup.php';
```
SKuipers:
```suggestion
                $takeAttendanceURL = '/modules/Attendance/attendance_take_byCourseClass.php';
```

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,447 @@
...
+
+            // define DataTable
+            $takeAttendanceURL = '/modules/Attendance/attendance_take_byRollGroup.php';
+            $attendanceByRollGroupTable = $getDailyAttendanceTable(
+                $guid,
+                $connection2,
+                'gibbonRollGroupID',
+                $takeAttendanceURL
+            );
+            $attendanceByCourseClassTable->setTitle(__('My Roll Group'));
```
SKuipers:
```suggestion
            $attendanceByRollGroupTable->setTitle(__('My Roll Group'));
```

modules/Attendance/attendance.php
```diff
@@ -16,488 +16,447 @@
...
-
-
-							print "</tr>" ;
-						}
-					print "</table>" ;
-				}
-			}
-		}
-	}
+<?php if (isset($form)) {
```
SKuipers: Getting there! You've removed some of the `<?php` `?>` escaping, but I wonder why not all of them? There's no HTML or templating here, so we likely don't need to escape from php execution at all.

SKuipers:
Everything checks out in local testing with these recent changes, so I've added Ross to do the final review and merge. This was certainly a complex one to tackle, many thanks for your hard work on this!

It's a big file, so the code is still a lot to read and understand, but I think it's a good step in refactoring this and it'll get more succinct as the queries are moved to gateway classes. The anonymous function to generate a table class feels a bit awkward, conceptually, but I think this is a new use-case for reusing DataTables, so I may tinker with some ideas going forward for the tables API 👍

yookoala:
I'm not sure how best to carry on with the function.

If there is a controller class, it should probably be a private method or something. But that's a whole new scope of work to do (e.g. the whole file should probably be rewritten as a single controller function that returns the `$page` object). Meanwhile, I'd leave it as an anonymous function.

SKuipers:
Yep, you're right. I think refactoring is all about incremental gains, so you've certainly helped move things forward. When we introduce routing and controllers, down the road, these kinds of imperative code can be moved into classes that work on more of a request-response model.

PR #669, opened 2018-10-26
System: change read-only methods to properties in Page, Module, Theme class

SKuipers:
I've changed my mind 😅 The use of methods to enforce read-only access to Page properties has been annoying me, because it's hack-ish and the intent of the method isn't clear. Best to make this change sooner rather than later. 

I've removed these methods and opted to use the built-in `__get()` method instead. This allows read-only property access and helps ensure the intent of code using those properties is more readable.

### DIFF
--- a/index.php
+++ b/index.php
@@ -255,7 +255,7 @@ $javascriptConfig = [
  */
 
 // Set page scripts: head
-$page->scripts()->addMultiple([
+$page->scripts->addMultiple([
     'lv'             => 'lib/LiveValidation/livevalidation_standalone.compressed.js',
     'jquery'         => 'lib/jquery/jquery.js',
     'jquery-migrate' => 'lib/jquery/jquery-migrate.min.js',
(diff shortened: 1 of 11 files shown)


crayner:
Hi Sandra,
    You perhaps need to read up on the use of mutable and immutable within
PHP and Javascript.  This may help you make a decision that reflects
current trends.  (Not saying that your decision is wrong or right, just
perhaps not informed.)  I do use dumb objects that use getters and setters
for each exposed property, as that is the way the framework that I am using
ensures appropriate levels of mutability.
Grace and Peace

Craig Rayner
craig@craigrayner.com
Mobile: +61 456 560 018

crayner:
Hi Sandra,
    You perhaps need to read up on the use of mutable and immutable within PHP and Javascript.  This may help you make a decision that reflects current trends.  (Not saying that your decision is wrong or right, just perhaps not informed.)  I do use dumb objects that use getters and setters for each exposed property, as that is the way the framework that I am using ensures appropriate levels of mutability, but I also use __get  and __set callables when required.

SKuipers:
Thanks Craig, a fair point to consider. I'm not entirely unaware of immutable vs. mutable objects 😉 
Refactoring is always a learning-process along with a balancing-act: rewiring the code to preserve the original intent and business logic, keeping future-maintainability in mind, and ideally not introducing too much complexity. 

I think my next step would be to move the object creation out of the Page and properly dependency-inject them, to make good use of the DI container, so I wonder if this would still be an applicable case for immutability 🤔 As you've said, there's use-cases for both. Worth considering 👍

PR #678, opened 2018-10-31
Modules: Class autoloading

yookoala:
**New Feature**

## Description
A new autoloader script to autoload module class files if the modules are following standard folder structure:

- modules
  - My Module
    - src
      - SubPackage1
        - MyClass1.php
          (`Gibbon\Module\MyModule\SubPackage1\MyClass1`)
      - SubPackage2
        - MyClass2.php
          (`Gibbon\Module\MyModule\SubPackage2\MyClass2`)
      - MyClass.php
        (`Gibbon\Module\MyModule\MyClass`)

That is,
1. All class should be namespaced `Gibbon\Module\{ModuleName}\{SubpackageIfApplicable}`
2. All class files should be within the `src` folder in the module folder.
3. Class files can resides in sub-folder structure. The sub-folder names should also reflect in the namespace subpackage name. (as specified in PSR-4).
4. The `ModuleName` portion will be converted to spaced string `Module Name` for directory search (to provide backward compatibility for spaced module folders).

## Motivation and Context
To easier debug and develop new feature, It is more preferable to have a standardized way to autoload class files from modules. 

## How Has This Been Tested?
* Path speculation is tested with phpunit.
* Functionality is tested with Travis / Codeception.

### DIFF
--- a/cli/attendance_weeklySummaryEmail.php
+++ b/cli/attendance_weeklySummaryEmail.php
@@ -33,7 +33,6 @@ if (!isCommandLineInterface()) {
     setCurrentSchoolYear($guid, $connection2);
 
     require_once __DIR__ . '/../modules/Attendance/moduleFunctions.php';
-    require_once __DIR__ . '/../modules/Attendance/src/AttendanceView.php';
     $attendance = new AttendanceView($gibbon, $pdo);
     
     $firstDayOfTheWeek = $gibbon->session->get('firstDayOfTheWeek');
--- a/composer.json
+++ b/composer.json
@@ -63,6 +63,7 @@
         }
     },
     "autoload": {
+        "classmap": ["modules/"],
         "files": ["src/Scripts/autoload.php"],
         "psr-4": { "Gibbon\\": ["src/", "src/Gibbon"] }
     }
--- a/composer.json
+++ b/composer.json
@@ -63,6 +63,7 @@
         }
     },
     "autoload": {
+        "files": ["src/Scripts/autoload.php"],
         "psr-4": { "Gibbon\\": ["src/", "src/Gibbon"] }
     }
 }
--- a/gibbon.php
+++ b/gibbon.php
@@ -45,16 +45,6 @@ if (!$gibbon->isInstalled() && !$gibbon->isInstalling()) {
     exit;
 }
 
-// Autoload the current module namespace
-if (!empty($gibbon->session->get('module'))) {
-    $moduleNamespace = preg_replace('/[^a-zA-Z0-9]/', '', $gibbon->session->get('module'));
-    $autoloader->addPsr4('Gibbon\\Module\\'.$moduleNamespace.'\\', realpath(__DIR__).'/modules/'.$gibbon->session->get('module').'/src');
-
-    // Temporary backwards-compatibility for external modules (Query Builder)
-    $autoloader->addPsr4('Gibbon\\'.$moduleNamespace.'\\', realpath(__DIR__).'/modules/'.$gibbon->session->get('module'));
-    $autoloader->register(true);
-}
-
 // Initialize using the database connection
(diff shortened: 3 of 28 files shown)


SKuipers:
Hi Koala. I can see what you're aiming for here, but I'm not quite keen on the idea of adding a separate `spl_autoload_register` layer when the composer $autoloader class already manages this. There also seems to be a lot of `preg_match`ing happening here, along with a static method on the Module class, which is an approach that could interfere with this classes separation of concerns.

For namespacing, we have two different kinds of modules: Core Modules, and Third-party Modules. You're right, all Core modules should use the `Gibbon\Module\{ModuleName}` namespace. This will help keep things clear and autoloadable. In this case, it looks like the autoloader on line 50 of gibbon.php just needs updated to this pattern.

For third-party modules, they should in theory use their own Vendor namespace, based on common namespace conventions, rather than `Gibbon\Module`. In this case, it may be as simple as the abstract class having a `getNamespace` method that modules can implement and return their own namespace, and the gibbon.php autoloader can resolve it for them to the module's `/src` directory.

Does this make sense? Let me know : )

yookoala:
The line 50 of gibbon.php only maps the "current module". But if a module is calling another module's class file (e.g. this happens with the `AttendanceView` class), the autoloader doesn't help at all.

We can do similar thing, without the `spl_autoload_register` black magic, by using the "[classmap](https://getcomposer.org/doc/04-schema.md#classmap)" directive. But then developers will have to run `composer dump-autoload` everytime after a class file is added (or removed). The `spl_autoload_register` makes it easier to develop new class files.

We can still do the `classmap` directive and have the `spl_autoload_register` as a fallback. That would be more performant.

yookoala:
Another more universal way to do it is to:
1. Get a list of all the installed module; then
2. do the `$autoloader->addPsr4()` magic against all of them.

yookoala:
Unless there is no inter-dependencies between modules, some inter-module class call is bound to happen. And the current `$autoload->addPsr4()` trick wouldn't be of much help. It'd be preferable to have all module class always autoladable by autoloader.

SKuipers:
Hmmmmm 🤔  I'd like to avoid having to fetch the full module list and autoload each one if at all possible, as that seems like some added overhead for what is a fairly edge-case. For the most part, if all of the Core files are located in the root `/src` folder, they're available to all modules. 

Perhaps an option might be to have a system-wide method to request a single module be autoloaded, as it's needed.

yookoala:
I think unless there is no class file in module (i.e. all moved to `src` folder), it's wrong to assume that those files are useless elsewhere. And it is not developer friendly to make exception in class discovery than autoloading. Making extra functions to include and register module class is do-able, but not preferable.

For performance sake, Ive added 2 changes:
1. I've added the modules folder to `autoload.classmap` in `composer.json`. So the autoloading script is only run if a developer created a class but has not run `composer dump-autoload` yet.
2. Autoloading now uses `explode` to validate the namespace. So even if the script is run, the speed would be faster.

yookoala:
Rebased on v18.0.00.

SKuipers:
Hi Koala,

Thanks for putting this PR together, I realize it's been sitting open for a crazy long time 😆 I do agree with the need to handle module autoloading, and I think it ties into a broad need to address loading the current module scope in several respects: autoloading, templates, routes, service providers, assets, etc. With a system as large as Gibbon, it makes sense to aim to load only what is needed for a particular runtime, and lazy-load as much as possible. 

I'd love to revisit this in v19, where we're ideally going to start looking at HTTP routing and controllers for the core. For the time being, our scope in v18 is focused on wrapping up and testing the features introduced in this version. I'm going to close this issue for now, but it's certainly on my mind and something we're looking to address in future versions 👍 I'd be keen to catch-up and chat sometime about your thoughts on the larger roadmap of Gibbon refactoring!

PR #680, opened 2018-10-31
Added data visualisation to Rubrics

rossdotparker:
**New Feature**

## Description
Adds a new way to view rubric data, taking Historic data and displaying it as a polar area chart.

## Motivation and Context
It fits with some Mastery Transcript Consortium work we are looking into. 

## How Has This Been Tested?
Locally, using a range of different data.

## Screenshots (if appropriate):
screenshot 2018-10-31 at 5 26 18 pm

### DIFF
--- a/modules/Rubrics/moduleFunctions.php
+++ b/modules/Rubrics/moduleFunctions.php
@@ -106,6 +106,7 @@ function rubricEdit($guid, $connection2, $gibbonRubricID, $scaleName = '', $sear
                 $col->addContent('<b>'.__($outcome['name']).'</b>')
                     ->append(!empty($outcome['category'])? ('<i> - <br/>'.$outcome['category'].'</i>') : '')
                     ->append('<br/><span class="small emphasis">'.$outcome['scope'].' '.__('Outcome').'</span>');
+                $rows[$i]['title'] = $outcome['name'];
             } else {
                 $col->addContent($rows[$i]['title'])->wrap('<b>', '</b>');
             }
@@ -220,7 +221,8 @@ function rubricView($guid, $connection2, $gibbonRubricID, $mark, $gibbonPersonID
                 $cells[$rowCells['gibbonRubricRowID']][$rowCells['gibbonRubricColumnID']] = $rowCells;
             }
 
-            //Get other uses of this rubric in this context
+            //Get other uses of this rubric in this context, and store for use in visualisation
+            $contexts = array();
             if ($hasContexts) {
                 $dataContext = array('gibbonPersonID' => $gibbonPersonID);
                 $sqlContext = "SELECT gibbonRubricEntry.*, $contextDBTable.*, gibbonRubricEntry.*, gibbonRubricCell.*, gibbonCourse.nameShort AS course, gibbonCourseClass.nameshort AS class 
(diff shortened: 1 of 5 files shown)


modules/Rubrics/moduleFunctions.php
```diff
@@ -240,128 +242,244 @@ function rubricView($guid, $connection2, $gibbonRubricID, $mark, $gibbonPersonID
...
-                        $output .= "$(this).removeClass('rubricCellHighlight');";
-                        $output .= "mode = 'Remove';";
-                    $output .= '}';
-                    $output .= 'var request=$.ajax({ url: "'.$_SESSION[$guid]['absoluteURL'].'/modules/Rubrics/rubrics_data_saveAjax.php", type: "GET", data: {mode: mode, gibbonRubricID : "' . $gibbonRubricID.'", gibbonPersonID : "'.$gibbonPersonID.'", gibbonRubricCellID : cellID, contextDBTable : "'.$contextDBTable.'",contextDBTableID : "'.$contextDBTableID.'"}, dataType: "html"});';
-                    $output .= '});';
-                $output .= '});';
-                $output .= '</script>';
-            }
+                $output .= "<div id='canvasHolder' style='width:100%;'>";
+                    $output .= "<canvas id='birdsCanvas'></canvas>";
```
SKuipers: Birds? 🤔

modules/Rubrics/moduleFunctions.php
```diff
@@ -240,128 +242,244 @@ function rubricView($guid, $connection2, $gibbonRubricID, $mark, $gibbonPersonID
...
+                $output .= "</div>";
+
+                $labels = "" ;
+                foreach ($means as $mean) {
+                    $labels .= "\"".$mean['title']."\",";
+                }
+                if ($labels != '') {
+                    $labels = substr($labels, 0, -1);
```
SKuipers: This is a somewhat clunky way to concatenate csv strings 😉 An alternate might be to collect them as an array and then implode with a `,`

modules/Rubrics/rubrics_edit_editRowsColumns.php
```diff
@@ -187,25 +187,29 @@
...
-								$row->addSelect('gibbonScaleGradeID['.$count.']')
+								$column->addSelect('gibbonScaleGradeID['.$count.']')
 									->setID('gibbonScaleGradeID'.$count)
 									->fromQuery($pdo, $sql, $data)
 									->isRequired()
 									->selected($rubricColumn['gibbonScaleGradeID']);
 							}
 							$form->addHiddenValue('gibbonRubricColumnID['.$count.']', $rubricColumn['gibbonRubricColumnID']);
 
+							//$row = $form->addRow();
```
SKuipers: Some dead code?
```suggestion
							
```

modules/Rubrics/rubrics_edit_editRowsColumns.php
```diff
@@ -187,25 +187,29 @@
...
+							$column->addLabel('columnVisualise'.$count, sprintf(__('Visualise?'), ($count + 1)));
```
SKuipers: Hmmm ... maybe this field needs a description. I'm not quite sure what `Visualise?` will mean in this particular context.

SKuipers:
Mastery Transcript Consortium sounds like a super-villain organization, got any links to read more about this? 😄

modules/Rubrics/moduleFunctions.php
rossdotparker: I used some sample code, and then never cleared it up. Just changed it to "canvas" now.

modules/Rubrics/rubrics_edit_editRowsColumns.php
rossdotparker: I think a description would be good, but I was not sure how to format it. It looks rather clunky. Any suggestions from the table OO guru?

modules/Rubrics/moduleFunctions.php
rossdotparker: I've had a go at this, and it is now simpler. Could it be more elegant? Perhaps!

PR #685, opened 2018-11-02
Modules: update .gitignore

yookoala:
**Nitpicking**

## Description
* Use a whitelist based structure for `modules/.gitignore`.
* Ignore everything except for the core modules.

## Motivation and Context
Kind of a nitpicking. But it seems strange for the current [`modules/.gitignore`](https://github.com/GibbonEdu/core/blob/ca259435be1a8de70d7340b9f643315122f4e819/modules/.gitignore) to specifically only ignore one optional module, "Query Builder".

For the convenience of developing and committing live installation, its seem to be reasonable to ignore everything except for the core modules.

## How Has This Been Tested?
Tested locally.

### DIFF
--- a/modules/.gitignore
+++ b/modules/.gitignore
@@ -51,3 +51,7 @@
 !/Tracking/**
 !/User Admin
 !/User Admin/**
+
+# black list some files and folders
+# in the module folders
+**/.DS_Store
--- a/modules/.gitignore
+++ b/modules/.gitignore
@@ -1,1 +1,53 @@
-Query Builder
+# ignore all modules
+*
+
+# whitelist .gitignore file
+!.gitignore
+
+# whitelist all core modules
+!/Activities
+!/Activities/**
+!/Attendance
+!/Attendance/**
+!/Behaviour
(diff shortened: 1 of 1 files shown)


SKuipers:
Looks like an interesting approach, and could save a developer from maintaining their own `.git/info/exclude` file for installed modules (which is what I do). 

@rossdotparker what do you think?

modules/.gitignore
```diff
@@ -1 +1,53 @@
...
+!/System Admin
+!/System Admin/**
+!/Timetable
+!/Timetable/**
+!/Timetable Admin
+!/Timetable Admin/**
+!/Tracking
+!/Tracking/**
+!/User Admin
+!/User Admin/**
```
SKuipers: We may also need to ignore certain files inside the whitelisted modules:

```suggestion
!/User Admin/**

# Ignore system files in any module
**/.DS_Store
```

rossdotparker:
Looks good to me @SKuipers if you want to go ahead and merge. Thanks Koala!

PR #686, opened 2018-11-02
Tables: Allow user to override no result output of Simplerenderer

yookoala:
**New Feature**

## Description

Add `SimpleRenderer::renderNoResult` and `SimpleRenderer::setNoResultRenderer` to make the default no result behaviour overridable.

## Motivation and Context
In refactoring tables in Attendance module, I found out that some tables need to have different output that the default in SimpleRenderer. For example:

* People > Attendance > Reports > Class Not Registered
  (`modules/Attendance/report_courseClassesNotRegistered_byDate.php`)

In that case, an empty table simply means "all classes have been registered".

The current no result behavior limited to either:
* If this is a multi-page data set, you'll have a no result warning message.
* If this is a simple page data set without data, you'll have a no result error message.
So neither fit the case. And there was nothing you can do about it.

## How Has This Been Tested?
* Locally
* Travis

### DIFF
--- a/src/Tables/Renderer/SimpleRenderer.php
+++ b/src/Tables/Renderer/SimpleRenderer.php
@@ -35,7 +35,18 @@ use Gibbon\Forms\Layout\Element;
 class SimpleRenderer implements RendererInterface
 {
     use BasicAttributesTrait;
-    
+
+    /**
+     * A callback to render string display if there is no result
+     * in the DataSet.
+     *
+     * The callback should have the same function siguature as
+     * the `SimpleRenderer::renderNoResult` method.
+     *
+     * @var callable|null
+     */
+    protected $noResultRenderer = null;
+
     /**
      * So simple ...
      */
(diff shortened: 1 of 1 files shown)


yookoala:
WIP. Will PR later.

SKuipers:
Great idea, I certainly see the need for this. Before we get too far and add custom noReturn messages into individual files, it may be a good opportunity to address the default state for empty results. I'd like us to move towards something closer to what GitHub and other interfaces use (rather than an error message):

screen shot 2018-11-05 at 4 04 44 pm

I may tinker with this idea a bit, and possibly push some changes to your PR. Thanks!

yookoala:
Is there anything relies on the current "error message" behavior? I'm not familiar enough with the code to make the call to change it.

SKuipers:
No problem. Lets keep this PR open for now, and I'll take a look & come back to it.

SKuipers:
This one's kind of stuck in a chicken-or-the-egg state 🤔 I'd love to update the tables to change how we display the 'blank slate' message, but until we have more of the table ooification done it doesn't make sense to make a visual change like this or things will start looking pretty inconsistent. But, your ooifcation is also likely pending on this change 🐔🥚

I'll tinker with it and see if we can use some templating to solve this problem, and use the template for the blank slate rather than injecting the html directly. This way, we could improve the behaviour sooner, and add the visual parts to the templates later.

yookoala:
Take your time.

Note that I created this PR because I encountered something that's hard to translate to the new OOification code (see context). I'll wait for the result of this before I resume my work there.

SKuipers:
Hi Koala. Haven't forgot about this one! We've been refactoring a lot of the templates recently and I have one in the pipe for an updated Data Tables template. I'm working your suggested changes into the new templates, so your current PRd code won't nessesarily be compatible, but has certainly helped highlight something we needed to tackle. I'm closing this PR for now, many thanks for you work 👍

PR #688, opened 2018-11-05
Migrate Tcpdf library to composer vendor

yookoala:
**Refactor**

## Description
Remove the "/lib/tcpdf" folder in favor of using composer to manage it.

## Motivation and Context
See #691.

## How Has This Been Tested?
* Locally
* Travis


yookoala:
**Note:** WIP. Don't merge yet.

SKuipers:
You mentioned a few other lib > vendor migrations, can we start with the easy ones first? 😁

yookoala:
I made this so its easier to review the changes. I don't expect this to be merged anytime soon :-)

SKuipers:
Is this one still a work-in-progress? Is so, what are the remaining issues to sort out, just the differences in the fonts folder?

yookoala:
Yup. The files in the original `lib/tcpdf` is mixed versions. Although the version file claims it is '6.0.038', some of the files are clearly from the older version of the same library. Also there are some extra files.

I think it is relatively safe to use the whole set of files from the real '6.0.038' library. But I think test is need. If all the PDF generation works, then I think there is no point not using this PR. The problem is I don't know what feature is using the tcpdf library here.

SKuipers:
Awesome, will test it out. Is this PR using the real '6.0.038' then? PDF generation is used largely by external modules: Free Learning and Reporting (free and commercial). Plus, personally I have a couple custom modules for my school that use it, so I will test them out :)

Were the extra PHP files in `lib/tcpdf` mostly font related? I think Free Learning and Reporting are both using an added DroidSansFallback font because it has better UTF-8 support for Chinese characters. I've been using it as well. I think we could perhaps add this font into the resources folder, and use `TCPDF_FONTS::addTTFfont` to load it in, by managing the TCPDF class through the DI container.

I think anyone using additional custom fonts in their modules may just need to add the `TCPDF_FONTS::addTTFfont` method to their scripts too. It's not likely there's many cases of that though, and they would be custom modules not listed in the Extend page if they are.

rossdotparker:
Our version of Andy's reporting module uses some different fonts for Chinese and Japanese, but I am not sure if these are "custom fonts" or included within TCPDF. Is there a list of core fonts, or is any extra font considered a custom font? Thanks!

yookoala:
This PR is using '0.0.038' from https://github.com/yookoala/TCPDF, which is merely a fork of https://github.com/tecnickcom/TCPDF with a lot of unused tags removed to improve composer speed.

The extra files includes:
* font files,
* font related PHP,
* demo PHP files,
* various unknown PHP files, includes:
  * `2dbarcodes.php`
  * `barcodes.php
  * `pdf417.php`
  * `qrcode.php`
  * `spotcolors.php`
  * `tcpdf_barcodes_1d.php`
  * `unicode_data.php`

For the last category, I have no idea what they are, or if they are useful.

SKuipers:
> Our version of Andy's reporting module uses some different fonts for Chinese and Japanese, but I am not sure if these are "custom fonts" or included within TCPDF. Is there a list of core fonts, or is any extra font considered a custom font? Thanks!

@rossdotparker Try searching your version of the module for `addTTFfont` and `SetFont`, if its using custom fonts they'll be loaded in with these methods (I think it's the only way).

rossdotparker:
As just confirmed in Slack (but here for public record), our Reporting module uses only helvetica, dejavusans and stsongstdlight.

PR #690, opened 2018-11-05
Migrate Google API client to composer vendor

yookoala:
**Refactor**
## Description
* Remove the Google API Client from `/lib` folder
* Add Google API Client library (of the same version) to composer's management
* Remove all manual include for using Google API Client. Rely on Composer autoloader.

## Motivation and Context
See #691.

## How Has This Been Tested?
* Locally
* Travis


yookoala:
I don't understand why it marks 10,551 files "changed".

I'll check again the version of Google API Client to use and to migrate to.

yookoala:
We can do the gitignore + remove trick to keep only what we need in the repository.

yookoala:
I've reduced the number of changed files from 10,551 to 9724. I think this is as far as I can get. For readability, I broke down the changes into several commit steps.

The main reason of the large number is because the current "/lib/google/google-api-php-client" folder includes its own vendor folder and dependencies. They are essential for the library to work properly. So the only way to do the migration is to merge that vendor folder with the root one.

If you read the diff of 3cc829d, we're merely moving the main library files around. We're using the latest version of the depending libraries supporting PHP 7.0 (the libraries originally in the vendor folder). And I trust Composer resolved the dependencies correctly.

I'd be more understandable if you read the diff of each individual commit in this PR.

SKuipers:
Yep, looking good. The way you've broken the commits apart does make it easier to read, thanks! I think I'll aim to merge #692 next, then get you to rebase this one afterwards so it's ready to merge.

SKuipers:
Okay, PHPMailer and PHPExcel are looking good 👍  If you can rebase this one I'll aim to merge & test it next. Thanks!

yookoala:
Rebased. Pending Travis.

Issue #691, opened 2018-11-05
Composer based library management


yookoala:
**Refactor**

## Description

This issue is supposed to be for tracking a few upcoming PRs that aims to solve the same problem.

Composer provides easier to track version control and locking mechanism for modern PHP libraries. It also come with an automatically generated autoloader for any OOP style libraries. It encourage and assist the use of PSR-4 compliant libraries. Switching to autoloader based library calls remove the problem with library path discovery and a whole class of problem.

* Every PHP library that can be managed by composer should be.
* Every PHP class that can be autoloaded by composer autoloader should be.

## Motivation and Context

* To simplify the external library management in long run.
* To make it easier to upgrade to different version of the same library in the future.

## Libraries to Migrate

* [x] PHPMailer
* [x] PHPExcel
* [x] Google API Client
* [x] TCPDF

## PHP in the "/lib" Folder That We Won't Touch Yet

* `/lib/google/index.php`
  A Gibbon custom script written with Google API Client. Not part of the API Client library itself. Will not touch this until we have a good solution for controller / routing for such case.
* `/lib/paypal/*.php`
  Custom written script for payment. Should rewrite this completely with official PayPal client library or other payment library available in Packagist.

SKuipers:
Just merged the last of these changes into the core 🎉 Thanks!!

PR #693, opened 2018-11-05
Migrate PHPMailer library to composer vendor 

yookoala:
**Refactor**
## Description
* Remove the PHPMailer from `/lib` folder
* Add PHPMailer (of the same version) to composer's management
* Remove all manual include for using PHPMailer. Rely on Composer autoloader.

## Motivation and Context
See #691.

## How Has This Been Tested?
* Locally
* Travis


crayner:
Looks like something I did in version 12. Rejected at the time. Hope you
survive!

yookoala:
@crayner: I think you're talking about #119. And from the [comments](https://github.com/GibbonEdu/core/pull/119#issuecomment-252548560), it seems the PR was rejected because its against master branch instead of v13 branch, the development branch at the time. So they're probably not that against the idea (or I hope so :-P).

SKuipers:
I think open source is all about learning, for the contributors and maintainers. 

As a small team creating new norms, we're continuously adapting and figuring out how to incorporate everyone's work in a way that fits in the bigger picture. For the particular PR mentioned, we learned that long-running branches with >8000 files changes aren't a good fit for the stability of the codebase. We're able to pass that learning experience on and let contributors know to keep their changes small, incremental and testable. 

I think in this way, we hope to build a developer community that loves to learn as much as we do: both from their successful PR merges, and also from the constructive feedback they receive.

composer.json
```diff
@@ -35,14 +35,20 @@
...
     },
     "require-dev": {
         "twbs/bootstrap": "^4.1"
     },
     "replace": {
         "pimple/pimple": "*"
     },
+    "config": {
+        "platform": {
+            "php": "7.0"
```
SKuipers: The docs say this configuration "Lets you fake platform packages" https://getcomposer.org/doc/06-config.md#platform Can you let me know why we might need to do this (since we're using real php, why does it need to be faked?) Thanks!

yookoala: Because I am using PHP 7.2, if my machine is resolving the new composer.lock, they will not be compatible with 7.0. This is to ensure that the versions in lock is compatible with the lower bound of target compatibility.

yookoala: This also works for developers with similar situation in the future.

SKuipers: Okay, that makes sense. Thanks!

PR #699, opened 2018-11-08
Markbook: refactor breadcrumbs

yookoala:
**Refactor**
## Description
* adapt the new `$page` global for breadcrumb.
* add missing copyright declaration to files that missed it.

## Motivation and Context
* See #696
* Some files has missing copyright declaration.

## How Has This Been Tested?
* Locally
* Travis

### DIFF
--- a/modules/Markbook/markbook_edit_add.php
+++ b/modules/Markbook/markbook_edit_add.php
@@ -17,8 +17,9 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
-use Gibbon\Forms\Form;
 use Gibbon\Forms\DatabaseFormFactory;
+use Gibbon\Forms\Form;
+use Gibbon\Services\Format;
 
 //Module includes
 require_once __DIR__ . '/moduleFunctions.php';
@@ -53,7 +54,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Markbook/markbook_edit_add
         echo __($guid, 'The highest grouped action cannot be determined.');
         echo '</div>';
     } else {
-        $gibbonCourseClassID = $_GET['gibbonCourseClassID'];
+        $gibbonCourseClassID = $_GET['gibbonCourseClassID'] ?? '';
         if ($gibbonCourseClassID == '') {
             echo "<div class='error'>";
             echo __($guid, 'You have not specified one or more required parameters.');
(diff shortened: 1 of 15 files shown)


modules/Markbook/markbook_edit_add.php
```diff
@@ -78,9 +78,22 @@
                 echo '</div>';
             } else {
                 $course = $result->fetch();
-                echo "<div class='trail'>";
-                echo "<div class='trailHead'><a href='".$_SESSION[$guid]['absoluteURL']."'>".__($guid, 'Home')."</a> > <a href='".$_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/'.getModuleName($_GET['q']).'/'.getModuleEntry($_GET['q'], $connection2, $guid)."'>".__($guid, getModuleName($_GET['q']))."</a> > <a href='".$_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/'.getModuleName($_GET['q']).'/markbook_view.php&gibbonCourseClassID='.$_GET['gibbonCourseClassID']."'>".__($guid, 'View').' '.$course['course'].'.'.$course['class'].' '.__($guid, 'Markbook')."</a> > </div><div class='trailEnd'>".__($guid, 'Add Column').'</div>';
-                echo '</div>';
+
+                $page->breadcrumbs
+                    ->add(
+                        sprintf(
```
SKuipers: I can see where you're aiming to clarify the code with `sprintf`, to me it may not be a great choice for string concatenation 🤔 Another readable way to format this might be:

```
$className = Format::courseClassName($course['course'], $course['class']);
$urlParams = ['gibbonCourseClassID' => $_GET['gibbonCourseClassID'] ?? ''];

$page->breadcrumbs
    ->add(__('View').' '.$className.' '.__('Markbook'), 'markbook_view.php', $urlParams)
    ->add(__('Add Column'));
```

It uses Format::courseClassName (which ideally will standardize the formatting) and it's likely that the $urlParams will be re-used by other actions on the page, and could even be supplied automatically once we have controllers & PSR7 request objects.

What do you think?

yookoala: Personally, I hate contcat long string with space. I think in long run, the function `__()` should be able to handle named substitutions, like:

```php
__('View @className Markbook', [
  '@className' => Format::courseClassName($course['course'], $course['class']),
]);
```

That might even fix the limitation to word order in the current translations (i.e. sensible word order in English may be ridiculous in Chinese). An intermediate `sprintf` can be translated to that easier, IMO.

But I can deal with it your way, if you insist.

Reference Designs:
* https://api.drupal.org/api/drupal/includes%21bootstrap.inc/function/t/7.x

As for `$urlParams`, I did not read the files well enough to do it yet. If there is no need to such an array variable further down, I see no reason to generate a new variable.

SKuipers: That's a fair point. I think when Ross wrote the function it was likely based off of the Wordpress equivalent: https://developer.wordpress.org/reference/functions/__/ which uses the text domain as a parameter. Then we messed around with it a bit to start phasing out the $guid. Changing the method signatures now may have an adverse affect on the gettext translation tools. I can certainly see the utility value of building string-formatting into the translation though 🤔 hmmmmm.

For string order, in much of the rest of the codebase it uses `sprintf` with `%1$s` tokens eg: `__('Column %1$s Title')`  Perhaps in this case, it's time to create some new strings for the Add/Edit/View Markbook breadcrumbs, so they do properly handle string order.

yookoala: I'd still prefer a named replacement. That would make the life of the translator much easier. The translator might not understand what `%1$s` means in the programming context. But `@className` would be much more humanly readable.

SKuipers: Yep, that's fair, and I can see the value of it in the long run. The tricky part with a codebase this size is we already have >6000 strings in use. There's a few different approaches to take for the long run (modifying the function signature? introducing a new similar function?). Feel like starting an issue to discuss?

SKuipers:
Thanks Koala for making these changes. I think your choice of strtr works as an interim solution, and for the long-term we'll carry on the discussion in #701 👍

Issue #701, opened 2018-11-08
Improve translation API by named parameter replacement


yookoala:
This is a comment on the general design of translation API [`__()`](https://github.com/GibbonEdu/core/blob/ca259435be1a8de70d7340b9f643315122f4e819/functions.php#L58-L79) and [`Gibbon\Locale::translate()`](https://github.com/GibbonEdu/core/blob/ca259435be1a8de70d7340b9f643315122f4e819/src/Gibbon/Locale.php#L146-L204). The changes needed will have to be a long term one. So I'm just raising this issue here for discussion.

## Current Situation

Translations are done through the `__()` translation function in `functions.php`. It is undergoing a transition to remove `$guid`. Before, it expects input like this:

```php
/**
 * @param string $guid GUID of Gibbon installation
 * @param string $text Text to translate
 * @param string (Optional) Text domain. Unique identifier for retrieving translated strings. Default value: null
 *
 * @return  string Translated Text
 */
function __($guid, $text, $domain=null);
```
After the transition, it'll expect something like this:

```php
/**
 * @param string $text Text to translate
 * @param string (Optional) Text domain. Unique identifier for retrieving translated strings. Default value: null
 *
 * @return  string Translated Text
 */
function __($text, $domain=null);
```

The underlying translation is done by the global `$gibbon->locale` object, which is an instance of `Gibbon\Locale`. The actual translation method looks like this:

```php
/**
 * Custom translation function to allow custom string replacement
 *
 * @param   string  Text to Translate
 * @param   boolean Use guid.
 *
 * @return  string  Translated Text
 */
public function translate($text, $domain = null);
```

## Problems of the Approach

In actual world, strings are not always translated as is. There will be parameters within a string for translation. And right now, the API does not handle them at all. Programmers would deal with it like this:

```php
echo __('View') . ' ' . $course['course'] . ' ' . $course['class'] . ' ' . __('Marks');
```

or

```php
echo __('Posted') . ' ' . $time['minutes'] . __('minutes') .  ' ' . __('ago');
```

There are 2 problems:
1. string concat is hard to read and easily get wrong; and more importantly
2. string order is fixed without context. (Which is not ideal for translating to some languages)

[...]

PR #708, opened 2018-11-14
Gettext guid cleanup

yookoala:
**Refactor**
## Description
* The translation function `__()` is not changed at all. Just the usages are refactored.
* Mass refactored all the php files in the repository to replace `__($guid, ...)` with equivlant call `__(...)`.
* The command used in modules, lib, core, cli and installer is `find . -type f -name '*.php' -exec sed -i 's/__(\$guid, /__(/g' "{}" \;`
* The root php files are handled separately in editor.
* Also checked and fix a few `__($guid,...)` calls that was not discoverable by `sed` pattern search.
* Since the changes are straightforward, the result should be safe to use.

## Motivation and Context
* Reduce complexity in the translation string references in the current core code base. Reduce probability of errors.
* This will also be hugely beneficial to the works in #701.

## How Has This Been Tested?
* Locally.
* `php -l` calls.
* Travis CI.


yookoala:
Note: The count of lines with `$guid` reference went down from 13503 to 8071 :-)

rossdotparker:
Wow, this is a monster PR, but long overdue and put together with a really elegant solution. Thanks for leading on this. I am going to merge now. Cheers!

PR #709, opened 2018-11-15
Gettext function

### LINKED ISSUE #701: Improve translation API by named parameter replacement

yookoala:
**New Feature**
## Description
* Updated `__()` and `Gibbon\Locale::translate` to accept `$params` for named parameter replacements.
* Changed the 3rd parameter to be an `$options` array. `__()` is made backward compatible to accept `string` as the text domain.
* Totally backward compatible.
* Added `__n()` and `Gibbon\Locale::translateN` for plural string translation (mimicing [ngettext()](http://php.net/manual/en/function.ngettext.php) of the core gettext extension).
* Resolves #701.

## Motivation and Context
* See #701.

## How Has This Been Tested?
* Locally.
* Travis CI.


SKuipers:
Koala, you're an open source hero! 🌟 Thanks for all the awesome work you've been doing.

What do you think of adding some unit tests too, to cover both the new method signature in `Gibbon\Locale`, and also the variety of backwards compatibility being supported in `__()`

Thanks!

yookoala:
I actually decided to do so. But the `Gibbon\Locale` class is so coupled with container, `Gibbon\Session` and `PDO`. I'd need more time to figure out how best to do the unit test.

SKuipers:
Ah yes, you're right 😔 That class predated the container (all the ones in src/Gibbon did) so it's not as decoupled as it should be.

yookoala:
* Refactored the `Locale::__construct` and its use in `CoreServiceProvider`.
* Added unit test to the translation functions.

yookoala:
Question: Do we need specific format for the named replacement keys?

For example,
```php
__(`'Hello ${name} (${link})'`, [
  'name' => htmlspecialchars('Alice'),
  'link' => '<a href="https://twitter.com/alice">link</a>',
])
```
or

Drupal style

```php
__(`'Hello @name (!link)'`, [
  '@name' => 'Alice', // will be processed with htmlspecialchars to remove.
  '!link' => '<a href="https://twitter.com/alice">link</a>', // will not be processed.
])
```

tests/unit/Gibbon/LocaleTest.php
```diff
@@ -0,0 +1,176 @@
...
+        $this->mockSession = $this->createMock(session::class);
+        $this->mockSession
+            ->method('get')
+            ->willReturn(null); // always return null
+
+        // mocked locale object
+        $locale = new Locale(__DIR__ . '/mock', $this->mockSession);
+
+        $i18ncode = 'zh_TW';
+        putenv('LANGUAGE=');
```
SKuipers: Hmm, this test didn't run successfully on my system until I removed these lines and used the Locale classes set locale method (which accounts for some variation in the way locales are set from system to system):

`$locale->setLocale('zh_TW');`

yookoala: You need to do `sudo locale-gen zh_TW.utf8` to make he test work on your system (assuming it is linux)

SKuipers: Yep, he locale is absolutely there & properly configured. The setLocale method on the Locale class includes the variations for setting a locale across different OSs, which can differ slightly. If you replace your `putenv` and `setlocale` lines with the `$locale->setLocale('zh_TW');` it will account for this :)

SKuipers:
I've added an extra set of tests for `__()`, to help cover our various function signatures. Just a basic implementation, and uses the existing translations.

tests/unit/Gibbon/LocaleTest.php
yookoala: Using `Locale::setLocale` now. Seems OK.

SKuipers:
Looks like you're still tinkering with this one. Give a shout when it's ready to review. Thanks!

yookoala:
I have changed the placeholder syntax according to our discussions in #701.

I'll also implement the `__n()` function today, if possible.

After that, I think this PR will be ready.

yookoala:
@SKuipers The code is ready for merge. Please check if the changes and the tests match what we discussed in #701.

functions.php
```diff
@@ -55,27 +55,82 @@ function emailBodyConvert($body)
...
+ * @param int    $n       The number (e.g. item count) to determine
+ *                        the translation for the respective grammatical
+ *                        number.
+ * @param array  $params  Assoc array of key value pairs for named
+ *                        string replacement.
+ * @param array  $options Options for translations (e.g. domain).
+ *
+ * @return string Translated Text
+ */
+function __n(string $msgid1, string $msgid2, int $n, array $params = [], array $options = [])
```
SKuipers: Looking good. Could we use $singular, $plural and $count as the variable names here, which are more descriptive (especially for editors that have code completion). I'm always a fan of readable variable names :)

```suggestion
function __n(string $singular, string $plural, int $count, array $params = [], array $options = [])
```

src/Gibbon/Locale.php
```diff
@@ -164,60 +171,123 @@ public function setStringReplacementList(Connection $pdo, $forceRefresh = false)
...
+     * @param int    $n       The number (e.g. item count) to determine
+     *                        the translation for the respective grammatical
+     *                        number.
+     * @param array  $params  Assoc array of key value pairs for named
+     *                        string replacement.
+     * @param array  $options Options for translations (e.g. domain).
+     *
+     * @return string Translated Text
+     */
+    public function translateN(string $msgid1, string $msgid2, int $n, array $params = [], array $options = [])
```
SKuipers:
```suggestion
    public function translateN(string $singular, string $plural, int $count, array $params = [], array $options = [])
```

Similar change + needs the corresponding code changes.

src/Gibbon/Locale.php
```diff
@@ -164,60 +171,123 @@ public function setStringReplacementList(Connection $pdo, $forceRefresh = false)
...
+            dngettext($domain, $msgid1, $msgid2, $n);
+
+        // apply named replacement parameters, if presents.
+        $text = strtr($text, array_reduce(array_keys($params), function ($carry, $key) use ($params) {
+            $carry['{' . $key . '}'] = $params[$key]; // apply quote to the keys for replacement
+            return $carry;
+        }, []));
+
+        // apply custom string replacement logic.
+        if (isset($this->stringReplacements) && is_array($this->stringReplacements)) {
```
SKuipers: This chunk is repeated in both methods, perhaps a good candidate for a reusable `protected doStringReplacements($text)` method?

yookoala:
Oh, `Gibbon\Session` is currently not an implementation of `Gibbon\Contracts\Services\Session`.

Will revert the last commit.

yookoala:
Ready to merge now.

SKuipers:
> Oh, `Gibbon\Session` is currently not an implementation of `Gibbon\Contracts\Services\Session`.
> 
> Will revert the last commit.

Oh, it should be! 😆

yookoala:
> > Oh, `Gibbon\Session` is currently not an implementation of `Gibbon\Contracts\Services\Session`.
> > Will revert the last commit.
> 
> Oh, it should be! laughing

Fixed this in the latest commit.

src/Gibbon/Session.php
```diff
@@ -29,7 +29,7 @@
  * @version	v13
  * @since	v12
  */
-class Session
+class Session implements SessionInterface
```
SKuipers: 😏 Thanks!

PR #720, opened 2018-11-21
Timetable: refactor breadcrumbs

yookoala:
**Refactor**
## Description
adapt the new `$page` global for breadcrumb.

## Motivation and Context
See #696

## How Has This Been Tested?
* Locally
* Travis

### DIFF
--- a/modules/Timetable/studentEnrolment_manage_edit.php
+++ b/modules/Timetable/studentEnrolment_manage_edit.php
@@ -19,6 +19,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use Gibbon\Forms\Form;
 use Gibbon\Forms\Prefab\BulkActionForm;
+use Gibbon\Services\Format;
 
 if (isActionAccessible($guid, $connection2, '/modules/Timetable/studentEnrolment_manage_edit.php') == false) {
     //Acess denied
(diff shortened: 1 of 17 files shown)


modules/Timetable/studentEnrolment_manage_edit.php
```diff
@@ -50,9 +51,13 @@
         } else {
             //Let's go!
             $values = $result->fetch();
-            echo "<div class='trail'>";
-            echo "<div class='trailHead'><a href='".$_SESSION[$guid]['absoluteURL']."'>".__('Home')."</a> > <a href='".$_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/'.getModuleName($_GET['q']).'/'.getModuleEntry($_GET['q'], $connection2, $guid)."'>".__(getModuleName($_GET['q']))."</a> > <a href='".$_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/'.getModuleName($_GET['q'])."/studentEnrolment_manage.php'>".__('Manage Student Enrolment')."</a> > </div><div class='trailEnd'>".sprintf(__('Edit %1$s.%2$s Enrolment'), $values['courseNameShort'], $values['name']).'</div>';
-            echo '</div>';
+
+            $page->breadcrumbs
+                ->add(__('Manage Student Enrolment'), 'studentEnrolment_manage.php')
+                ->add(strtr(__('Edit :courseNameShort.:name Enrolment'), [
```
SKuipers: I see that you're looking to use the named placeholders here, which from a programming perspective makes sense. But I don't think we should be going back and editing existing translations. In this case, it already uses numeric placeholders, and has already been translated into many languages: changing it now will create a whole new string. I think we have enough strings in the old sprintf style that we need to keep them and just focus on improving new strings going forward. 

In this case, we could still use `%1$s` and `%2$s` as the keys for the updated `__()` function, right?

SKuipers:
I tweaked Locale a _tiny_ bit so that we can still use the __() function for the old style placeholders, rather than sprintf. Does that seem like an okay change? https://github.com/GibbonEdu/core/commit/4f8b085a3ff26303e8e4bb118ee4597c916c3d06

PR #732, opened 2018-11-27
Students: ooification of New Students and Left Students reports 

SKuipers:
Another two tables done, making use of the added templates for family details.


rossdotparker:
Looking good, just one tweak, which I've pushed directly:

https://github.com/GibbonEdu/core/commit/51c6a8a105fe980e7148adbc3d96fd57f81fe69b

PR #733, opened 2018-11-27
Doctrine DBAL Discussion Conversion... (Mk II)  & Security Robot IP Check at Login

crayner:
### Security Robot IP Check at Login
This feature is design to identify machines by IP address that have attempted more that 3 login (failures) within the last 20 minutes and ignore that IP Address for a maximum of 20 minutes.  

The system Admin is able to whilte list IP addresses on the System Admin Setting page (Misc Section) using unique IP, subnet or partial IP addresses in a standard comma separated list. e.g  127.0.0.1, 172.16.0 and 192.168.1.1/24 are all valid.  Invalid entries are ignored.

127.0.0.1 matches 127.0.0.1
172.16.0 matches 172.16.0.x
192.168.1.1.24 matches 192.168.1.x   (this does a bit wise test, so 192.168.1.1/28 would match 16 machines. (Actually 15, due to match all requests on the network.)
______________________

I do not expect this to be merged, unless you wish to proceed with the refactor in full.
### New Feature
This is a single module refactor to use Doctrine DBAL (Database Abstraction Layer) The module is the Single page RollGroups at '/modules/Roll Groups/rollGroups.php' On the menu system, People=>Roll Groups.

### Description
Created managers to generate a Doctrine Connection, and EntityManager. This uses 'symfony/orm-pack', and a purpose built manager to provide the connection from the existing config.php for backwards compatibility.

### Motivation and Context
Allow Gibbon to move from MySQL to other server structures, such as PostgreSQL, and even local file system.
Simplify updating of entity (database) metadata.
Simplification of SQL in a majority of locations.
How Has This Been Tested?
Local Testing
Travis cl

Screenshots (if appropriate):
The render is not changed...


crayner:
Tested Locally and Travis

crayner:
Closed to separate into two pull requests

PR #738, opened 2018-11-29
System: add a function to help with module translations

SKuipers:
While translating additional modules outside the core, it becomes a bit clunky to include the module name for each string, especially with it now being an array parameter. Since the system already knows the name of the current module for any given page, it'd be pretty easy to include this automatically.

With this change, the following:
`__('Import From File', [], ['domain' => 'Data Admin'])`

is identical to:
`__d('Import From File')`

This can also help to ensure module strings are not accidentally included in the core, as the xgettext script won't be looking at `__d()` functions.

@yookoala What do you think?

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -133,6 +133,26 @@ function __n(string $singular, string $plural, int $n, array $params = [], array
     return $gibbon->locale->translateN($singular, $plural, $n, $params, $options);
 }
 
+/**
+ * Identical to __() but automatically includes the current module as the text domain.
+ *
+ * @see __()
+ * @param string $text
+ * @param array  $params
+ * @param array  $options
+ * @return string
+ */
+function __d(string $text, array $params = [], array $options = [])
+{
+    global $gibbon;
+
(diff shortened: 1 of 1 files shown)


yookoala:
I understand the need, but I'd prefer the letter `m` (i.e. module) to `d` (i.e. domain). Its better describing what it does.

SKuipers:
Good point, I wondered about that too. 

Does that still leave enough differentiation between `__n()` and `__m()`? Code-readability wise.

yookoala:
That bugs me a bit.

Another option would be to have a Gibbon\Module::__ function to distinct them (instead of plain function).

SKuipers:
I've tweaked the PR to change the name to `__m()`. The goal is brevity, so I think `Gibbon\Module::__` might still be too long, but I agree that m as a letter may be more descriptive than d. With code completion and a different function signature, the difference between `__n()` and `__m()` is hopefully clear enough.

PR #740, opened 2018-11-29
Modules: remove uses of getAge

yookoala:
**Refactor**
## Description
* remove all usage of the function `getAge()` in the codebase.

## Motivation and Context
There is only a few usage left of `getAge()` in the core. Removing them helps to lift dependencies from it and make it ready to be removed.

## How Has This Been Tested?
* Locally
* Travis


SKuipers:
Pushed a quick change post-merge. Format::age takes a Y-m-d date, so no need to convert to timestamp and back again. https://github.com/GibbonEdu/core/commit/a43a514251ca0882389c98aa296838576a32c992

PR #760, opened 2018-12-14
Messenger: fixed missing Message Wall targets for Transport and Group

SKuipers:
A fix, but nothing elegant, I've just copied your SQL and added the appropriate targets 😬 It tests out though.


rossdotparker:
Thanks for fixing this Sandra. I think this file contains some of my least favourite code in all of Gibbon...and that is saying something!

crayner:
Ha!😂😂😂
Grace and Peace

Craig Rayner
craig@craigrayner.com
Mobile: +61 456 560 018

PR #766, opened 2018-12-19
Render tables - Support to internationalization

leofreitas:
## Motivation and Context
Add support for internationalization for content columns, for example:
schoolYear_manage.php
gradeScales_manage.php
attendanceSettings.php
fileExtensions_manage.php

## How Has This Been Tested?
Local and Travis

### DIFF
--- a/src/Tables/Columns/Column.php
+++ b/src/Tables/Columns/Column.php
@@ -277,7 +277,7 @@ class Column
         if ($this->hasFormatter()) {
             return call_user_func($this->formatter, $data);
         } else {
-            return isset($data[$this->getID()])? $data[$this->getID()] : '';
+            return isset($data[$this->getID()])? __($data[$this->getID()]) : '';
         }
     }
 }


rossdotparker:
@SKuipers can you foresee any cases where this backfires?

SKuipers:
Hmm, I think there could be, if partial string replacements were applied to user-entered text where it was not intended. I can see where it's needed though for some fields like enums being pulled from the database. The way to handle this could be a translatable flag/method for columns that require the translation wrapper, and update the columns where this is applicable.

rossdotparker:
Do you see that as something for v17 or v18? I'd lean towards the latter at this point I think.

SKuipers:
Hi Leo,

We're back again, and since it's v18 not it's be great to get this issue sorted. I'm thinking we may need to flag a Column in the DataTable as `translatable()`. This way, they're not all translated by default, but the funcionality would be there for the columns that need it (such as Enums coming from the DB). What do you think? Want to try your hand at that change? 😃

leofreitas:
Hi Sandra,
That's great! I'll start working on it.

crayner:
Hi Leonardo,'
   This is only an idea, and you do not have to use it.
   One of the frameworks I use often has a flag on the Select and Radio
Group called translate_choice.  If translate choice is false, then no
translation is done, or if otherwise the prompt is translated.

So

$row->addSelect('gibbonYearGroupID')->fromArray(array('all' => __('All')))->
fromQuery($pdo, $sql)->selectMultiple()->selected($yearGroups);
may look like
$row->addSelect('gibbonYearGroupID')->fromArray(array('all' => __('All')))->
fromQuery($pdo, $sql)->selectMultiple()->translateChoice(true)->selected(
$yearGroups);
The renderer would then be responsible to translate the prompts if the
translate_choice is not false.
Grace and Peace

Craig Rayner
craig@craigrayner.com
Mobile: +61 456 560 018

leofreitas:
Hi Crayner,
Thanks for sharing. I think that's a good idea! :wave:

leofreitas:
Continues in #785

PR #768, opened 2019-01-08
Incorrect Error Message Logic in Attendance Course Class

crayner:
The logic of the error message states that the date SHOULD BE IN THE FUTURE, but the TEST does the opposite.  The test is correct.

Bug Fix

### DIFF
--- a/modules/Attendance/attendance_take_byCourseClass.php
+++ b/modules/Attendance/attendance_take_byCourseClass.php
@@ -36,7 +36,7 @@ if (isActionAccessible($guid, $connection2, "/modules/Attendance/attendance_take
 } else {
     //Proceed!
     if (isset($_GET['return'])) {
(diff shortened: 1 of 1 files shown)


SKuipers:
Looks like the same error message is used in a few other instances, and was incorrect as well (& missing translation tags). Fixed these here: https://github.com/GibbonEdu/core/commit/3c79e4c97e01d82a3af6883120ab7929ace1fa7b https://github.com/GibbonEdu/core/commit/aa064a759c1a9eccab2f2429341546d5a407863d

crayner:
Thanks for that.  I had seen the other copies, but had NOT investigated
usage, and I was aware that some future stuff was done, and therefore the
original message may have been valid.
Grace and Peace

Craig Rayner
craig@craigrayner.com
Mobile: +61 456 560 018

PR #771, opened 2019-01-18
ActivitiesFix warning in function sprintf

leofreitas:
## Motivation and Context
Fix - When I edit an activity: Warning in function sprintf on line 22 

## How Has This Been Tested?
Local
Travis


SKuipers:
Well spotted! Is this a PHP7.3 only error? I don't think sprintf has ever complained about type-strictness before 🤔 I ran into some twig compatibility issues with PHP7.3 running on nginx too, so we may recomment PHP7.0-7.2 for production for now, and keep 7.3 for development as we find and fix any issues.

leofreitas:
So, i think this is the best way.

PR #773, opened 2019-01-21
V18 Database Inconsistencies

crayner:
This is a fix for table identifiers using different size parameters in calling tables.


crayner:
This is only failing on Travis, not on phpunit tests conducted locally.  Unfortunately Codeception fails ALL tests on my machine, due to to setup issues I believe.  
   `
cURL error 7: Failed to connect to 127.0.0.1 port 8888: Connection refused (see http://curl.haxx.se/libcurl/c/libcurl-errors.html)
`

for all tests.

Continuing to work on that issue, and then the reason for the test failure.

SKuipers:
Hi Craig,

Theses database changes look good, and will certainly help to keep the database consistent, I'd be happy to test them out. It looks like a number of composer-related changes have made their way into this PR, are you able to reset/rebase your code so that only the CHANGEDB and CHANGELOG files are included.

Thanks!

crayner:
I have rebased the vendor directory from the upstream/v18.0.00.  All of the composer related changes are removed, except the **composer.lock** file.  I have asked that this be added to the general .gitignore file as it is a local reference file for composer.

SKuipers:
Thanks for making these changes! Good question. In this case, I think the composer.lock file is an important piece to keep in source control. Particularly because we also commit the `/vendor` folder to git, the lock file helps provide consistency across developer codebases. For example, the composer.lock changes in this PR indicate different dependency versions than is what is installed in `vendor`. A couple small differences might not cause an issue right now, but without the lock file developers could over time end up with a widely different set of files than what is used in the core.

crayner:
Thanks Sandra. My standard operating process is to always do a composer update when I pull the files from the master copy, to ensure the latest version currently listed in the composer.json file are correctly installed, which always modifies the composer.lock file.  I will deal this this in my own processes.

PR #774, opened 2019-01-21
V18 Password Management

crayner:
If accepted, his pull request moves the password storage management towards industry encryption standards, allowing the system to select the highest level of encryption available to the current system on which the software is running.  So MD5, to SHA256, to BCrypt and finally Argon2i.  At the moment that should be BCrypt, but SHA256 is currently the Gibbon Standard

BCrypt and Argon2i code are currently not in this software package, although PHP 7.0 supports BCrypt natively.  PHP 7+ will support Argon2i with libsodium installed, but PHP 7.2+ supports Argon2i natively.

I will be happy to add the appropriate code to handle both forward and backward compatibility should this pull be accepted.   It also meets the requirement to separate logic and display code, as much as is possible with the current container infrastructure.

I do hope that you consider this as I do have serious concerns about password security should a school find itself exposed by hacking of their site.  

**Connection Manager.**
I have added some functionality to the src/Database/Connection.php class to better mirror the code with the various pages and maintain backwards compatibility.  The major key here is an ignoreError key which is set to a default of false, but if set to true mirror the 
```
try{
    // do something 
} catch PDOException $e) {}
```
that appears often in much of the code. 

The page code all but disappears into the LoginSupervisor class for both the Login page and the preferencesPasswordProcess page.

As usual, please let me know of any errors, and also question about why I think this is SO IMPORTANT to the future of Gibbon.


SKuipers:
I'm going to start sounding like a broken record 😂  but would would be able to split the PasswordEncoder changes from the introduction of the LoginSupervisor? I think refactoring the authentication process could be helpful on the whole, but its also a pretty vital part of the system,  it's worth taking time to carefully test and consider the changes,and how they fit into the upcoming refactoring too. 

I think the PasswordEncoder could certainly be merged in first, which would give a great chance to test part that in production and possibly even add some unit tests.

PR #775, opened 2019-01-23
V18 logger integration

crayner:
During debugging of recent errors I found an issue actually seeing the cause of the issue with placing breaks in the code.  

### Features:
The logger uses a Factory to generate as many different logger instances as required for the page.  As this is a full object based implementation, the same object(s) is used throughout code execution.  An example of use is to generate the factory
```
$factory = new LoggerFactory(new SettingGateway($connection));
$logger = $factory->getLogger('{channel}');
```
The channel defaults to **'gibbon'**.  To add a message to the logger use:
```
$logger->{levelName}(string $message, {array $content});
```
This will add messages to the file called {channel}-{date}.log in the log directory.  The logger also rotates the files, keeping the last 7 files only.
The logger also looks at the installation type, and in production mode only logs WARNING and above.

### Requirements:
- monolog/monolog was added to the core requirements.  It appears to have been a sub requirement of another package, so already in vendor, but not in composer.json
- creation on {document_root}/var/log directory to handle logs.
- modification of ,gitignore to ignore *.log files.

An an example this PR modifies the Connection Class to capture all error generated in the PDO Engine.  This would generate a log entry for any installation type, as ERROR is >= WARNING. I would see this as a development tool and a production tool, helping in code failure analyses, and logging details for security (yes I know some of this is done in MySQL., but this would be more verbose.)  Eventually this logger could also manage these log writes as well, and due to cascade ability, also would allow duplication to both SQL and file.  In the example, I have used the translator to set the message. 

The log result is in standard log format. e.g. 
```
[2019-01-23 18:48:47] mysql.INFO: I am a MySQL Log Message. [] []
[2019-01-23 18:48:48] mysql.INFO: I am a MySQL Log Message. [] []
[2019-01-23 18:48:58] mysql.INFO: I am a MySQL Log Message. [] []
[2019-01-23 18:49:08] mysql.INFO: I am a MySQL Log Message. [] []
[2019-01-23 18:49:18] mysql.INFO: I am a MySQL Log Message. [] []
```

Please let me know of any issues.  This passed the local Codeception tests and phpunit on my machine.
Looking forward to your response.
Craig


crayner:
All sound good.  Where can I find details on DI Container usage within Gibbon?

crayner:
So this now defines two loggers in the container, called gibbon_logger for general use, and mysql_logger for mysql capture.  Both are identical at the moment.  Let me know if I need to modify the container usage, as I have not used the raw PHPLeague container before now.

src/Services/LoggerFactory.php
```diff
@@ -0,0 +1,118 @@
...
+
+    /**
+     * @param string $channel
+     * @return mixed
+     */
+    public function getLogger(string $channel = 'gibbon')
+
+    {
+        if (isset($this->loggerStack['gibbon']))
+            return $this->loggerStack['gibbon'];
```
SKuipers:
```suggestion
            return $this->loggerStack[$channel];
```

src/Services/LoggerFactory.php
```diff
@@ -0,0 +1,118 @@
...
+    private $loggerStack;
+
+    /**
+     * @param string $channel
+     * @return mixed
+     */
+    public function getLogger(string $channel = 'gibbon')
+
+    {
+        if (isset($this->loggerStack['gibbon']))
```
SKuipers:
```suggestion
        if (isset($this->loggerStack[$channel]))
```

src/Services/LoggerFactory.php
SKuipers:
```suggestion
            return $this->loggerStack[$channel];
```

src/Services/LoggerFactory.php
SKuipers: I suspect these are intended to use the `$channel` variable, with 'gibbon' as the default. I've tagged it as a code suggestion, so GitHub can auto-accept and merge these changes for you.

crayner:
The Monolog system allows for Handler chaining, so messages can be handled by multiple files, rejected or written as the per the handler rules,  This is currently hard coded in the factory, but could be easily expanded for use in the container call,

PR #777, opened 2019-01-24
Repair getWeekNumber

crayner:
The getWeekNumber() of functions was caught in a never ending loop.


crayner:
my bad
Grace and Peace

Craig Rayner
craig@craigrayner.com
Mobile: +61 456 560 018

PR #781, opened 2019-01-29
Logger Stacked Handler Exmple

crayner:
This PR demonstrates logging to two separate locations.  

Tested Locally using Connection Class.

As Gibbon uses multiple connection strategies for the database, not all transactions are recorded.  

As Gibbon requires the container to be manually configured, it occurs to me that the factory should be used to handle channel flexibility.

Note that handler rejects Logger::DEBUG messages in a production site (not logged) but in a testing or development site, logging of Logger::DEBUG level is accepted.  See example output below.  The channel name is recorded in the gibbon log and the mysql log.

```
[2019-01-29 12:52:35] mysql.DEBUG: SELECT * FROM gibbonModule WHERE name=:moduleName AND active='Y' {"moduleName":"Attendance"} []
[2019-01-29 12:52:35] mysql.DEBUG: SELECT * FROM gibbonTheme WHERE active='Y' [] []
[2019-01-29 12:52:35] mysql.DEBUG: SELECT gibbonModule.category, gibbonModule.name, gibbonModule.type, gibbonModule.entryURL, gibbonAction.entryURL as alternateEntryURL, (CASE WHEN gibbonModule.type <> 'Core' THEN gibbonModule.name ELSE NULL END) as textDomain                 FROM gibbonModule                  JOIN gibbonAction ON (gibbonAction.gibbonModuleID=gibbonModule.gibbonModuleID)                  JOIN gibbonPermission ON (gibbonPermission.gibbonActionID=gibbonAction.gibbonActionID)                  WHERE gibbonModule.active='Y'                  AND gibbonAction.menuShow='Y'                  AND gibbonPermission.gibbonRoleID=:gibbonRoleID                  GROUP BY gibbonModule.name                  ORDER BY FIND_IN_SET(gibbonModule.category, :menuOrder), gibbonModule.category, gibbonModule.name, gibbonAction.name {"gibbonRoleID":"001","menuOrder":"Admin,Assess,Learn,People,Other"} []
```


src/Services/LoggerFactory.php
```diff
@@ -79,7 +91,8 @@ public function __construct(SettingGateway $settingGateway)
     {
         $this->settingGateway = $settingGateway;
         $this->filePath = $settingGateway->getSettingByScope('System', 'absolutePath') . DIRECTORY_SEPARATOR . 'var' . DIRECTORY_SEPARATOR . 'log' . DIRECTORY_SEPARATOR;
-        $this->loggerLevel = $settingGateway->getSettingByScope('System', 'installType') === 'Production' ? Logger::WARNING : Logger::DEBUG;
+        $this->installType = $settingGateway->getSettingByScope('System', 'installType');
```
SKuipers: I wonder if the SettingGateway dependency used in this class creates a tight coupling between it and the LoggerFactory. Another approach could be to use the DI container to pass in/set only the necessary configuration when the factory is constructed.

crayner:
Decoupled SettingGateway from LoggerFactory and injected required settings directly in Container builder.

crayner:
Is there a problem with this PR that requires my attention?

PR #786, opened 2019-02-04
gitlab-ci: fix composer installation

yookoala:
* update composer installer's SHA hash check.


yookoala:
Since the upstream compser installer updated recently, the hash of it changed. It has nothing to do with what we did.

I'm considering to make the hash check optional. Probably only check if environment variable `COMPOSER_INSTALLER_CHECKSUM` exists.

PR #787, opened 2019-02-05
Improvement in support of internationalization

leofreitas:
## Motivation and context
Improvement in support of internationalization using new method isTranlatable

## How Has This Been Tested?
Local and Travis


SKuipers:
Hi Leo,

As Craig pointed out on Slack, it may be worth re-naming this method for overall codebase consistency, so that it doesn't conflict with the similar case of using an `isSomething()` method for boolean getters. In this case, a method name like `translatable()` would fit with similar `sortable()` and `notSortable()` methods in this class. Could you make this change?

There's another similar conflict for the `isRequired()` method on the Form class, I'll see about updating that one too. These classes are aimed to be fluent method names, so they're easily readable chained together (like the builder pattern), so I don't think we need to go as far as adding `set` in front of each of the setter names.

Thanks!!

crayner:
Great solution.
Grace and Peace

Craig Rayner
craig@craigrayner.com
Mobile: +61 456 560 018

leofreitas:
Hi Sandra and Craig,
I made the changes. I think that way is better. 
Best regards!

modules/Attendance/report_graph_byType.php
```diff
@@ -110,14 +110,22 @@ function getDateRange($guid, $connection2, $first, $last, $step = '+1 day', $out
         $row->addDate('dateEnd')->setValue(dateConvertBack($guid, $dateEnd))->isRequired();
 
     $typeOptions = array_column($attendance->getAttendanceTypes(), 'name');
+    $arrayTypeOptions = array();
+    foreach($typeOptions as $types){
```
SKuipers: Another option here is to use `$typeOptions = array_map('__', $typeOptions);`. It's useful when you have one function you'd like to apply to the whole array.

PR #794, opened 2019-02-12
V18 travis

crayner:
Yes, it is a fix for PHP 7.2, whilst still working for PHP 7.0 and PHP 7.1


SKuipers:
Hi Craig,

Looks like a good fix for Travis. I've noticed a number of version changes in the composer.lock file that appear unrelated to the phpunit change: google/apiclient-services, guzzlehttp/psr7, phpmailer/phpmailer, etc. Could it be that you've made changes to your local vendor folder that have also been bundled into this PR?

I realize there are ongoing discussions about how best to handle the vendor folder, but for the time being with the vendor contents currently being tracked in version control, it's important for the composer.lock file to be an exact representation of the versions of libraries installed. This will help ensure consistency between developers and prevent future problems.

crayner:
Unfortunately the changes in the composer.lock are intrinsic to the update.  As such, I have taken some time to do a full 'composer update  --no-dev'  after a 'composer update'  This results in the following updates to packages, plus one installation.  Some of these are requirements of other packages.


  - Updating tecnickcom/tcpdf (6.0.038 => 6.2.26): Loading from cache
  - Updating symfony/polyfill-ctype (v1.9.0 => v1.10.0): Loading from cache
  - Updating symfony/polyfill-mbstring (v1.9.0 => v1.10.0): Loading from cache
  - Updating twig/twig (v2.5.0 => v2.6.2): Loading from cache
  - Updating slim/slim (3.11.0 => 3.12.0): Loading from cache
  - Updating phpmailer/phpmailer (v5.2.23 => v5.2.27): Loading from cache
  - Updating google/apiclient-services (v0.74 => v0.83): Loading from cache
  - Installing ralouphie/getallheaders (2.0.5): Loading from cache
  - Updating guzzlehttp/psr7 (1.4.2 => 1.5.2): Loading from cache
  - Updating phpseclib/phpseclib (2.0.12 => 2.0.14): Loading from cache
  - Updating psr/log (1.0.2 => 1.1.0): Loading from cache


The process used was:

deleted composer.lock
git checkout upstream/v18.0.00 -- composer.lock
Push and failed Travis.
composer update
composer update --no-dev
push and passed travis

Should you wish to stop these upgrades, the correct way to do so is to lock the version in composer.json

SKuipers:
I certainly understand what you mean, and appreciate you looking into a fix for the CI process.

This seems to be a case of a mole-hill (a single fix for an external CI service) becoming a mountain (updating all of Gibbons dependencies). It looks like the same fix can be made with 5 files changed rather than 1072:

https://github.com/SKuipers/core/commit/e8f94746bac2763042812992ad1b1c8cc9c3588f
https://travis-ci.org/SKuipers/core/builds/493576366

The process used was:

git checkout upstream/v18.0.00
composer require --dev --update-no-dev "phpunit/phpunit:6.5.14"
push and pass CI tests

Re:
> Should you wish to stop these upgrades, the correct way to do so is to lock the version in composer.json

Respectfully, the .lock in the composer.lock file is [the exact reason this file exists](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control).

crayner:
Thanks.  I will leave this one with you.

crayner:
Respectfully https://getcomposer.org/doc/articles/versions.md

SKuipers:
Hi Craig,

In this community, you’ll find we place a strong emphasis on learning, collaboration, and inclusivity. This isn’t lip service: it’s at the core of who we are as individuals and educators. Sharing ideas is part of the process: as developers we’re only growing and improving if we’re open to learning new things. I think you’ll find choosing to focus on language that is competitive and confrontational will continuously put you at odds with our team.

As all the tests in the CI process are currently passing, it sounds like you’ve been working on a new change and encountered a problem. One way to approach this constructively might be “When I try to do X I’m having a problem with Y, I suggest that we make changes to Z”.

crayner:
Hi Sandra,
   Thanks for your reply.  It appears that you have misinterpreted my actions and intent.  Please let me take to time to correct your interpretation.
   
   1. I was not trying to be confrontational.  I believe that you have taken it this way, but it was not my intent.    
   2. This was never about the need for my fix to be included in Gibbon.  The fix is already done, and it used the suggestions that I made available through this PR.  Thank you.  I only reopened to comment.
   3. I took the time to read the documentation that you suggested, and realised that you do not have a correct operational awareness of how composer works.  I replied in the same tone that you replied to me, respectfully.
   4. My intent was that you learn (one of your performance keys for the group) about composer, and correct your inaccurate ideas about the reason for the composer.lock and composer.json.  Your answer appeared to say that you edit the composer.lock file to ensure package/version references.  This is not the case.  The composer.json holds these references.  The lock file is a reference file used by composer itself, acting a lock to package/version already installed, thus speeding up any composer commands.

PR #796, opened 2019-02-22
System: update date and time localization in Format class

SKuipers:
Updates the `dateReadable`, `dateTimeReadable` and `dateRangeReadable` methods to use `strftime`, which handles localization of date strings: http://php.net/manual/en/function.strftime.php 

Also adds a `relativeTime` method, which aims to provide an easy-to-use, translatable way to display simple time strings like 1 hr ago, 3 mins from now, etc.

@leofreitas Let me know what you think. We should be able to use something like `Format::dateReadable($date, '%b');` to properly handle the date strings in `modules/Attendence/attendence.php`. By using this class, in the future we should be able to enable more customization of preferred date formats for admins, by pulling in the system-wide preferred date formats from settings.

### DIFF
--- a/src/Services/Format.php
+++ b/src/Services/Format.php
@@ -188,6 +188,53 @@ class Format
     }
 
     /**
+     * Formats a Date or DateTime string relative to the current time. Eg: 1 hr ago, 3 mins from now.
+     *
+     * @param string $dateString
+     * @return string
+     */
+    public static function relativeTime($dateString, $tooltip = true)
+    {
+        if (strlen($dateString) == 10) $dateString .= ' 00:00:00';
+        $date = DateTime::createFromFormat('Y-m-d H:i:s', $dateString);
+
+        $timeDifference = time() - $date->format('U');
+        $seconds = abs($timeDifference);
+
+        switch ($seconds) {
+            case ($seconds < 60):
+                $time = __('Less than 1 min');
+                break;
+            case ($seconds >= 60 && $seconds < 3600):
+                $minutes = floor($seconds / 60);
+                $time = __n('{count} min', '{count} mins', $minutes);
+                break;
+            case ($seconds >= 3600 && $seconds < 86400):
+                $hours = floor($seconds / 3600);
+                $time = __n('{count} hr', '{count} hrs', $hours);
+                break;
+            case ($seconds >= 86400 && $seconds < 2419200):
+                $days = floor($seconds / 86400);
+                $time = __n('{count} day', '{count} days', $days);
+                break;
+            default:
+                return self::tooltip($date->format(
+                    strlen($dateString) == 10
+                        ? static::$settings['dateFormatPHP']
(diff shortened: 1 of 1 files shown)


rossdotparker:
I've had a look at this, and I can't see any immediate issues that jump to mind. I've not tested it, though. @SKuipers I'll approve this for merging, and you feel free to merge once you've tested to your satisfaction.

leofreitas:
Hi @SKuipers, I think that we could use timestamp rather than datetime as parameter. What do you think?

SKuipers:
Hmm, a fair point. I think the Format class currently tends to focus on strings because in most of our dates and times come from the database in sql Date and Date+Time formats. There are some cases in the codebase where timestamps are being used (particularly in the timetable rendering). Although ideally once they're refactored, these would best be handled with PHP's DateTime and DateInterval classes 🤔  I wonder if updating the date methods in the Format class to handle either strings or DateTime objects would help bridge this gap? As both strings and timestamps can easily be converted into DateTime objects.

yookoala:
I agree that all date and time should be handled with PHP's `DateTime` and `DateInterval` class. They are more flexible to use in general (than string and timestamp). Rewriting `Format` to handle both string and `DateTime` would be a nice transitional change.

SKuipers:
Okay, sounds good 👍  I'll work on some DateTime changes to the Format class and push them to this PR to review.

leofreitas:
Hi, I wrote wrong. I think that we could use timestamp rather than string in date format as parameter, for example:
`    public static function dateReadable($dateTimestamp, $format = '%b %e, %G')
    {
        return mb_convert_case(strftime($format, $dateTimestamp), MB_CASE_TITLE);
    }`
In file modules/Attendance/attendance.php (line 123) we should be able to use something like Format::dateReadable($day['currentTimeStamp'], '%b'). What do you think?

SKuipers:
The `DateTime` classes in PHP provide a great set of tools for handling dates. Ideally, as we refactor the codebase, I think we could aim to handle our date and time manipulation using these classes.

Right now however, nearly all of our dates are coming from the database as strings, so to support DateTime it makes sense for our methods to also accept strings. Luckily the [constructor for DateTime](http://php.net/manual/en/datetime.construct.php) is pretty flexible, so I've made some tweaks to the `Format` class to accept `DateTime` objects, as well as convert any strings into `DateTime` objects. This means we can pass in DateTime objects, Y-m-d and Y-m-d H:i:s strings from the database, as well as timestamps using the `@` prefix.

The modules/Attendance/attendance.php (line 123) could be written like:
`Format::dateReadable('@'.$day['currentDayTimestamp'], '%B')`;

For consistency in the codebase, I think this can help us to keep things clear & simple by ensuring the `Format` date and time methods each use a common method signature for the first parameter.

SKuipers:
Thanks for your code review! I've merged this in 👍

PR #801, opened 2019-02-26
Forms, Tables: update all instances of renamed setter methods

SKuipers:
This PR is the followup to the changes in https://github.com/GibbonEdu/core/pull/798, which updates the existing instances of the following methods in the core:

`isUnique` → `uniqueField`
`isDisabled` → `disabled`
`isRequired` → `required`
`isModal` → `modalWindow`
`isDirect` → `directLink`

Sorry it's a big PR 😬  a large number of files changed, but the scope is small: no other changes aside from these method renames.


rossdotparker:
Wow, that was eye-ball numbing stuff. All looks good. Will now test a few random screens locally. Also, updated the Form OOification cheat sheet.

PR #804, opened 2019-03-01
Forms, Tables: tweaks for usability, checkboxes, readonly select

SKuipers:
- Fixes an issue where tables with check all/none boxes weren't working if the table was nested in another table.
- Updates the `FormFactory` to allow simple tables as well as paginated data tables to be nested inside a Form
- Adds `chainedTo` and `chainedFrom` methods to `Date` inputs, allowing them to limit the range of dates for each other. Eg: once you select a Date Start, the Date End can't be earlier than the selected date. Uses a similar method signature as other chained inputs.
- Allows readonly select lists by reducing the available options to those already selected. This is useful in the case where the readonly-ness of the input is determined dynamically (based on permissions or other settings).
- Automatically sets the named `{count}` parameter for `__n()` function calls.
- Adds an option to pre-check all the boxes in a `CheckboxColumn`


rossdotparker:
I've scanned the changes, and they look good. Are there are areas of the system where I can see the solved issues in action before I update my production install, or are you solving for future cases? Just curious how best I might test these, without spending too long digging around for test cases. Thanks : )

SKuipers:
Ah yep, it's mostly tweaks for edge cases in some upcoming functionality 😄

PR #808, opened 2019-03-07
Applying the strings replacement syntax

leofreitas:
## Motivation and Context
Make the strings more readable for the translators and provide more flexibilities.

## How Has This Been Tested?
Local and Travis

### DIFF
--- a/modules/Crowd Assessment/moduleFunctions.php
+++ b/modules/Crowd Assessment/moduleFunctions.php
@@ -307,10 +307,10 @@ function getThread($guid, $connection2, $gibbonPlannerEntryHomeworkID, $parent, 
             $namePerson = __('{name} said', [
                 'name' => formatName($rowDiscuss['title'], $rowDiscuss['preferredName'], $rowDiscuss['surname'], $rowDiscuss['category'])
             ]);
-            $datetimePosted = __('Posted at <b>{hourPosted}</b> on <b>{datePosted}</b>',[
-                'hourPosted' => substr($rowDiscuss['timestamp'], 11, 5), 
-                'datePosted' => dateConvertBack($guid, substr($rowDiscuss['timestamp'], 0, 10))
-            ]);            
+            $datetimePosted = __('Posted at {hourPosted} on {datePosted}', [
+                'hourPosted' => '<b>'.substr($rowDiscuss['timestamp'], 11, 5).'</b>', 
+                'datePosted' => '<b>'.dateConvertBack($guid, substr($rowDiscuss['timestamp'], 0, 10)).'</b>'
+            ]);
             if ($level == 0) {
                 $classExtra = 'chatBoxFirst';
             }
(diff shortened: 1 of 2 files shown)


yookoala:
There is room for richer string replacement like:

```php
__('Posted by {name}.', [
  '!name' => 'Some name',
]);
```

So that we found `!` prefix, we quote the string with `<strong>`.

leofreitas:
> Thanks for putting this together! I wonder in cases where we have HTML tags in the strings, if they might be more useful if applied to the parameters rather than the translation string 🤔 eg: `Posted by {name}` and apply the `<b>` to the values. I know the current codebase is a mix of approaches, and there might be benefits and drawbacks for different cases. What do you think?

leofreitas:
Sorry devs,
I clicked the Close button accidentally and reopened.

leofreitas:
> Thanks for putting this together! I wonder in cases where we have HTML tags in the strings, if they might be more useful if applied to the parameters rather than the translation string 🤔 eg: `Posted by {name}` and apply the `<b>` to the values. I know the current codebase is a mix of approaches, and there might be benefits and drawbacks for different cases. What do you think?

I think that including tags HTML in translatable strings depends in the context. Other systems such as Moodle and Wordpress includes tags in your translation strings. If the tag is not separated from text , it's not a problem use it. In our case, the tag <i> is separated and not included in translatable string. eg: (modules/Planner/moduleFunctions.php line 249)

SKuipers:
Seems like a bit of a conundrum. Koala's suggestion seems useful, for basic markup like bold or italic. I wonder what the best way to handle more complex embedded html, like linked text? Often both the link's string and the url need to be dynamic, but keeping the HTMl context out of the translated string seems to make sense too.

yookoala:
We can always do this:

```php
$name = 'Some name';

__('Posted by {name}.', [
  'name' => "<strong>{$name}</strong>",
]);
```

or this,

```php
$id = 'some-user-id';
$name = 'Some name';

__('Posted by {name}.', [
  'name' => Gibbon\Services\Format::link('/user/' . $id, $name),
]);
```

The notation in my previous comment are simply short hand for convenient. It is not necessary for separating HTML from data and translation strings.

leofreitas:
I agree, is not essential separate HTML from data and translation strings.

SKuipers:
Yep, I can see both arguments here. I think it does make life easier on the translators if the string does not contain HTML, eg: `Posted by {name}` vs `Posted by <b>{name}</b>`. It gets more helpful to split the HTML out when the string gets longer and more complex too, and keeps the string re-usable for other output where HTML is perhaps not needed (PDFs, spreadsheet exports & other data views).

I like the examples you've posted Koala, to me they seem readable both from the code standpoint and for the translation string itself.

SKuipers:
Thanks for the PR Leo. I've updated it a bit to move the <b> tags out of the translation string, aside from that it looks great, going to merge it in 👍

leofreitas:
Thanks Sandra!

PR #811, opened 2019-03-19
Line 64 is generating a POE term of " ("

fvlasie:


SKuipers:
Hi FV. Thanks for your work on the translation issues 👍  In this case, it appears the `'( '` is outside the `__()` function and wouldn't have been picked up for translation (perhaps the code has since changed for that reference)? Your suggested update would break the array key in this instance, as `'name'` is the string used for the array key, rather than the string used to translate.

yookoala:
Gettext only extracts text from string arguments of all `__()` calls. It would not pick up the translation unless the string content of the array is otherwise generated in another `__()` call.

SKuipers:
Yep, looks like gibbonDaysOfWeek are part of the dbStringsForTranslation 👍

yookoala:
If `$day['name']` is a part of `dbStringsForTranslation`, then the original code of this change should work properly.

fvlasie:
So perhaps the POE term is left over from a previous version of the file. Shall I just delete the bad term in POE and hope it will not reappear?

SKuipers:
Yep, sounds good 👍  I've done a search through the codebase and don't see any other instances of it.

fvlasie:
Okidoki

PR #814, opened 2019-03-20
Typo fix

fvlasie:
Name of Google Project in Devlopers Console.


rossdotparker:
Oh no! I didn't see this, and then went ahead and made the change myself. Sorry...means I stole your thunder : ( I'll try and read through all PRs in future before acting on one.

You'll see I used a slightly different WHERE in the end....out of habit, not because it works any better.

PR #834, opened 2019-04-07
System: transplant dashboard contents from functions.php into classes

SKuipers:
Currently, the `functions.php` file contains large chunks of HTML mixed with php and sql, much of it used for rendering dashboards and interface elements. Luckily, because these functions are only called from one location, `index.php`, they're pretty good candidates to transplant & start working on turning into templates. 

Right now, this PR only transplants the chunks of content, it doesn't re-write them. The idea is to isolate them out of `functions.php` into their own classes, where we can then work on refactoring them into data + templates. Something to create an issue/tracker for once we're done the forms and tables.

I've made two tweaks to the transplanted code: One to help the parent dashboard display better on small screens. And another to wrap sidebar items in a div (useful for the next PR).

Functions affected (~2,200 lines removed from functions):
- getStaffDashboard
- getStudentDashboard
- getParentDashboard
- getParentPhotoUploader
- getNotificationTray
- getMinorLinks
- sidebar

Tested locally.


rossdotparker:
It's incredible seeing some of these monolithic files being broken up!

PR #839, opened 2019-04-10
Forms: fix the Finder input width, remove hard-coded styles

SKuipers:
Currently, there's lots of little bits of inline style around to fix and overwrite the token finder styles and width. This PR removes them all, and completely re-styles the finder input. It now supports flexible field widths, and uses tailwind classes to help standardize the input style.

It also fixes a bug where some finder inputs weren't allowing new tokens to be created because the current jQuery token script we're using has a different parameter name `allowFreeTagging`.


rossdotparker:
@SKuipers something has gone wrong here, as existing tags don't show up as suggestions any more (e.g. biology, which does exist, in the example below):

Screenshot 2019-04-10 at 9 48 01 PM

Thanks!

Ross

SKuipers:
Hmm, I haven't been able to reproduce this one, perhaps the query that loads the possible tags needs checked? In the demo data, there's no `biology` keyword, but there is `problem solving`, and it seems to come up as expected:
Screen Shot 2019-04-11 at 11 17 04 AM

rossdotparker:
Turns out it was just me being a bad user. Case closed!

PR #840, opened 2019-04-10
Templates: fix the BulkAction header, and more responsive tweaks

SKuipers:
The layout for the bulk action form got a bit out of whack when switching to flexbox layouts. This PR updates it to position and scale better. It also fixes the placement of the action-selector's validation.

This PR also resets the font size for all inputs on mobile screens. This prevents the annoying zoom-in effect that happens on mobile browsers when you click on an input (the zoom is annoying because it doesn't zoom you back out when you're done). 

See the commit list for the full set of tweaks.


rossdotparker:
This all looks good, but there is a conflict with the scss and min.css files in resources. Can you resolve those, and then I'll look back into it?

PR #842, opened 2019-04-12
Tables: update Data Table classes to use template-based renderers

SKuipers:
This is another major step for the mobile-responsive templates in v18, following #824. Similar to that PR, this one replaces the renderers for DataTable classes with those that output HTML using a view template.

This changeset introduces the `context` method to columns when creating tables. Adding a context is a way to provide additional information that the renderer can use to make decisions for how to display that column. In this case, `context('primary')` and `context('secondary')` can tell the renderer that these columns should be displayed on small screens, and other columns will be hidden.

For tables that don't have any `context` defined, it automatically applies some default contexts to the first three table columns. For the majority of cases, the first couple columns have the most important details, so a large number of our tables are already responsive with no additional tweaking. From here, we can use the the `context` method to manually tweak any tables that need to have specific columns displayed on mobile screens. I've updated the Manage Users table as an example.

In this PR:
- Creates two new view classes: `DataTableView` and `PaginatedView`
- Sets up the Twig templates for rendering tables in `dataTable.twig.html` and `paginatedTable.twig.html`
- Adds the table views to the `ViewServiceProvider`
- Similar to the Forms, it re-wires the DataTable::create method internally to use the new renderers. In a follow-up PR I'll go through and swap these over to using the DI container to instantiate new tables.
- Tweaks the behaviour of hover-over tooltips to not display on mobile. I'm testing this out as a solution to the issue that mobile browsers make you double-click any link with a hover-over effect.
- Adds a mobile-friendly menu for actions:
  Screen Shot 2019-04-11 at 3 37 51 PM

Tested locally and Travis is happy.

### DIFF
--- a/src/Tables/DataTable.php
+++ b/src/Tables/DataTable.php
@@ -57,9 +57,8 @@ class DataTable implements OutputableInterface
      * @param string $id
      * @param RendererInterface $renderer
      */
-    public function __construct($id, RendererInterface $renderer = null)
+    public function __construct(RendererInterface $renderer = null)
     {
-        $this->id = $id;
         $this->renderer = $renderer;
     }
 
@@ -72,7 +71,7 @@ class DataTable implements OutputableInterface
      */
     public static function create($id, RendererInterface $renderer = null)
     {
-        return new static($id, $renderer ? $renderer : new SimpleRenderer());
+        return (new static($renderer ? $renderer : new SimpleRenderer()))->setID($id);
     }
 
     /**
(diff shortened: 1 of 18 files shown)


src/Tables/DataTable.php
```diff
@@ -57,9 +57,8 @@ class DataTable implements OutputableInterface
      * @param string $id
```
rossdotparker: Should this not be removed too?

PR #844, opened 2019-04-16
Templates: update data table headers, add blank slates, cleanup css files

SKuipers:
Building some more on the previous PRs, this updates the UI elements in the data table headers by applying some tailwind classes and tweaking the template files.

**Header & pagination:**
Screen Shot 2019-04-16 at 1 05 25 PM

**With filters applied:**
Screen Shot 2019-04-16 at 1 05 37 PM

This PR also adds in the template code for blank slates, which display as an empty space rather than displaying an error-message style alert (see  #686). Blank slates currently display either `There are no records to display.` or `No results matched your search.` if there's any filters/search applied. Blank slate content can also be customized through the table's meta data, allowing for custom messages (eg: `All students are present.` for attendance reports)

**Default blank slate:**
Screen Shot 2019-04-16 at 1 06 01 PM

**No results:**
Screen Shot 2019-04-16 at 1 06 19 PM

Now that these are templated, we have the option to make them a little more user friendly. For example, on many sites an empty table will have a larger, more noticeable `Add` button inside the blank slate.

This PR also continues to cleanup the main.css file and transition styles over to theme.scss (primarily alerts and table styles), as well as delete more unused stuff.


rossdotparker:
There are some really nice visual enhancements here. I love the direction you are taking the UI : )

PR #847, opened 2019-04-19
Planner: Edit Lesson OOification 100% Complete

rossdotparker:
Further progress...


SKuipers:
I've pushed a fix for the smart block tinymce editors, and tweaked the layout a bit. Turned out the `getEditor()` param `tinymceInit` being true was causing it to double-initialize those text areas.

PR #849, opened 2019-04-21
Rubric Visualisation Image Export

rossdotparker:
**Description**
This PR separates out the rubric visual into an object, which can be invoked to display normally (e.g. within the rubric) or to export as a PNG image, via ajax, to the file system.

**Motivation and Context**
ICHK are moving towards including a visual rubric overview in reports, and this is a first step. 

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2019-04-22 at 12 59 58 AM

### DIFF
--- a/modules/Rubrics/src/Visualise.php
+++ b/modules/Rubrics/src/Visualise.php
@@ -19,28 +19,17 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 namespace Gibbon\Module\Rubrics;
 
-use Gibbon\Contracts\Database\Connection;
-use Gibbon\session;
 use Gibbon\UI\Chart\Chart;
 
 /**
  * Attendance display & edit class
  *
- * @version 12th Sept 2016
- * @since   12th Sept 2016
- * @author  Sandra Kuipers
+ * @version v18
+ * @since   v18
  */
 class Visualise
 {
-    /**
-     * Gibbon\Contracts\Database\Connection
-     */
-    protected $pdo;
-
-    /**
-     * Gibbon\session
-     */
-    protected $session;
+    protected $absoluteURL;
 
     protected $page;
 
@@ -59,18 +48,13 @@ class Visualise
     /**
      * Constructor
      *
-     * @version  3rd May 2016
-     * @since    3rd May 2016
-     * @param    Gibbon\session
-     * @param    Gibbon\config
-     * @param    Gibbon\Contracts\Database\Connection
+     * @version  v18
+     * @since    v18
      * @return   void
      */
-    public function __construct(\Gibbon\Core $gibbon, Connection $pdo, $page, Int $gibbonPersonID, Array $columns, Array $rows, Array $cells, Array $contexts)
+    public function __construct($absoluteURL, $page, Int $gibbonPersonID, Array $columns, Array $rows, Array $cells, Array $contexts)
     {
-        $this->session = $gibbon->session;
-
-        $this->pdo = $pdo;
+        $this->absoluteURL = $absoluteURL;
 
         $this->page = $page;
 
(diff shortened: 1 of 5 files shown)


modules/Rubrics/src/Visualise.php
```diff
@@ -0,0 +1,168 @@
...
+use Gibbon\Contracts\Database\Connection;
+use Gibbon\session;
+use Gibbon\UI\Chart\Chart;
+
+/**
+ * Attendance display & edit class
+ *
+ * @version 12th Sept 2016
+ * @since   12th Sept 2016
+ * @author  Sandra Kuipers
```
SKuipers: Looks like a copy-pasted file 😂 I didn't think my name was still lurking on any of these file headers, gotta fix that.

modules/Rubrics/src/Visualise.php
```diff
@@ -0,0 +1,168 @@
...
+     * Constructor
+     *
+     * @version  3rd May 2016
+     * @since    3rd May 2016
+     * @param    Gibbon\session
+     * @param    Gibbon\config
+     * @param    Gibbon\Contracts\Database\Connection
+     * @return   void
+     */
+    public function __construct(\Gibbon\Core $gibbon, Connection $pdo, $page, Int $gibbonPersonID, Array $columns, Array $rows, Array $cells, Array $contexts)
```
SKuipers: Looks like the `$pdo` dependency isn't being used by this class. The `$gibbon` dependency appears to only be used for the absoluteURL, so to keep things simple it may work to just pass in the `$absoluteURL` string, and remove the other two class dependencies.

modules/Rubrics/src/Visualise.php
```diff
@@ -0,0 +1,168 @@
...
+    protected $rows;
+
+    protected $cells;
+
+    protected $contexts;
+
+    /**
+     * Constructor
+     *
+     * @version  3rd May 2016
```
SKuipers: 👀

modules/Rubrics/src/Visualise.php
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            $columnCount = count($this->columns);
+            $data = array_map(function ($mean) use ($columnCount) {
+                return !empty($mean['denonimator'])
+                ? round((($mean['cumulative']/$mean['denonimator'])/$columnCount), 2)
+                : 0;
+            }, $means);
+
+            $this->page->scripts->add('chart');
```
SKuipers: This `$page` dependency appears to be pulled in just for the script loading 🤔 I wonder if we an bump it out of this script back into the parent script, since it's more related to the final rendering than the chart data itself.

modules/Rubrics/src/Visualise.php
```diff
@@ -0,0 +1,168 @@
...
+
+            $chart = Chart::create('visualisation', 'polarArea')
```
SKuipers: I wonder if this class could extend the Chart class, rather than creating one internally 🤔

modules/Rubrics/src/Visualise.php
rossdotparker: Hahaha, not as thorough as I thought I was being. Fixed now.

modules/Rubrics/src/Visualise.php
rossdotparker: Thanks for setting me straight on these. A lot of this work was speculative, and I meant to come back and clean it up...but never did. Fixed now.

modules/Rubrics/src/Visualise.php
rossdotparker: Fixed.

modules/Rubrics/src/Visualise.php
rossdotparker: This definitely makes sense, but I'm out of time for this one. Feel free to tinker, but otherwise I'll leave it as it is for now. I'll commit the changes above for merging. Thanks Sensei!

SKuipers:
Looking good, I'm going to merge this in and may yet tinker with the class inheritance when I have some time. Thanks!

PR #852, opened 2019-05-03
System: add a BackgroundProcessor service

SKuipers:
Occasionally we need to start long-running processes; for messenger, notifications, report generation, etc. This PR adds a `BackgroundProcessor` service and abstract `BackgroundProcess` base-class.

I have been using a similar set of classes in my own modules to start and monitor separate php processes, and since it's used for all the notifications in my staff absence and coverage feature, I've tidied up the classes for use in the core. It isn't intended to be a full-on queue manager or general-purpose task runner: it's specifically geared for speeding up page loads by executing long-running code in a separate php process.

How it works:

One could create a class called `MessageSenderProcess` which extends `BackgroundProcess` and call a method `startSendingMessages($arg1, $arg2, ...)` on it. The BackgroundProcessor will start up a separate php process and store the arguments so they can be passed along safely to the new process. When the new process starts up, it first validates access to that process, then grabs the stored arguments, creates a `MessageSenderProcess` class and calls `runSendingMessages($arg1, $arg2, ...)` on it. The whole thing adds up to a single line:

`$container->get(MessageSenderProcess::class)->startSendingMessages($arg1, $arg2, ...);`

If for any reason the background process fails to start (shared hosts, disallowed `exec`) then the process will still start in the current runtime. This helps ensure there's a fallback for various systems, so processes don't get lost.

The DI container handles all the instantiation and auto-wiring, so the `MessageSenderProcess` class could use a constructor to get any services or gateways it may need access to. The aim is for background processes to be as painless as possible, with a fair bit of flexibility.

The `BackgroundProcessor` makes a best effort to support Windows systems which handle `exec` functions differently, based on PHP documentation out there and inspired by [cocur/background-process](https://github.com/cocur/background-process). I have yet to test it on Windows though.

I've commented this one a bunch, give a shout with any questions though 😄

_Requires #851_

### DIFF
--- a/cli/system_backgroundProcess.php
+++ b/cli/system_backgroundProcess.php
@@ -18,6 +18,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
 use Gibbon\Services\BackgroundProcessor;
+use Gibbon\Services\Format;
 
 $_POST['address'] = '/modules/'.($argv[3] ?? 'System Admin').'/index.php';
 
@@ -27,6 +28,9 @@ require __DIR__.'/../gibbon.php';
 if (!isCommandLineInterface()) {
     die(__('This script cannot be run from a browser, only via CLI.'));
 }
+
+// Setup some of the globals
+Format::setupFromSession($container->get('session'));
 
 // Override the ini to keep this process alive
 ini_set('memory_limit', '2048M');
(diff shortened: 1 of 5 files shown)


rossdotparker:
That was a lot to get my head around. I'm not 100% on every detail, but I get the big picture, and I've learned some new PHP tricks (e.g. splat, abstract, etc). Looking forward to seeing this used in practice.

PR #855, opened 2019-05-06
Timetable: fix some staff not showing up in View Timetable by Person

SKuipers:
A small fix. The staff `type` field allows selecting general "Teaching", "Support" types, as well as specific roles. This was causing the particular query used for View Timetable by Person to not list some staff if their type was set a role, eg: "Teacher".


rossdotparker:
Looking good. I made this change for a requesting user, and it is one that I regret. This field used to serve a useful purpose, but I am not sure it does since that change. Win some, lose some...

SKuipers:
Ah, the change for staff type to accept any role? It does seem to overlap with role category and job title, but not quite enough of both that it seems safe to remove 🤔

PR #857, opened 2019-05-06
Forms: update Time input class, update DatabaseFormFactory

SKuipers:
A couple tweaks to the form classes. Primarily adds minimum and maximum methods to the Time input class, used to limit the available times client-side.

### DIFF
--- a/src/Forms/Input/Time.php
+++ b/src/Forms/Input/Time.php
@@ -30,7 +30,10 @@ namespace Gibbon\Forms\Input;
 class Time extends TextField
 {
     protected $format = 'H:i'; // Default to 24 hour clock
+    protected $min;
+    protected $max;
     protected $chained;
+    protected $showDuration;
 
     /**
      * Set the format to output time values (default 'H:i').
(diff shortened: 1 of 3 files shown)


src/Forms/Input/Time.php
```diff
@@ -42,14 +45,38 @@ public function setFormat($format)
...
 
+    /**
+     * Define a minimum for this time value.
+     * @param   string  $value
+     * @return  self
+     */
+    public function minimum($value)
+    {
+        $this->min = $value;
+        return $this;
```
rossdotparker: Why is this not:

`return $this->min;`

src/Forms/Input/Time.php
```diff
@@ -42,14 +45,38 @@ public function setFormat($format)
...
+
+    /**
+     * Define a maximum for this time value.
+     * @param   string  $value
+     * @return  self
+     */
+    public function maximum($value)
+    {
+        $this->max = $value;
+        return $this;
```
rossdotparker: Likewise, why is this not:

`return $this->max;`

rossdotparker:
I dropped a couple questions, but happy enough to merge in the meantime. I'd definitely be interested in learning the answer. Is it to do with the way the method calls can be chained together?

SKuipers:
Yep! They're a chained method call, so it sets a value and returns itself, so the chain can continue. I've tried to make all the setters in the Form and Table APIs chainable, to give us the ability to compose objects in a readable fashion.

jimbo8098:
That's a good idea, I like it 👍

PR #860, opened 2019-05-06
Attendance: Prefill Settings

rossdotparker:
**Description**
This PR removes the prefillClass setting, and replaces the prefillPerson and prefillRollGroup with countClassAsSchool (which inherits its value from prefillRollGroup). The result is the option to co-mingle or separate class and school attendance.

As a knock on, the Mini History has been tweaked to become context aware (in v17, it draws data from all class and school attendance, even in class view, which is potentially confusing).

**Motivation and Context**
Sandra and Ross have been discussing shortcomings in the current attendance model (present in v17 and several versions before), specifically, that the system does cannot make a clean split between class and school attendance. This can lead to confusion, when a student is marked absent from class, and thus deemed absent from school.

**How Has This Been Tested?**
Locally, but I'd like @SKuipers to really kick the tires on this one to make sure I've not missed anything.

**Screenshots**
Settings are much simpler now:
Screenshot 2019-05-06 at 10 53 40 PM


modules/Attendance/attendance_take_byCourseClass.php
```diff
@@ -214,21 +213,31 @@
...
-                            $sql .= " ORDER BY timestampTaken DESC";
+                                    AND date LIKE :date
+                                    AND context='Class' AND gibbonCourseClassID=:gibbonCourseClassID
+                                    ORDER BY timestampTaken DESC";
                             $result = $pdo->executeQuery($data, $sql);
 
                             $log = ($result->rowCount() > 0) ? $result->fetch() : $defaults;
 
+                            //Check for school prefill if attendance not taken in this class
+                            if ($result->rowCount() == 0 ) {
```
SKuipers: This section is very similar to a modification I made for my school, so this is a question from scenarios that cropped up over the past couple years:

Once attendance is taken, should this page always display a definitive log of exactly what was recorded at the time, or should it continue to prefill in contexts? Option A: always display exactly what was recorded, Option B: continue to fill in changes. I ask, because Option A sounds logical on face-value, but Roll Group attendance actually follows Option B, which makes sense for its context.

- Option A: If we always prefill this class first, regardless of other attendance, it doesn't allow for the Attendance by Person to fill in, should a student sign-out at the front office chronologically after the class attendance was taken. 
  - For example, a teacher might take attendance and mark a student as Absent - Unexcused (or Absent - Pending I think for your school). The student wasn't in their class because they went to the nurse, and later signed out at the office to go home sick. Their new Attendance by Person log is Absent - Excused: Medical, taken after the class attendance. If anyone later look at the logs for that class, they would see the student has an unexcused absence (nut no reason), when the reason actually is known. When signing out a student, would the office need to check and amend their class attendance logs too?

- Option B: If the attendance log can continue to "prefill" after it has been recorded, then it's not a definitive record of what was taken by that teacher at that time, but instead represents current the status of the student, which fills in accurately based on any Attendance by Person that was taken later on. This is how Roll Group attendance currently works.

At TIS we went with Option B for class attendance. One way I got around this was to combine the two queries into one, and use `AND (context<>'Class' OR (context='Class' AND gibbonCourseClassID=:gibbonCourseClassID)) ORDER BY timestampTaken DESC`. 

Does this make sense? Amy I missing something or over-complicating things? My head hurts trying to figure all this out, so maybe 😬

rossdotparker: Just reading your Option A vs Option B comment. My feeling is very much Option B: once a teacher has entered attendance for a Class, other contexts should not prefill, otherwise it risk becoming very confusing. It has the benefit of being simpler on the logic side, I feel.

modules/Attendance/attendance_take_byCourseClass.php
```diff
@@ -214,21 +213,31 @@
...
+                                $data = array('gibbonPersonID' => $student['gibbonPersonID'], 'date' => $currentDate . '%');
+                                $sql = "SELECT type, reason, comment, context, timestampTaken FROM gibbonAttendanceLogPerson
+                                        JOIN gibbonPerson ON (gibbonAttendanceLogPerson.gibbonPersonID=gibbonPerson.gibbonPersonID)
+                                        WHERE gibbonAttendanceLogPerson.gibbonPersonID=:gibbonPersonID
+                                        AND date LIKE :date
+                                        AND NOT context='Class'
```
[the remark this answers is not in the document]
rossdotparker: My thinking was to avoid classes cross-filling each other, as it is less authoritative than school attendance. For example, a student might be in school but have skipped a lesson and be marked absent. The following class should not fill as Absent, otherwise the teacher might expect to not see them.

modules/Attendance/attendance_take_byCourseClass.php
SKuipers: Sounds good 👍

PR #864, opened 2019-05-08
Staff: add tables for staff absences, coverage and substitutes

SKuipers:
The first part in the staff absence & coverage features. I've separated the tables out into an initial PR, because many of the scripts in following PRs have queries and table joins which require these.

I didn't squish them all onto one line because it gets really hard to read, but I can update it if you prefer less whitespace. Is the gibbon.sql update perhaps not necessary, because we'll be re-building it this release?

Tested locally & in production.


rossdotparker:
Hi Sandra, the tables look good. Do you want to squash the CHANGEDB.php statements down to single lines...and you are right, gibbon.sql does not need to be changed. For stable installs, we update during the release process: for cutting edge, it is done dynamically using CHANGEDB.php. If you can make these changes, then I'll merge : )

SKuipers:
Thanks, it's squashed! Once you merge this I'll push an update to the other PRs CHANGEDBs too.

PR #875, opened 2019-05-14
System: Install Fix "...you do not have access to this action"

jimbo8098:
- Moving jquery script src references in install.php above core.js to solve an error showing in dev tools at runtime.
- Fix a typo in install.php created through the generation of config.php
- Fixing "Your request failed because you do not have access to this action."

This 3rd one was much larger. Previously, a $guid variable was used to write $_SESSION variables directly in place of the $gibbon->session method used on other pages. This change adds a setGuid(string $guid) to the class so this can be used. This actually solves a few issues with this process:

For one, stringReplacement and absolutePath were previously added using $gibbon->session->set() but as no $guid was present in the config file, these would be added to the "undefined" key within the $_SESSION. Additionally, it resolves an issue between steps 3 and 4 where the guid would not be present on the session. Technically, there are ways I could have gone about resolving this using $_SESSION and the POSTed guid instead but this would mean we would be utilising the builtin session handling instead of the Gibbon one that people have spent so long integrating into the existing codebase.

I did try assigning $GLOBALS['guid'] the posted $guid in order to use the global guid to assign at __construct of Gibbon\Session but this didn't work. I don't see any reason why since all the session does is read from the config which is present at the time of pageload on transition from step 3 to 4. The guid is even correct and a var_dump on $gibbon->session->guid() returns the expected POSTed guid.

Since it's there in the config, I'd have expected it to be assigned on line 72 of Session.php but this doesn't appear to be the case.

**Description**

All changes refer to install.php and Session.php.

**How Has This Been Tested?**
Locally in a test environment. No issues found.


jimbo8098:
Thanks for checking my PR Sandra!

I'll test it on Linux today all going well. I was deleting the config.php
and database each time to fully replicate all errors when I was testing on
Windows yesterday.

MacOSX is the only one I don't have so if someone could try and install
using that, that would complete the circle.

jimbo8098:
I was trying to install the old version of the Gibbon Core (prior to my PR) and that went perfectly fine on Linux so it seems this issue affects specifically Windows servers using the default PHP session handler. My setup is using file sessions, not DB ones which could be another factor. 

Still, no harm in using the new Sessions across the board as far as I can see. Will try to install using master and see how I get on.

jimbo8098:
Confirmed working on both Linux and Windows now 👍

SKuipers:
Awesome, thanks for testing! Looks ready to merge.

PR #877, opened 2019-05-15
System: update the installer to write config.php from a template

SKuipers:
With our recent installer changes I noticed it'd be pretty easy to write the `config.php` file from a template rather than a long concatenated string. Also moves the installer-related templates into a sub-folder.

**Motivation and Context**
Easier to maintain the config file the future.

**How Has This Been Tested?**
Tested locally, PHP 7.0 - PHP 7.3.


jimbo8098:
Excellent idea. Works perfectly on my Windows server.

PR #887, opened 2019-05-20
Fix term smark to smart

leofreitas:
**Motivation and Context**
Fix the term "Smark" to "Smart" in file "modules/Planner/units_edit_smartBlockify.php" and translation space

**How Has This Been Tested?**
Local


rossdotparker:
Thanks Leo, well spotted. I've merged and will update in POEditor and the i18n sub module as well.

leofreitas:
Hi Ross, I updated in POEditor.

Issue #886, opened 2019-05-20
[v18] Windows: Login Session Handling


jimbo8098:
I've spotted a weird, reproducable, bug with Windows IIS using file-based PHP session handling during login. When you try to log in on Windows, the login.php file sets your header() to index.php, as expected, however am incorrect session is used. Here are the headers for a server running Windows (localhost in this case):

![image](https://user-images.githubusercontent.com/12417556/58013824-60201b00-7aef-11e9-8d0d-a41f8d948e2b.png)

So here we see the 302 response which forwards us to index.php. As you can see, a PHPSESSID is sent back in the response. When this returns, I am placed back on the index page with the Gibbon intro text.

Following that, I changed my hosts file to redirect the same URL over to my laptop which runs CentOS with Apache. When I use Apache with PHP's default, file based session handling, the following request and response are observed:

![image](https://user-images.githubusercontent.com/12417556/58013970-abd2c480-7aef-11e9-96ec-1df5a63ddc2a.png)

Note that when the request is sent to Apache a cookie is embedded in the request. This is not present in the request sent to the Windows and looks to be the cause of this error and, as I suspect, in many other pages for Windows clients until that successful login is received (see #875). I believe there may be some flag required on Windows hosts 

Another interesting thing is that if I switch my hosts file back to point at localhost after authenticating on the CentOS laptop, I remain logged in! The password and username on both the laptop and localhost are the same, as is the Guid for the installation.

I have been playing around and using the same process as #875, I converted the $_SESSIONs over to $gibbon->session->get/set but this did not resolve the problem (though works on Linux too so might be worth a PR to use that style of session handling?) 

I'm raising here to see if anyone else has this problem and to also see if someone has came across a solution.

jimbo8098:
The difference following Linux login is that the PHPSESSID cookie is sent to the server in the request. This is not sent prior to Linux login.

jimbo8098:
Something between fabb9666dd58839b06d18a5892b3accaa53111ce and c520e605101fe7d1b083a64a52fabf4379c071e0 appears to have fixed it. Interesting!

jimbo8098:
I need to get a chance to re-test this but alas my Windows install is gone. My old PR was unfortunately pretty garbled due to my dev build at the time but I will keep it alive on my fork for the moment pending futher tests. At the moment, it's just too out of date to be useful.

PR #889, opened 2019-05-21
modules/Badges OOification Changes

jimbo8098:
This PR adds OOification requirements for my latest PR in the module-badges repo. It adds a domain for use with the Badges module.


SKuipers:
I see some interesting changes in here I'm keen to check out 👀 It looks like the whole badges module has tagged along for the ride in this PR. I suspect it may need ignored globally on your system or locally in a `.git/info/exclude` file. Also, as we just mentioned in slack, the gateway for the Badges module can live inside the modules/Badges/src folder if it has a `Gibbon\Module\Badges` namespace. Are you able to update these and push your changes to this PR? Thanks!

jimbo8098:
Thanks for the feedback @SKuipers , I've removed the ignored files from the PR along with the BadgesGateway which has been moved to the module-Badges repo instead (based on our Slack conv).

I also spotted some leftover debug in index.php which I've removed.

jimbo8098:
Right, finally got it! So, to recap:

- Moving to $gibbon->session->* instead of $_SESSION where issues were noted. This is mainly within login.php (since I was trying to fix #886 (unsuccessfully). I just noticed student_view_details was hard coded with that module name while using it as an example.
- QueryCriteria now supports array values. It sanitises the input then adds it to the array.
- PaginatedView has been updated to support the array input change above. Prior this caused errors.

You might question the need for array support in QueryCriteria, it's used in order to allow more flexible data transfer. Without it, I'd have had to string encode my array of students rather than simply sending the array.

This can be used in place right now since there is a check on the value of the object prior. However, I suspect this may have been what @SKuipers was referring to when she mentioned "array criteria"?

jimbo8098:
Thanks for your feedback Sandra!

On the subject of the query criteria supporting an array going into the
criteria. This is because on one of he pages in module/Badges, there is a
multi student search. To maintain it I'd either need to use the base PDO or
make the Query criteria handler for it. I can use PDO if that's preferred?

SKuipers:
Hi Jim. This PR ended up with an interesting mix of changes, which is why it hasn't been merged yet. The `$_SESSION` to `$gibbon->session` updates in login.php on their own look good, yet there's some other SID stuff happening in the changes that I'm not sure is necessary 🤔 

It looks like the QueryCriteria changes may be adding a level of complexity to this object that isn't quite intended. The goal of QueryCriteria is to hold values that are passed to and from the front-end via strings (on URL parameters) and as simple JSON objects (via the PaginatedTable renderer). It's possible that there are other ways to achieve what you're looking for without modifying this functionality, such as imploding your IDs to a csv string, or passing them directly as a parameter to the Gateway method being called (eg: `queryBadgesByStudents($criteria, $gibbonPersonIDList = [])`.

jimbo8098:
I'm actually just going to close this, it's such an old PR and it's all over the place anyway so I don't see it being helpful in the current meta. I suspect the recent overhauls will in time resolve it anyway. I'll keep it in my fork and see if it happens again during my tests.

PR #891, opened 2019-05-24
Feature multiple traits from DataSet

jimbo8098:
This commit simply adds a fromDataSet call which can be used with the DataSet object returned by Domains. It's basically just the toArray function with configurable column names for the value and name.

With the move to Domains, it isn't necessarily code-friendly to run toArray and transform the output into the correct form. The other option, creating specific queries for each type of select box you might have, could result in mutiple SQL queries per page and/or extra lines of code just to have the column names set to value and name.


src/Forms/Traits/MultipleOptionsTrait.php
```diff
@@ -104,6 +105,42 @@ public function fromQuery(Connection $pdo, $sql, $data = array(), $groupBy = fal
...
+     * @param Dataset $dataset
+     * @param string $valCol
+     * @param string $nameCol
+     * @param string $groupBy
+     * 
+     * @return self
+     */
+    public function fromDataSet(Dataset $dataset, $valCol, $nameCol, $groupBy = false)
+    {
+        if(empty($dataset) || !is_object($dataset)){
```
SKuipers: I think in this case the type-hint for `Dataset` in the method signature will already throw an exception if the argument is not an object of the correct type.

src/Forms/Traits/MultipleOptionsTrait.php
```diff
@@ -104,6 +105,42 @@ public function fromQuery(Connection $pdo, $sql, $data = array(), $groupBy = fal
...
+        }
+
+        if($dataset->getTotalCount() > 0)
+        {
+            $options = array_filter($dataset->toArray(), function ($item) use ($valCol,$nameCol) {
+                return isset($item[$valCol]) && isset($item[$nameCol]);
+            });
+        }
+
+        foreach ($options as $option) {
```
SKuipers: `$options` is being set inside the preceding `if`, so it looks like this loop may also be best inside the same conditional, otherwise the current code could throw a PHP notice for using an unassigned variable.

SKuipers:
There's a couple small tweaks to this PR, then I think it's ready to merge. Also can you merge the v19 branch into this branch, and update the PR branch to GibbonEdu:v19.0.00. Thanks!

PR #892, opened 2019-05-24
Added createSelectHouse

jimbo8098:
Simply adding a createSelectHouse function to DatabaseFormFactory. Not 100% sure if this should be using a domain but I'm also not sure if that is technically possible at this early stage.


src/Forms/DatabaseFormFactory.php
```diff
@@ -107,6 +107,22 @@ public function createSelectRollGroup($name, $gibbonSchoolYearID, $all = false)
             return $this->createSelect($name)->fromArray(array("*" => "All"))->fromResults($results)->placeholder();
     }
 
+    public function createSelectHouse($name)
+    {
+        $data = [];
+        $sql = "SELECT gibbonHouseID as value, name FROM gibbonHouse;";
+        try
+        {
+            $results = $this->pdo->executeQuery($data, $sql);
```
SKuipers: Hi Jim! There's a couple outstanding PRs from the tail end of v18. It'd be great to get them merged into v19. In this case, its a tiny change, but the whole try/catch can be replaced with a `$results = $this->pdo->select($sql);` line.

PR #896, opened 2019-06-04
User Admin: move Manage Student Settings and Manage Staff Settings into User Admin 

SKuipers:
Based on discussions in v18 testing, this moves the Manage Student Settings and Manage Staff Settings into User Admin, and merges with the Manage Staff Settings page already there.

**Motivation and Context**
Tidies up the action categories to help them make a bit more logical sense.

**Screenshots**
The end result:
Screen Shot 2019-06-04 at 1 56 18 PM


rossdotparker:
Looking good. This should bring some clarity to a slightly murky situation.

PR #900, opened 2019-06-14
V18 support translation review01

leofreitas:
**Motivation and Context**
Support to translation some strings and applying pattern Gibbon in code (methods translatable, dateRangeReadable)

**How Has This Been Tested?**
Local and Travis

### DIFF
--- a/modules/Library/library_manage_catalog.php
+++ b/modules/Library/library_manage_catalog.php
@@ -287,7 +287,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Library/library_manage_cat
                 }
                 if ($resultSpace->rowCount() == 1) {
                     $rowSpace = $resultSpace->fetch();
-                    echo $rowSpace['name'].'<br/>';
+                    echo __($rowSpace['name']).'<br/>';
                 }
             }
             if ($row['locationDetail'] != '') {
@@ -316,7 +316,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Library/library_manage_cat
             }
             echo '</td>';
             echo '<td>';
-            echo $row['status'].'<br/>';
+            echo __($row['status']).'<br/>';
             echo "<span style='font-size: 85%; font-style: italic'>".ynExpander($guid, $row['borrowable']).'</span>';
             echo '</td>';
             echo '<td>';
--- a/modules/Library/library_manage_catalog.php
+++ b/modules/Library/library_manage_catalog.php
@@ -287,7 +287,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Library/library_manage_cat
                 }
                 if ($resultSpace->rowCount() == 1) {
                     $rowSpace = $resultSpace->fetch();
-                    echo __($rowSpace['name']).'<br/>';
+                    echo $rowSpace['name'].'<br/>';
                 }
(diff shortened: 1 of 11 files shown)


leofreitas:
Thanks Sandra! :clap:

modules/Library/library_manage_catalog.php
```diff
@@ -287,7 +287,7 @@
                 }
                 if ($resultSpace->rowCount() == 1) {
                     $rowSpace = $resultSpace->fetch();
-                    echo $rowSpace['name'].'<br/>';
+                    echo __($rowSpace['name']).'<br/>';
```
rossdotparker: I am not sure it makes sense to wrap this one, as it is a user-entered value, and so is unlikely to be in the string translation list, but is likely to be in the school's language already. Same with __($rowSlots['nameShort']) in  modules/Activities/report_attendanceExport.php I believe. I could be wrong though...

modules/Staff/staff_view_details.php
```diff
@@ -37,20 +37,14 @@
         echo __('The highest grouped action cannot be determined.');
         echo '</div>';
     } else {
-        $gibbonPersonID = $_GET['gibbonPersonID'];
-        if ($gibbonPersonID == false) {
+        $gibbonPersonID = $_GET['gibbonPersonID'] ?? '';
```
rossdotparker: Would it not be better to use null here?

modules/System Admin/import_run.php
```diff
@@ -97,7 +97,7 @@
         $availableModes = array();
         $modes = $importType->getDetail('modes');
         if (!empty($modes['update']) && !empty($modes['insert'])) {
-            $availableModes['sync'] = __('Update').' & '.__('Insert');
+            $availableModes['sync'] = __('Update & Insert');
```
rossdotparker: This possibly creates a new string for translation, but definitely more elegant!

leofreitas: Hi Ross! This string already existed! :clap:

modules/Library/library_manage_catalog.php
leofreitas: This really does not make sense here, but in modules/Activities/report_attendanceExport.php the value of  $rowSlots['nameShort'] is not user-entered.

modules/Staff/staff_view_details.php
leofreitas: In all the code we're using '', then I continued with the pattern.

SKuipers: Yep, thanks, that's my doing! 😄 I prefer an empty string for GET values rather than null. Since they're being parsed from a URL string, the empty value is already a string type. Introducing null means scripts need to be more careful to use `empty()` to check for '' and null, and some of the legacy scripts still do an ` == ''` check.

SKuipers:
> Hi Sandra, it looks unproblematic to me, but I've left a few comments to help you decide if/how to merge. My feeling is that it is very late in the v18 day, but if the changes are well tested, then we can proceed. Thanks!

Thanks for the code review! I agree it's late, they're testing-related tweaked so it'd be nice to wrap them up in this version, so I'm going to merge this as the final PR of v18 and then test it again post-merge.

PR #901, opened 2019-06-14
Fixed Installation Script

leofreitas:
**Description**
Removing white space in the records: tables gibbonAction, gibbonNotificationEvent and gibbonSetting.

**How Has This Been Tested?**
Local


SKuipers:
Very thorough 😄  Are you able to check if these white spaces exist in POEditor and update them there too? To ensure we don't end up with duplicate strings. Looks like these would be part of the dbStringsForTranslation.php file when we run it for release.

leofreitas:
Hi Sandra, I updated the POEditor and dbStringsForTranslation.php. I found one more term. 😆

PR #905, opened 2019-07-31
Receipt emailed is not showing payment details

jian118:
**Description**
Receipt emailed doesn't show payment details as email is sent before payment record is saved. 
- Receipt number should start with 0 hence the -1
- ReceiptNumber 0 when compared with null is true so a type and value comparison (===) should be used instead


SKuipers:
Hi @jian118. Thanks for your PR! I'm keen to check out the fix you've provided. I have my hands full this week with our school year start-up, but I hope to take a look next week.

SKuipers:
Thanks @jian118. I've merged this in and tested, looks great! 👍 It turns out, this was the solution we needed for an outstanding [forum thread](https://ask.gibbonedu.org/discussion/2590/bug-in-receipt-emailing#latest) too. Many thanks!

PR #906, opened 2019-08-09
 Behaviour: new Copy To Notes feature

rossdotparker:
**Description**
This new feature adds a Copy to Notes checkbox to the bottom of the Add and Add Multiple behaviour record screens.

**Motivation and Context**
Sometimes Behaviour incidents spiral into something bigger, which requires the creation of a Note in the student profile. Rather than manually copying content to a new note, this automates it.

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2019-08-09 at 10 19 14 AM

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -27,6 +27,7 @@ v19.0.00
     Security
 
     Tweaks & Additions
+        Behaviour: new Copy To Notes feature
         Timetable Admin: included student reportable flag in student enrolment
 
     Bug Fixes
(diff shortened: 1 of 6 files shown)


SKuipers:
Looks good. You're still using some $connection2 and try/catch blocks, which we no longer need in v16+ 😄 but otherwise looks merge-able.

rossdotparker:
I'd generally opt to try my hand, but the start of year is really busy. If you are happy to push a commit as an example, I can then review, test and merge. Thanks : )

rossdotparker:
Code looks great, and it tests out nicely. I like the fact that you get the Behaviour category ID if available, but otherwise fall back to null. Better than my lazy solution! Thanks : )

PR #909, opened 2019-09-03
Students: improved display of Teachers' emails in student profile

rossdotparker:
**Description**
For various historical reasons, Tutors and Learning Support Assistants aren’t listed with teachers in the Student’s Teachers of the student profile. This PR unifies everything into a single list.

**Motivation and Context**
Some people weren't getting emails they ought to have.

**How Has This Been Tested?**
Locally and in production.

### DIFF
--- a/modules/Students/student_view_details.php
+++ b/modules/Students/student_view_details.php
@@ -627,7 +627,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Students/student_view_deta
 
                         //Get and display a list of student's teachers
                         echo '<h4>';
-                        echo sprintf(__('Teachers Of %1$s'), $row['preferredName']);
+                        echo __('Teachers Of {student}', ['student' => $row['preferredName']]);
                         echo '</h4>';
                         echo '<p>';
                         echo __('Includes Teachers, Tutors, Educational Assistants and Head of Year.');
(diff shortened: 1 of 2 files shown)


modules/Students/student_view_details.php
```diff
@@ -627,14 +627,34 @@
 
                         //Get and display a list of student's teachers
                         echo '<h4>';
-                        echo __("Student's Teachers");
+                        echo sprintf(__('Teachers Of %1$s'), $row['preferredName']);
```
SKuipers: This would be a good opportunity for the new translation format, which helps indicate the context of the string replacement.

```suggestion
                        echo __('Teachers Of {student}', ['student' => $row['preferredName']]);
```

modules/Students/student_view_details.php
```diff
@@ -657,45 +677,6 @@
                             echo '</ul>';
```
SKuipers: Now that Teachers, Tutors, Educational Assistants and Head of Year are in the mix, would it be helpful to indicate who they are in this list? At TIS we're running a modified version of this code that includes the class name next to the teacher. In this case, it could say the class name, or their role (Educational Assistant, Head of Year, etc.)

modules/Students/student_view_details.php
rossdotparker: Thanks for the suggestion. I've implemented this, and committed it.

modules/Students/student_view_details.php
rossdotparker: That is an interesting idea. One aim here was that you could copy and paste the whole list straight into an email client. Does your modified code support that? If so, it would be worth adding it in. Thanks!

SKuipers: Good question! I guess since modding that part many years ago no one here has been using it that way 😁 I’ll take a look and see if there’s a way to get the best of both worlds. Otherwise, looks great, going to squash and merge 👍

PR #910, opened 2019-09-04
Form Groups: added staff-only summary of Year Groups after Form Group…

rossdotparker:
**Description**
Adds a Year Group table, with student totals, after the Form Group table

**Motivation and Context**
For schools with vertical form groups (e.g. groups including students from multiple years), counting the number of students in a year group can get tricky. This change addresses that need.

**How Has This Been Tested?**
Locally and in production

**Screenshots**
Screenshot 2019-09-04 at 8 34 03 PM

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -30,6 +30,7 @@ v19.0.00
         System: added Maldivian Rufiyaa as a currency option
         System: update DatePicker to use gibboni18n date format
         Behaviour: new Copy To Notes feature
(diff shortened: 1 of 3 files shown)


rossdotparker:
You are right, HOY would be great. I've no time to add it, but could try and come back to it. Thanks for merging.

SKuipers:
Okay, sounds good, I'll see if I can add it in quickly (rather than put it on the never-ending todo list 😅).

PR #911, opened 2019-09-05
Attendance: add more options to control attendance prefilling

SKuipers:
This PR further refines the available settings for attendance by adding two options:

- A setting to enable the first class attendance taken in a day to optionally count as school-wide attendance.
  Screen Shot 2019-09-05 at 1 08 07 PM

  Which enables a checkbox on Take Attendance by Class:
  Screen Shot 2019-09-05 at 1 05 08 PM

- An option to disable the prefilling of an attendance code, applied to Present - Late by default.
  Screen Shot 2019-09-05 at 1 04 06 PM

The notable code and database changes in this PR are:
- Added and implemented a `recordFirstClassAsSchool` setting.
- Added a `prefill` column to the `gibbonAttendanceCode` table.
- Added a `gibbonRollGroupID` column to the `gibbonAttendanceLogPerson` table. This was necessary to track the roll group for each log, so the difference between recorded and prefilled logs could be determined.
- Updated the prefill logic for Take Attendance by Class and Take Attendance by Roll Group to ignore non-prefillable attendance codes.
- Refactored the `attendanceSettings_manage_addProcess.php` and `attendanceSettings_manage_editProcess.php` pages to use a gateway class.

_Best viewed with whitespace changes ignored_


rossdotparker:
Tested locally, and all looks good : ) I'll roll to production shortly, and let you know if we run into any problems. This is certainly one of our more complex modules.

PR #919, opened 2019-09-11
User Admin: add a View Role page in Manage Roles

SKuipers:
This PR:
- Adds a View action in the Manage Roles page. This displays a list of all the actions available to that role, as well as the users who have that role.
- Adds a `Manage Roles_viewOnly` option.
- Renames the `Manage Roles` action to `Manage Roles_all`.

**Motivation and Context**
This enables schools to keep a closer eye on what each role is able to do, and who has those capabilities. With the Manage Roles_viewOnly option, users can help audit roles in the system without needing access to change the roles or permissions.

**How Has This Been Tested?**
Tested locally and in production.

**Screenshots**
Screen Shot 2019-09-11 at 1 17 30 PM


CHANGEDB.php
```diff
@@ -923,4 +923,6 @@
...
 INSERT INTO `gibbonPermission` (`gibbonRoleID` ,`gibbonActionID`) VALUES ('001', (SELECT gibbonActionID FROM gibbonAction JOIN gibbonModule ON (gibbonAction.gibbonModuleID=gibbonModule.gibbonModuleID) WHERE gibbonModule.name='Attendance' AND gibbonAction.name='Consecutive Absences'));end
 ALTER TABLE `gibbonPerson` CHANGE `studentID` `studentID` VARCHAR(15) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL;end
+UPDATE `gibbonAction` SET name='Manage Roles_all', URLList='role_manage.php,role_manage_add.php,role_manage_edit.php,role_manage_delete.php,role_manage_duplicate.php,role_manage_view.php' WHERE name='Manage Roles' AND gibbonModuleID=(SELECT gibbonModuleID FROM gibbonModule WHERE name='User Admin');end
```
rossdotparker: Does the precdence of this action need to be incremented to 1, to show it is higher than the view only counterpart?

SKuipers: Well caught, yep it should be! Just pushed the change.

SKuipers:
Thanks! 😄 It's been on my wishlist for a while, today it got bumped up to a must-have.

CHANGEDB.php
SKuipers: Wait, changed the wrong precedence! One sec 😬

PR #920, opened 2019-09-12
Attendance: fix Set Future Absence to only allow future dates

SKuipers:
- Fixes the date check in attendance_future_byPersonProcess to not allow the current date.
- Applies the Date class `minimum()` method to the dates in Set Future Absence to ensure they only allow future dates:
  Eg:
  Screen Shot 2019-09-12 at 2 29 08 PM


SKuipers:
Hi Ross, this one should be ready to merge. I think in the long run restricting Set Future Absence back to future-only dates is a good idea.

PR #932, opened 2019-09-17
Timetable Admin: enable users who can search All Users to also see Left students

SKuipers:
Currently, there's no way to view a timetable or change class enrolment for students set to Left. The workaround is to set them to Full, make the changes, and set them back to Left, which is cumbersome and not a permission all staff members have.

The All Users option already exists, and enables searching for Full and Expected students. This PR extends it to also allow searching for Left students.

This PR:
- Updates the All Users option in View Timetable by Person.
- Removes the status check from View Timetable by Person (based on permissions & role category).
- Updates the All Users option in Course Enrolment by Person.
- Removes the status check from Course Enrolment by Person.
- Adds an additional `$gibbonRoleIDCurrentCategory == 'Staff'` check to the Gateway class to ensure the All Users option is only available to staff.

**Motivation and Context**
On occasion staff members need to lookup or change class enrolment for past students to update a report card or transcript, particularly if they leave part way through the year. Staff may also need to look at timetables for past students or staff.

**How Has This Been Tested?**
Tested locally.

**Screenshots**
Screen Shot 2019-09-17 at 2 35 05 PM


src/Domain/Students/StudentGateway.php
```diff
@@ -141,7 +141,7 @@ public function queryStudentEnrolmentByRollGroup(QueryCriteria $criteria, $gibbo
         return $this->runQuery($query, $criteria);
     }
 
-    public function queryStudentsAndTeachersBySchoolYear(QueryCriteria $criteria, $gibbonSchoolYearID) 
+    public function queryStudentsAndTeachersBySchoolYear(QueryCriteria $criteria, $gibbonSchoolYearID, $gibbonRoleIDCurrentCategory)
```
rossdotparker: Should the new param not be set to null by default? Or are there no current other use cases?

SKuipers: Good call. There weren't any other uses that I could see, but that's a good habit to get in. Will make that change.

PR #933, opened 2019-09-18
Theme change message fixed and added translation support

leofreitas:
**Motivation and Context**
When I changed active theme the message was "Uninstall was successful" and I changed for the default [success0] "Your request was completed successfully." and add translation support to message "The specified theme cannot be found or is active and so cannot be removed."

**How Has This Been Tested?**
Travis and Local

### DIFF
--- a/modules/System Admin/theme_manage_uninstallProcess.php
+++ b/modules/System Admin/theme_manage_uninstallProcess.php
@@ -67,9 +67,9 @@ if (isActionAccessible($guid, $connection2, '/modules/System Admin/theme_manage_
             }
 
             if ($orphaned != 'true') {
-                $URLDelete = $URLDelete.'&return=success0';
-            } else {
                 $URLDelete = $URLDelete.'&return=warning0';
+            } else {
+                $URLDelete = $URLDelete.'&return=success2';
             }
             header("Location: {$URLDelete}");
         }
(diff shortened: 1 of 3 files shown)


modules/System Admin/theme_manage_uninstallProcess.php
```diff
@@ -67,9 +67,9 @@
             }
 
             if ($orphaned != 'true') {
-                $URLDelete = $URLDelete.'&return=success0';
-            } else {
                 $URLDelete = $URLDelete.'&return=warning0';
+            } else {
+                $URLDelete = $URLDelete.'&return=success2';
```
SKuipers: Has this change flipped the error state? 🤔 It looks like $orphaned being true is a bad thing, so $orphaned != 'true' would be the success state?

SKuipers:
Well caught Leo! I've added one note about `theme_manage_uninstallProcess.php`, otherwise looks great.

SKuipers:
Hi @leofreitas. Any updates on this PR? It looks like the error state in `theme_manage_uninstallProcess.php` has been reversed in these changes. Can you confirm if this is intentional or not? Thanks!

leofreitas:
Hi @SKuipers, I think so. In both situations, deleting the theme in the database is the main action. The difference is that if $ orfaned = true, you do not have to print the file deletion message. And even $ orfaned = true, the action is a delete, and in this case it was successful. What do you think?

SKuipers:
Okay, I think that makes sense. I'm going to merge it and test. Thanks! 👍

SKuipers:
Yep, tested and works great. Thanks for the fix! 😃

PR #934, opened 2019-09-24
Students: add delete option for View Student Profile_fullEditAllNotes

SKuipers:
This PR:
- Fixes an bug in View Students where the `View Student Profile_fullEditAllNotes` permission was causing a blank page to appear.
- OOifies the Student Notes table and adds a delete action for users with `View Student Profile_fullEditAllNotes`
- Implements the delete action and updates URLList for `View Student Profile_fullEditAllNotes`


rossdotparker:
Gibbon used to have a note delete function, but at one of our schools notes went missing, and so we removed it. However, as this action is not the default action, I believe this is really useful to have. Thanks for the OOification work done at the same time : )

PR #936, opened 2019-09-26
Staff: add a Staff Coverage Summary report with total days covered

SKuipers:
This PR adds a Staff Coverage Summary report, similar to the Staff Absence Summary, which lists substitutes and a total of days covered per sub. The report can be filtered by month as well as by substitute, which then lists the individual coverage dates for that substitute. 

**Motivation and Context**
Useful for keeping track of hours for internal and external subs, particularly when tracking and paying external subs for their coverage time.

**How Has This Been Tested?**
Tested locally and in production.

**Screenshots**
Screen Shot 2019-09-26 at 3 26 43 PM


SKuipers:
I've added a separate Full Day Coverage threshold in Staff Settings, so that these can be calculated separately than absence hours. For example, a staff absence less than 2 hours may count as 0 in some schools, but a sub covering for 2 hours wouldn't count as 0, it would round up to 0.5 (if the Full Day threshold was higher).

rossdotparker:
Looking good! I think this is the final feature that might get this in use in production at ICHK Secondary.

PR #940, opened 2019-09-27
School Admin: add drag-drop Year Group ordering

SKuipers:
Updates the Manage Year Groups page as an example of drag-drop row ordering.

**Screenshots**
Screen Shot 2019-09-27 at 3 29 54 PM


rossdotparker:
Looking forward to testing this one in a few minutes. Feels like a massive leap forward in terms of usability. Here we go...

leofreitas:
That's awesome!

PR #941, opened 2019-10-02
System: retire the Smart Workflow Help feature

SKuipers:
Smart Workflow Help has been the lovely blue presence at the top of many a Staff member's dashboard that they scroll past and don't appear to use (or click dismiss on) 😆 The discussion came up in the developer channel to remove it, and has come up [in the forums recently](https://ask.gibbonedu.org/discussion/2629/smart-workflow), so I've created a PR to remove it system-wide. Thanks Ross for having created this feature with the aim of being helpful though 😃 it looks like it's time to retire this feature for v19.

Screen Shot 2019-10-02 at 11 05 51 AM

This leaves a number of strings now unused. Shall we clean these up in POEditor now too?


rossdotparker:
Looks good...and good riddance! Is it worth removing the relevant field from gibbonStaff?

SKuipers:
I had initially removed it, then reverted that change, because there is still `gibbon.sql` and `gibbon_demo.sql` data that references that field, so it was causing an install failure 🤔 I think perhaps we leave it and remove next version?

rossdotparker:
One option would be to remove those files. Another would be, as you say, delay to v20. Either way, I don't mind. Thanks!

SKuipers:
Okay, I recall something recently about not editing the sql files mid-version 😏 But yes, it'd be nice to not leave any loose threads, so I'll see about making this change.

rossdotparker:
Haha, you got me there! I think my point was to avoid making changes in those files when we don't have to, as we can rely on CHANGEDB.php to do that for us. But, if we have to, and can tie up loose ends, then I'm OK with that!

PR #966, opened 2019-12-03
Tables: add sticky header to DataTables without breaking legacy tables

SKuipers:
**Description**
I've whipped up a small tweak that enables us to use sticky headers on tables that have been refactored, ideally without breaking any legacy tables. Sticky headers are quite handy for long data tables, but don't work if any of the containing divs have a css overflow property. 

This workaround uses a Page template variable to disable the overflow properties once a data table is created. The workaround can be removed once we've ooified all the tables, but for now I think it gives us the best of both worlds.

**How Has This Been Tested?**
Locally and in production.

**Screenshots**
Sticky header on Manage Users:
Screen Shot 2019-12-03 at 12 54 54 PM

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -41,6 +41,7 @@ v19.0.00
         System: updated TinyMCE allowable HTML to include details, summary and code blocks
         System: improved the display of checkbox and radio list items in forms
         System: added Malagasy as a selectable Language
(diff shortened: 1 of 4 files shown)


rossdotparker:
Another great, incremental improvement...made possible by all that refactoring work. Thanks Sandra : )

PR #983, opened 2020-01-13
Library Catalog Edit

leofreitas:
**Description**
Value 'Available' deleted from array statusNotBorrowable because it's a value associated borrowable

**Motivation and Context**
When we use another language the value of the statusBorrowable field is not valid (screenshot).
Edit form does not load value from statusNotBorrowable field

**How Has This Been Tested?**
Local and Travis

**Screenshots**
library_one


SKuipers:
Hi Leo,

I'm not too familiar with the Library module (we don't use it in my school), but I think perhaps the `borrowable` Y/N flag could relate to whether the item can or cannot be borrowed in general, but the `statusBorrowable` relates to whether it can currently been borrowed/is available. So even if an item is `borrowable`, it can still have a status of "In Use" and is not always available 🤔 Maybe @rossdotparker can confirm. 

Thanks!

rossdotparker:
Hi Sandra, you are spot on. An item that is borrowable has a status that is generally set through the Lending section, whereas an item that is not borrowable generally has its status set in the Edit screen in Manage. The item can have any status, the borrowable control is more related to when and how that is set. Hope this helps!

leofreitas:
Hi guys, in this PR we have two issues: the first and most important is is related to support translation in the file library_manage_catalog_editProcess.php (line 105). 
```
if ($borrowable == 'Y') {
-    $status = $_POST['statusBorrowable'];
+    $status = 'Available';
} 
```
Whenever the value of $_POST['bookable'] is equals Y, the variable $status will have the value 'Available' because in the form this fiel is readonly, But if I use textfield value ($ _POST ['statusBorrowable']) in other languages, the value won't always be 'Available'.

The second (library_manage_catalog_edit.php:184) is simpler. I think that 'Available' cannot be applied if the item is not borrowed.

rossdotparker:
I can see what you are getting at with the second point. However, an item might not be borrowable, but could be available for long-term assignment to a user. In which case, it can be available, and then becomes In Use when it is deployed. As an example, a laptop may be purchased and set up ready for use, but not deployed to a user, and thus be "available" for future usage.

For the first one, we need to find a way to translate the value on screen, but ensure that the value submitted to the server is not translated. Using a hidden value ought to achieve this.

SKuipers:
Hi @rossdotparker thanks for confirming. @leofreitas I'm going to close this PR, it sounds like the observed borrowable status is the intended behaviour. Thanks!

PR #996, opened 2020-02-22
Allow the administrator to add custom fields to classes

leofreitas:
**Description**
Allow the administrator to add custom fields to classes. This new feature was added both at install and update.

**Motivation and Context**
With this feature we can create fields in the classes, making the classes more flexible and adaptable.

**How Has This Been Tested?**
Local and travis


SKuipers:
Thanks Leo for being incredibly patient with this PR. You've made some great looking changes here! As Ross mentioned in slack, it’d be amazing to transform this feature set into a more generalized system for custom fields. What are your thoughts? Would you be interested in developing this a bit further? I think a lot of the work you’ve already done in this PR still applies, with some tweaks to the actions and tables used.

Some thoughts for generalizing the feature set could be:
- Rename the existing gibbonPersonField table to gibbonCustomField and add a scope field to the table, with initial options of Person and Class.
- Update the ClassFieldGateway to use it as a CustomFieldGateway for the exiting table.
- Update the Manage Custom Fields action to move it to System Admin (under Extend & Update?).
- Tweak your class custom fields to use the generalized table.

The current custom field table is called gibbonPersonField, so to make this change we would need to update all references to gibbonCustomField. I think this would be a significant enough change to tackle in a separate PR to be able to test it thoroughly. Perhaps the table-renaming PR could come first, then modify this PR to be merged after that, so it layers on the class custom fields.

It sounds like a lot of work, but I think it'd be worth it, as there are a number of areas in the system that could use benefit from custom fields. 

Let me know what you think : )

leofreitas:
Transforming this feature set into a more generalized system for custom fields is really interesting. I'll work on this feature in another PR for the next version.

SKuipers:
Thanks Leo! Hopefully there's plenty of code in your current PR that can be re-used for the more general custom field feature, your code changes were looking really solid and well-thought-out 👍

SKuipers:
Hi Leo! We've recently been revisiting the custom field code, based on some overall needs to improve the flexibility of the codebase, and I've just added custom fields to courses and classes as part of these recent changes. Feel free to check out how it works and give a shout with anything you spot to fix or improve: https://github.com/GibbonEdu/core/pull/1281

leofreitas:
Hi Sandra! That's awesome!

PR #1003, opened 2020-03-31
Staff: add Family and Activities subpages to Staff Profile

SKuipers:
I had a couple feature requests for the staff profile and decided to ooify and templatify things along the way. It's not a complete refactor, but it's a step forward:

- Adds an Activity subpage that lists activities the staff member is part of
- Adds a Family subpage that lists families, adults and children for that staff member
- Refactors the staff overview and sidebar to use a template
- Adds a template for family details that the student profile can eventually also use
- Makes the blank slate a component template so it can be easily re-used

**Motivation and Context**
It's helpful to be able to look up the students that belong to a staff member, or to see which activities they run (especially if there is someone covering for them).

**How Has This Been Tested?**
Locally and in production.

**Screenshots**
Screen Shot 2020-03-31 at 1 46 54 PM


rossdotparker:
Another step in the right direction : )

PR #1004, opened 2020-04-01
Students: add a Withdraw Student page with the option to notify staff

SKuipers:
This PR adds an action under Admissions to allow privileged users to withdraw a student. 
Screen Shot 2020-04-01 at 2 51 43 PM

**Motivation and Context**
Currently the workflow to withdraw a student is to edit their status in User Admin. This can be counter-intuitive for some staff, or they may forget to update other fields such as end date and departure reason. 

This feature combines the student withdraw workflow into one page and adds optional student notes and notifications to help keep relevant parties informed. This includes a notification event, which allows admin to subscribe users to automatically receive all withdraw notifications.

**How Has This Been Tested?**
Tested locally.

**Screenshots**
Screenshot_2020-04-01 TEST - Gibbon - Students


rossdotparker:
Looks great, and I am going to merge it now, and then test it. I like the way endDate feeds into gibbonPerson, so that the userAdmin_statusCheckAndFix.php CLI script picks up on and adjusts status like usual. No config required : )

PR #1013, opened 2020-04-18
Staff: OOified tables and queries in staff_manage_edit

raynichc:
All tables have been OOified and queries (all but one small one) have been moved to gateways in staff_manage_edit.php.

**Motivation and Context**
Hackathon

**How Has This Been Tested?**
Locally

### DIFF
--- a/modules/Staff/staff_manage_edit.php
+++ b/modules/Staff/staff_manage_edit.php
@@ -66,7 +66,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Staff/staff_manage_edit.ph
                 echo '</div>';
             } else {
                 //Let's go!
-                $values = $staff->toDataSet()->toArray()[0];
+                $values = $staff->fetch();
                 $gibbonPersonID = $values['gibbonPersonID'];
 
                 if ($search != '' or $allStaff != '') {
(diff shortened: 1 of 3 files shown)


modules/Staff/staff_manage_edit.php
```diff
@@ -53,22 +57,16 @@
...
 
-            if ($result->rowCount() != 1) {
+            if ($staff->isEmpty()) {
                 echo "<div class='error'>";
                 echo __('The specified record cannot be found.');
                 echo '</div>';
             } else {
                 //Let's go!
-                $values = $result->fetch();
+                $values = $staff->toDataSet()->toArray()[0];
```
SKuipers, line 69: In this instance $staff is a result object, so we can still call fetch on it rather than converting to and from a data set :)
```suggestion
                $values = $staff->fetch();
```

modules/Staff/staff_manage_edit.php
```diff
@@ -157,140 +155,72 @@
...
+                        ->setURL('/modules/Staff/staff_manage_edit_contract_add.php')
+                        ->addParam('gibbonStaffID', $gibbonStaffID)
+                        ->addParam('search', $search);
+
+                    $table->addColumn('title', __('Title'));
+                    $table->addColumn('status', __('Status'));
+                    $table->addColumn('dates', __('Dates'))
+                        ->format(function ($row) use ($guid) {
+                            if ($row["dateEnd"] == '') {
+                                return dateConvertBack($guid, $row['dateStart']);
```
SKuipers, line 208: dateConvertBack is deprecated, there's a Format::date function to replace it, which has the same parameters minus the $guid. In this case there's also a Format::dateRange method which takes a start and end and automatically handles the conditional for checking if they're the same date :)

src/Domain/Staff/StaffContractGateway.php
```diff
@@ -0,0 +1,60 @@
...
+
+use Gibbon\Domain\Traits\TableAware;
+use Gibbon\Domain\QueryCriteria;
+use Gibbon\Domain\QueryableGateway;
+
+/**
+ * Staff Gateway
+ *
+ * @version v16
+ * @since   v16
```
SKuipers, line 30:
```suggestion
 * @version v20
 * @since   v20
```

raynichc:
Hi Sandra, I've made the requested changes.

PR #1017, opened 2020-04-18
Library Browse OOification

jimbo8098:
Following Gibbon hackathon work, here's the library's browse page which introduces OOified datatables on the Library browse page. This also includes the addition of a Gateway for the library module under the src/ dir. I did consider if it should be under the module/Library folder instead but I wasn't certain if the intention there was for that style to be used in cases where the module is not a core one.

Additionally, this adds a formatter for simple photos due to the userPhoto formatter insisting on the image being local. The photo formatter can use either external images or internal ones by simply looking for http/https within the imageLocation column in the gibbonLibraryItem table.

Unsure on the spacing you guys prefer here.


jimbo8098:
I don't, no. Is there one you would suggest? Would make sense for me to do it considering there's more PRs to come.

SKuipers:
Okay, no problem. I'll go ahead and merge this and tweak the formatting, then feel free to experiment with formatter options for vim on your end. I think there is also a GitHub robot somewhere we could install that formats PRs automatically, if we wanted to get fancy 😄

PR #1021, opened 2020-04-22
Library Manage Catalog

jimbo8098:
- Using DataTable object instead on the Library module's Manage Catalog page.
- Added a new query to the LibraryGateway
- Null-coalesce added to the delete process for the Library Catalog. There's already an if there to check what that value is supposed to do.
- Delete process for the Library Catalog now no longer passes you back
to the same search filter. I think that's right for this? I could be wrong though.

I'm going to work on getting some PSR-2 formatter going on my environment but if you guys beat me, can you format this for me?


jimbo8098:
Ah, forgot about those headers! That's them gone now. I've refactored the delete files too so they send back to the same filter on library_manage_catalog since you're making sense there.

SKuipers:
Thanks! Going to merge it now and test it out.

PR #1022, opened 2020-04-23
Codesniffer

jimbo8098:
I believe Sandra mentioned in passing that PSR-2 was something we could have some robot fix for us, this PR kind of spawned from that but also my own lack of tooling with a GUI-less editor. 

This PR introduces codesniffer alongside a composer script for testing for PSR-2 styling. In doing so, it also bumps a couple of other packages due to what I believe to be fuzzy package version references.

It bumps:

aura/sqlquery
container-interop/container-interop
league/container
tecnikcom/tcpdf
twig/twig

I'm not 100% on these packages so would appreciate if one of you guys could see if you can explain that change. I'm not exactly an expert with composer but I get the gist of it through the use of npm.

I can, by all means, revert those vendorfiles and look to add some kind of versionlock to composer.json. We might even find that the new packages work without a hitch and, in my short testing for my instance running, that checks out. Might be worth letting Travis do it's thing tbh! 🎉

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,11 +20,15 @@
         ],
         "test:phpunit": [
             "cd tests && ../vendor/bin/phpunit --verbose --configuration phpunit.xml"
+        ],
+        "test:codesniffer": [
+          "vendor/bin/phpcs --standard=PSR2 modules/Library/" 
         ]
     },
     "scripts-descriptions": {
         "test:codeception": "Initialize CI environment and run acceptance tests with Codeception.",
-        "test:phpunit": "Run unit tests with PHPUnit."
+        "test:phpunit": "Run unit tests with PHPUnit.",
+        "test:codesniffer": "Check code complies with PSR-2 standards"
     },
     "require": {
         "php" : "^7.0",
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,8 @@
         "phpoffice/phpspreadsheet": "^1.6",
         "phpoffice/phpexcel": "1.8.1",
         "fzaninotto/faker": "^1.8",
-        "mpdf/mpdf": "^8.0"
+        "mpdf/mpdf": "^8.0",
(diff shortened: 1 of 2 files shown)


SKuipers:
Hmm, there seems to be a couple strange things happening: it has deleted the i18n submodule, which is something we very much need (and isn't composer-related 🤔 ), and it's removing a bunch of tests and docs from folders, which is odd but shouldn't break anything. I wonder though, did you delete and re-install the vendor folder using composer? Otherwise, the composer.json and composer.lock look good, it doesn't appear to have modified any of the core dependencies. Travis just appears to have tripped up over the missing i18n submodule.

I use the squizlabs/php_codesniffer library on my machine by doing a composer global install and it has worked well. If we install it as a dev dependency here, is there a GitHub action that lets us run it automatically on PRs?

jimbo8098:
Yes, it's possible to run tests using codesniffer. At last commit, you can run `composer run-script test:codesniffer` and it'll check for PSR2 formatting on the entire codebase. The tool does include `vendor/bin/phpcbf` which can correct some of the issues. Codesniffer will also highlight warnings which generally appears to be cases where thre line is longer than 120 characters.

jimbo8098:
I *think* this syntax should work but it's not something I've tested before.

SKuipers:
Hi Jim! It looks like you've been tinkering with this PR, but I'm not sure that it's ready to merge 🤔 It's gone from 250 deleted files to 1300 added files, I think perhaps some dev dependencies got checked into version control. I'm seeing the Travis check is now failing with 81000 lines of errors from code-sniffer 😧

jimbo8098:
Gah! Okay let me redo. I was able to do this the other day. Will update later.

jimbo8098:
There, that looks better! Let me know what you think here. Nothings amended in .travis.yml but it'd be as simple as adding it to the composer tests to impose that standard.

PR #1031, opened 2020-05-01
OO: Library Lending

jimbo8098:
Adding OOified DataTable to the Library Lending page. There are two functional differences here:

1. Removing number column on the far left. I don't see a reason for it being there.
2. The row is NOT red if the item is on loan. I'm not sure if the current DataTable would support such an "optional" CSS class. If there is, lemme know and I'll update it.

A screenshot detailing the difference is below. The new table is above the old one here.

image

### DIFF
--- a/modules/Library/library_lending.php
+++ b/modules/Library/library_lending.php
@@ -19,6 +19,9 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use Gibbon\Forms\Form;
 use Gibbon\Forms\DatabaseFormFactory;
+use Gibbon\Tables\DataTable;
+use Gibbon\Domain\Library\LibraryGateway;
+use Gibbon\Services\Format;
 
 $page->breadcrumbs->add(__('Lending & Activity Log'));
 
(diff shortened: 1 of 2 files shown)


jimbo8098:
I could maybe look at re-adding the Number column if it's helpful but I'd need to do so in SQL since I don't think there's a way to do that on the current DataTable class.

rossdotparker:
Hi Jim, thanks for this : ) The ID feild is definitely useful, especially as for many cases it is not numeric, but rather alphanumeric (e.g. my laptop is MBP00319).

For the row colouring, check out lines 84-87 of the file linked below, which use an anonymous function and a CSS class (error) to colour a row red:

https://github.com/GibbonEdu/module-masteryTranscript/blob/master/Mastery%20Transcript/credits_manage.php

I wonder if the warning class (orange) is better for loans. Perhaps overdue loans could be red?

Cheers,

Ross

jimbo8098:
Hi Ross,

No worries!

The ID field is still there, I'm referring to a numerical field which
counted 1-10 as it went down the current 10 items on that paginated table.
Seemed pretty redundant unless there were perhaps multiple versions of the
same item but I'm such cases I'd expect the ID would be more beneficial. I
realise those terms are quite synonymous so I'd say check out the
screenshot below. The upper table is the DataTable object.

Your colouring makes sense too. From memory the only thing coloured is on
loan so I think it would be helpful to have some other colours. I'll look
into that.

jimbo8098:
Added new colouring now so when the item is pastDue, it will show using the "error" class. When it is on loan, as suggested, it will show using the "warning" class. All files committed here have been run through codesniffer to use PSR2 formatting.

rossdotparker:
Hi Jim, thanks for clarifying, and you are right, the number/count/# field is superfluous once the table is refactored, as the refactoring shows the total record count at the top, and I believe the count was there for similar ends.

rossdotparker:
I'll leave this with @SKuipers to review and merge in due course. Thanks for continuing to plug away at this : )

PR #1033, opened 2020-05-05
OOification: Library Lending Item

jimbo8098:
Finished a first pass of library_lending_item.php OOification. A couple of possibilities here still:

- There's another hard coded table above the log which shows the details of the item. I suspect this can be OOified too. Is this worthwhile? Would this need to be build through some kinda lower level DataTable class?
- There's a header on the log on the original version. Is there a way to add a title to those DataTables? I suspect there was but I didn't find anything on the docs.
- This table uses italicised text where the action header usually appears to show that the item is not available for sign out. I don't believe there's a way to do that with the current OO tables so in place of that I've used a div with error class. Is that best practice? Could switch it for a warning class maybe?

A couple of bugs I found too:

- When you try to open the library_lending_item_renew or _edit pages where no responsible person is set, an error occurs. As a workaround, I've set the actions to not show if this is the case since both the current version of those pages requires it.

### DIFF
--- a/modules/Library/library_lending_item.php
+++ b/modules/Library/library_lending_item.php
@@ -17,7 +17,9 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 */
 
+use Gibbon\Tables\DataTable;
 use Gibbon\Services\Format;
+use Gibbon\Domain\Library\LibraryGateway;
 
 //Module includes
 require_once __DIR__ . '/moduleFunctions.php';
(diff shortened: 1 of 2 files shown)


modules/Library/library_lending_item.php
```diff
@@ -129,169 +131,117 @@
             echo '<h3>';
             echo __('Lending & Activity Log');
             echo '</h3>';
```
SKuipers, line 133:
```suggestion
```

modules/Library/library_lending_item.php
```diff
@@ -129,169 +131,117 @@
...
+            $criteria = $gateway->newQueryCriteria(true)
+               ->filterBy('gibbonLibraryItemID',$gibbonLibraryItemID)
+               ->filterBy('name',$name)
+               ->filterBy('gibbonLibraryTypeID',$gibbonLibraryTypeID)
+               ->filterBy('gibbonSpaceID',$gibbonSpaceID)
+               ->filterBy('status',$status)
+                                ->fromPOST();
+            $item = $gateway->queryLendingDetail($criteria); 
+            $table = DataTable::createPaginated('lendingLog',$criteria);
+
```
SKuipers, line 145: $table->setTitle(__('Lending & Activity Log'));

modules/Library/library_lending_item.php
```diff
@@ -129,169 +131,117 @@
...
+              ->addParam('gibbonLibraryItemID')
+              ->addParam('name',$name)
+              ->addParam('gibbonLibraryTypeID',$gibbonLibraryTypeID)
+              ->addParam('gibbonSpaceID',$gibbonSpaceID)
+              ->addParam('gibbonLibraryItemEventID')
+              ->addParam('status',$status)
+              ->format(function($event,$actions) {
+                if($event['rowNum'] == 1 && $event['status'] != 'Returned')
+                {
+                  var_dump($event);
```
SKuipers, line 220:
```suggestion
```

SKuipers:
I've pushed a few post-ooficiation tweaks https://github.com/GibbonEdu/core/commit/610503942066fb53ff89458935a330da382d8849:
- ROW_NUMBER() appears to be new as of MySQL 8, and not supported by the minimum MySQL version for Gibbon which is still 5.6. I've removed it and adjusted the code.
- Editing strings that have a `__()` wrapped around them creates a new string to translate. I've tweaked a few columns to use the `description()` method and removed the edited strings (eg: `__('Status (Date In/Out)')`)
- Query methods that use a `$criteria` object can't have an explicit orderBy, otherwise the front-end table sorting doesn't work. Instead, you can use the sortBy() method on the `$criteria` object itself.
- The first parameter of `addColumn()` should match a valid field in the row data, this is what enables them to sort. Invalid column names will cause an SQL error which makes the whole table disappear when you click the sort arrows 😅

PR #1034, opened 2020-05-05
Table OOification: Library moduleFunction

jimbo8098:
This PR removes the moduleFunctions.php flat table within the Library module in favour of an OOified table where it was used. I've added this on both places it's used with a couple of changes.

- The image used is displayed by the Format::photo service instead which results in a slight visual difference. Since the getImage function (which was in use previously) is used elsewhere, I've not touched that.
- The borrow and return date should be easier to read now with the more important return date shown in bold first in that column but replaced with N/A if return is not expected (e.g. the item has already been returned).
- The borrow and return dates both use the Format service too so will be localised based on the formatting the user has chosen.
- SQL side timestamp comparison is used in place of the previous PHP based one. This should be less likely to be more robust.
- Now using Format service for the italicised text instead which should be more robust.
- Details section is now placed in an expander column instead of triggering manually added JS.

If it were possible to edit the moduleFunction to use the gateway that probably would have been nicer but I've had issues with that in the past. Either way, the general structure should be fine.

All changed files have been run through CodeSniffer so should already be compliant for the most part.


jimbo8098:
I noticed but when I looked again I reckoned it could be done. Turn that
frown upside down 😂

PR #1043, opened 2020-05-17
Library SessionTimeout

leofreitas:
**Motivation and Context**
Updated SessionTimeout with translation support and jQuery/jQuery UI dependency removed

**How Has This Been Tested?**
Local


SKuipers:
I've tested it out and it's working great 😃  I made one fix: it looks like the library uses miliseconds instead of seconds, so the session was timing out much quicker. Added a conversion here: https://github.com/GibbonEdu/core/commit/f6a33ad6d40c3c533351dc6e56d6708d5fd8b85b

leofreitas:
> I've tested it out and it's working great 😃 I made one fix: it looks like the library uses miliseconds instead of seconds, so the session was timing out much quicker. Added a conversion here: [f6a33ad](https://github.com/GibbonEdu/core/commit/f6a33ad6d40c3c533351dc6e56d6708d5fd8b85b)

Perfect!

PR #1052, opened 2020-06-01
System: add comment editor and code editor to the Form library

SKuipers:
This moves the CodeEditor and CommentEditor classes in to the core so they can be used by other modules. They're currently used by the Reports module, but also share functionality with Query Builder.

**Comment Editor:**
Will only check names and pronouns if you pass in the value to check against, otherwise off by default.

```
$col->addCommentEditor('comment')
    ->checkName($preferredName)
    ->checkPronouns($gender)
    ->maxLength(500)
```

Screenshot 2020-06-01 at 5 31 21 PM

**Code Editor:**
Currently handles `twig` and `mysql` modes.
```
$col->addCodeEditor('templateContent')
    ->setMode('twig')
    ->setValue($templateContent);
```

Screenshot 2020-06-01 at 5 31 47 PM


rossdotparker:
Looks great, thanks! I've made a not to update Query Builder to use the new codeEditor object. A fitting bit of rationalisation to round off the v20 development cycle : )

PR #1053, opened 2020-06-02
Timetable Admin: fix Course Enrolment by Class unable to delete duplicates

SKuipers:
On the off chance that duplicate enrolments end up in the Course Enrolment by Class page, either through a manual import or other tool, they are currently not deleteable. 

- For the Delete action, this is because the page looks at the gibbonPersonID, sees more than one entry for that student, and returns an error. 
- For the Bulk action, the result is even more alarming: instead of deleting the duplicate or not deleting anything, it deletes everything, both the original and duplicate records are deleted.

This PR changes the Course Enrolment by Class page to use the gibbonCourseClassPersonID when editing and deleting enrolments, to ensure that the duplicates can be removed without causing an issue with the original record.

**How Has This Been Tested?**
Locally.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -96,6 +96,7 @@ v20.0.00
         Timetable: escaped calendar event names in timetable tooltips
         Timetable: incremented gibbonTTImport course and class name field lengths
         Timetable Admin: implemented bulk Tie Days to Dates duplicate checking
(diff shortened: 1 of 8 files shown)


rossdotparker:
I can't tell you how many times I've encountered this issue over the past 8 years, and meant to fix it...but never quite got around to it. Another idiosyncratic Gibbonism fixed. Thanks Sandra.

PR #1057, opened 2020-06-05
Edit Invoicee Fix

jimbo8098:
I spotted an issue where, if the GET params for search and allUsers was not set when hitting this form, the form would show a couple of error messages. This fix simply null-coalesces the GET parameters so that if they are not provided, it is assumed they are empty strings. 
Fixes #1058

### DIFF
--- a/modules/Finance/invoicees_manage_edit.php
+++ b/modules/Finance/invoicees_manage_edit.php
@@ -30,7 +30,6 @@ if (isActionAccessible($guid, $connection2, '/modules/Finance/invoicees_manage_e
     echo __('You do not have access to this action.');
     echo '</div>';
 } else {
-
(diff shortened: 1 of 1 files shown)


leofreitas:
Great Jim, 
I have some suggestions:
Line 36: I would keep the pattern $gibbonFinanceInvoiceeID = $_GET['gibbonFinanceInvoiceeID'] ?? '';
Line 53: change to //Check if invoicee specified
Line 54: You can delete this line because it's redundant
Line 82: I would remove the char ;

Best regards

jimbo8098:
Makes sense, I'll make some changes soon.

jimbo8098:
Done! Re-merged with v20.0.00 too. Thanks @leofreitas !

PR #1060, opened 2020-06-05
System admin update

leofreitas:
**Description**
Changed the method to identify tables with engine other than innodb

**Motivation and Context**
If I have a view in my database, the system identifies it as a table, but a view does not have the engine attribute, so we cannot update it either.

Captura de Tela 2020-06-05 às 16 14 52

View
Captura de Tela 2020-06-04 às 20 19 15

**How Has This Been Tested?**
Local

### DIFF
--- a/modules/System Admin/update.php
+++ b/modules/System Admin/update.php
@@ -223,7 +223,7 @@ if (isActionAccessible($guid, $connection2, '/modules/System Admin/update.php') 
             $tablesInnoDB = 0;
             try {
                 $data = array();
-                $sql = 'SHOW TABLE STATUS';
+                $sql = "SELECT * FROM information_schema.tables WHERE table_schema = DATABASE() AND TABLE_TYPE = 'BASE TABLE';";
                 $result = $connection2->prepare($sql);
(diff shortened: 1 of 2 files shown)


SKuipers:
Great consideration, we certainly want to properly handle databases that have views in them. It looks like information_schema isn't guaranteed to always exist, particularly for shared hosts. A similar option may be to combine the `SHOW TABLE STATUS` command with `SHOW FULL TABLES` to see the the table type. Since this PR affects the updater and we're very close to the v20 release, lets hang onto this PR and test it some more so we can merge it into v21.

leofreitas:
Excellent observation! Let's test more.

SKuipers:
Hi Leo, it turns out in v21 we ended up refactoring the updater to add simple migration files, and in the process I believe we were able to address this issue. The majority of the updated code is in the `src/Database/Updater.php`, however the specific changes affecting this PR will be in the new `src/Database/Migrations/EngineUpdate.php` file. Feel free to check these out and let me know if the updater works and properly ignores the views. Thanks!

PR #1072, opened 2020-06-14
OO: Attendance - Consecutive Days Absence

jimbo8098:
OOification of the consecutive days absence report alongside a couple of usability improvements:

- Now complies with PSR2.
- Added a description to the form to make it easier to understand.
- Added a description to the new datatable.
- I believe this is a bugfix, feel free to correct me, but the table will now show any absences within those provided days. Previously it only showed absences if the student was absent for all of those days which I don't believe is what this is for.

A couple of suggestions:

- There is an inline gibbonPerson query which I'm not too fond of. I suspect this may be a common domain somewhere.
- I'm not sure it's best practice to run separate functions within the gateway so no gateway has been made for this report. What's the consensus here?
- I suspect it may be helpful to have some kind of group or student select in here too.


SKuipers:
> Previously it only showed absences if the student was absent for all of those days which I don't believe is what this is for.

Thanks for the ooification! I'll have to run this one by @rossdotparker, but I think the aim is to only show consecutive absences where all the days are absent. It's a use case for schools looking for specific attendance patterns I believe 🤔

rossdotparker:
Hi Sandra, you are exactly right, we are looking for strings of consecutive absences. It is to meet Hong Kong government regulations where a student absent for more than x consecutive days needs to be reported for followup.

SKuipers (approved):
Better late than never! 😬  I'm going to merge and test this. It looks like we'll need to tweak the logic back to only counting consecutive absences, but otherwise looks good. Thanks!

PR #1075, opened 2020-06-16
OO: Individual Needs moduleFunctions

jimbo8098:
This PR contains the OOified version of the moduleFunctions script for Individual Needs with a number of improvements.

- Now using DataTables in place of inline HTML for the most part.
- Error handling for archives (see note below)
- Gateway functions for alert levels and descriptors for students have been added. 
- sequenceNumber within alert and descriptor tables is now used as part of the IN gateway. This does affect other pages but will only order them based on the ingrained sequenceNumber instead of whatever order SQL outputs it in.

**Archives**
It was previously possible to have an archive show which would not actually tick any of the boxes giving the appearance that the student never had any IN attributes. The table will now check that the relevant descriptor and alert level exist and, if not, will display an error message (created using Format::alert) depending on what couldn't be found. An example of how this shows is below where both descriptor and alert cannot be found:

image

The archive functionality was tested using the following snippet at the top of the printINStatusTable function:

```
   $archiveArray = [
    [
      'gibbonINDescriptorID' => '005',
      'gibbonAlertLevelID' => '005'
    ]
  ];
  $archive = serialize($archiveArray);
```
I did test these side-by-side by resetting the archive using `$archive = serialize($archiveArray);` after echoing the table to make sure the format would be read as expected on both versions of the table.

**Suggestions**
- There is currently an inline formatter for checkboxes, this might be something worthwhile changing into a Format function instead, it may be a bit more universal that way.

### DIFF
--- a/modules/Individual Needs/moduleFunctions.php
+++ b/modules/Individual Needs/moduleFunctions.php
@@ -15,127 +15,117 @@ GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
-*/
+ */
+
+use Gibbon\Domain\IndividualNeeds\INGateway;
+use Gibbon\Services\Format;
+use Gibbon\Tables\DataTable;
+use Gibbon\Domain\DataSet;
 
 //$mode can be blank or "disabled". $archive is a serialized array of values previously archived
 function printINStatusTable($connection2, $guid, $gibbonPersonID, $mode = '', $archive = '')
 {
-    $output = false;
-
-    try {
-        $dataDescriptors = array();
-        $sqlDescriptors = 'SELECT * FROM gibbonINDescriptor ORDER BY sequenceNumber, nameShort';
-        $resultDescriptors = $connection2->prepare($sqlDescriptors);
-        $resultDescriptors->execute($dataDescriptors);
-    } catch (PDOException $e) {
-        $output .= "<div class='error'>".$e->getMessage().'</div>';
-    }
-
-    try {
-        $dataSeverity = array();
-        $sqlSeverity = 'SELECT * FROM gibbonAlertLevel ORDER BY sequenceNumber, nameShort';
-        $resultSeverity = $connection2->prepare($sqlSeverity);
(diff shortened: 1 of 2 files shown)


src/Domain/IndividualNeeds/INGateway.php
```diff
@@ -86,11 +86,38 @@ public function queryINBySchoolYear(QueryCriteria $criteria, $gibbonSchoolYearID
...
+
+      return $this->runQuery($query,$criteria);
+    }
+
     public function queryIndividualNeedsDescriptors(QueryCriteria $criteria)
     {
         $query = $this
             ->newQuery()
             ->from('gibbonINDescriptor')
+            ->orderBy(['gibbonINDescriptor.sequenceNumber'])
```
SKuipers, line 120: Anytime we have a $criteria object, it's best to not explicitly set the orderBy in the query, as this overrides the ability to use the front-end column sorting by clicking the column names. The alternative is to call sortBy() on the $criteria object as it's created.

modules/Individual Needs/moduleFunctions.php
```diff
@@ -15,127 +15,117 @@
...
-        $descriptors = array();
-        $count = 0;
-        while ($rowDescriptors = $resultDescriptors->fetch()) {
-            $descriptors[$count][0] = $rowDescriptors['gibbonINDescriptorID'];
-            $descriptors[$count][1] = $rowDescriptors['name'];
-            $descriptors[$count][2] = $rowDescriptors['nameShort'];
-            $descriptors[$count][3] = $rowDescriptors['description'];
-            ++$count;
+    $output = "";
+    global $container;
```
SKuipers, line 29: Hmmmm. I know this is needed, to make this particular function work, but I'm hesitant to start adding too many `global` calls to the codebase (although I have had to use them in the odd place too, as a transition to more oo code). Ideally, the whole function could be replaced by an object that is instantiated from the container.

SKuipers (approved):
Thanks Jim! I've added a couple comments as points to ponder, I'm going to merge and test this first, and then maybe look at the best way to handle DataTables/Forms in functions, ideally by moving them into their own objects.

PR #1077, opened 2020-06-21
OO: Student Activity Report

jimbo8098:
OOification of report_activityChoices_byStudent.php. Includes:

- Conversion of inline HTML table to DataTable objects.
- Error handling in cases where date-based activities are used as opposed to term based ones and vice versa on incompatible setups. More below.
- Using Format service throughout in place of old inline styling tags. A big one here is the date range formatting simplified from a big complicated substr comparison to Format::dateRangeReadable.
- Using isModal feature on action in place of direct link, should allow for resolutions to be a bit more dynamic as time goes on.
- There is now one less SQL query performed per row since I now use GROUP_CONCAT to provide the term names in place of a lookup on the terms per row returned.

**Suggestions**
- It is difficult to get from this page to the activity to correct any errors/change enrolment status. It might be advisable to link from this page to the enrolment page for easier access. It may also be handy to do the same for the activity itself.
- The issue with activities being assigned to a method opposing the set one is an unavoidable one if one changes the dateType. It would therefore be advisable to either display a warning when setting this (if possible) or handle it in code e.g. Term 1 <-> 01-07-2020 - 31-07-2020.
- It may be worthwhile to have a filter for the different years one might wish to filter on and a filter for cases where the activity is currently not active.

**Error Handling**

While investigating the difference between activities booked by term and date, I spotted that one could feasibly change the dateType from being terms (default) to dates and activities booked by both term and date would show in the table. Activities booked by term, whilst still showing, would show with their date being set to "December". This appears to be down to how the programStart/programEnd columns were being parsed(these were previously null, the default value in mysql).

To combat the possibility of confusion here I've added two things:

1. The row such errors occur on will show a Format::small message displaying the reason no date/term is shown in that box. This will be one of:
- Assigned to x - where the activity is assigned to x when the system is setup for the opposite/
- Not assigned to date range - where the activity is assigned to neither date nor term. This is, at present, theoretically impossible due to the table columns in mysql and is merely preperation in case this occurs.

2. An error alluding to the term/date not being set in the table below the message is shown using Format::alert. This shows above the offending table which should assist in diagnosis.

A picture of both versions "side-by-side" is shown below with the new version above the second repition of the SchoolYear title.

image


SKuipers (approved):
Thanks Jim! You've touched on something that certainly needs addressed, in that changing the term/day type once there are already activities in the system can mess up the activity system. I think you're right, there definitely needs to be better ways to handle and indicate this. I wonder if this type of error is best displayed on the Manage Activities level, rather than on each individual report. This way, the people who manage activities can find and edit the ones that need corrected, and the reports themselves (which there are several of) simply handle the missing data per column, as you've added. I'm going to merge and test your code, and hang onto the error checking logic for now, so that we can add it into v22 (since it also involves new strings, and we're past the string freeze). Thanks for these well thought out improvements!

PR #1081, opened 2020-07-21
Students: add an optional application fee that can be sent manually by staff

SKuipers:
This PR adds an optional application processing fee that can be enabled in the User Admin > Application Form Settings page. This fee can be sent manually after the application is submitted by editing the application through Manage Applications and clicking Send Payment Request. The processing fee can be used in addition to the fee submitted with the form, or on its own.

**Description**
- Updates the Application Form Settings page to add a new fee option, and differentiates the Submission Fee from the Processing Fee:
  Screen Shot 2020-07-21 at 7 27 00 PM

- Adds a Send Payment Request button to the top of the Edit Application Page:
  Screen Shot 2020-07-21 at 7 44 34 PM

- Adds a page to send the payment request to the parent and optionally edit the message text before sending:
  Screen Shot 2020-07-21 at 7 48 07 PM

- Sends an email to the parent with a link to the payment page:
  Screen Shot 2020-07-21 at 7 49 27 PM

- Adds a payment page which hooks into the PayPay payment scripts:
  Screen Shot 2020-07-21 at 7 20 36 PM

- Sends a receipt to the parent after successful payment, and bcc's a copy to the admissions email:
Screen Shot 2020-07-21 at 7 18 22 PM

- Updates the Edit Application page to include payment information.
- Updates the Application Form to include a message about the processing fee.

**Motivation and Context**
Schools may need to send more than one fee, or may only wish to send a fee after they have reviewed the application.

**How Has This Been Tested?**
Tested locally and with PayPal sandbox.


modules/Students/applicationForm_manage_edit.php
```diff
@@ -175,23 +182,25 @@
     // PAYMENT
     $currency = getSettingByScope($connection2, 'System', 'currency');
     $applicationFee = getSettingByScope($connection2, 'Application Form', 'applicationFee');
+    $applicationProcessFee = getSettingByScope($connection2, 'Application Form', 'applicationProcessFee');
+    $applicationFee = getSettingByScope($connection2, 'Application Form', 'applicationFee');
```
rossdotparker, line 186: Does this not exist twice now, as it is two lines above?

modules/Students/applicationForm_manage_edit_fee.php
```diff
@@ -0,0 +1,100 @@
...
+        echo Format::alert(__('You have not specified one or more required parameters.'));
+        return;
+    }
+
+    $application = $container->get(ApplicationFormGateway::class)->getByID($gibbonApplicationFormID);
+    if (empty($application)) {
+        echo Format::alert(__('The specified record does not exist.'));
+        return;
+    }
+
```
rossdotparker, line 53: Is it worth having a check to see if payment2 has been made yet, in which case the screen should not be available?

rossdotparker:
This is looking really good. I love the changes to /modules/User Admin/applicationFormSettingsProcess.php, which cuts down the amount of code so much, and really enhances readability.

I've left two comments. If you are OK to address those, then I'm happy to merge and do some testing, before rolling to production.

Thanks!

modules/Students/applicationForm_manage_edit.php
SKuipers: Right, well caught! Must have double-tapped the duplicate line command.

modules/Students/applicationForm_manage_edit_fee.php
SKuipers: Yep, good call, I'll add the check here too.

PR #1084, opened 2020-07-29
OO: Student View Details - Activity

jimbo8098:
OOifying Student View Details page's Activity subpage. Some other improvements bundled too:

- Moving to common DataTable object in lieu of inline tables
- Using isModal() on actions to prefer the users modal size over the preset.
- Using Format::dateReadableRange in place of previous reused code block.
- Term concat is now done in SQL rather than PHP which should be marginally quicker.
- Using setTitle in place of basic inline h3 tags

**One unrelated fix** but so small I don't think it really requires a separate PR, now using setTitle on the student notes subpage.

All changes are PSR2 compliant according to PHPCS

### DIFF
--- a/modules/Students/student_view_details.php
+++ b/modules/Students/student_view_details.php
@@ -1856,11 +1856,8 @@ if (isActionAccessible($guid, $connection2, '/modules/Students/student_view_deta
                                 if ($resultCategories->rowCount() > 0) {
                                     $categories = true;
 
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Looking good! 👍 I've tweaked a couple strings in the column names so they match the original strings, I'm going to merge and test this. Thanks!

PR #1088, opened 2020-08-19
Planner: add Daily Attendance Summary CLI scripts for parents, tutors and heads of year

SKuipers:
This adds two new CLI scripts for sending daily attendance information. The first script sends one email to each family with an overview of their child/children's attendance for the day. The second script sends an overview of form group attendance to form group tutors, and year group attendance to heads of year.

This PR adds two new settings in School Admin > Planner Settings to customize the intro and postscript of the emails sent to parents. Like other CLI scripts, this PR also adds notification events for receiving a result of all the emails sent per day, which sends to the administrator by default. 

**Motivation and Context**
Schools may want to keep a closer eye on attendance, or may want to share daily attendance information with parents.

**How Has This Been Tested?**
Tested locally.

Note: Testing this CLI is very spammy. You can add a line like `$families = array_slice($families, 0, 2);` or `$yearGroups = array_slice($yearGroups, 0, 1);` to trim down the results when testing.

**Screenshots**
Daily Attendance Summary for Parents:

Screenshot 2020-08-19 at 12 06 04 PM

Daily Attendance Summary for Tutors:
Screenshot 2020-08-19 at 12 11 33 PM


cli/planner_parentDailyEmailSummary.php
```diff
@@ -0,0 +1,163 @@
...
+
+    $body = $parentDailyEmailSummaryIntroduction;
+    $body .= '<br/><br/>';
+    $body .= $content;
+    $body .= '<br/>';
+    $body .= $parentDailyEmailSummaryPostScript;
+
+    // Add recipients and sender
+    foreach ($familyAdults as $adult) {
+        $mail->AddAddress($adult['email'], Format::name('', $adult['preferredName'], $adult['surname'], 'Parent', false, true));
```
rossdotparker, line 122: Does $mail->AddAddress remove previously added addresses? I don’t recall? The $mail initiation happens outside the loop, so is there a risk that we keep accumulating addresses, one family after another?

cli/planner_parentDailyEmailSummary.php
```diff
@@ -0,0 +1,163 @@
...
+    }
+    $mail->setDefaultSender($subject);
```
rossdotparker, line 124: Adding the $subject as the default sender?

cli/planner_parentDailyEmailSummary.php
rossdotparker: Right, I've just seen the clear down below! Ignore.

cli/planner_parentDailyEmailSummary.php
SKuipers: Aha right, in hindsight that method signature doesn't look quite right with the method name. The default sender is already in session data, so it's setting that, then prepending the organisationNameShort to the beginning of the subject:

```
public function setDefaultSender($subject)
    {
        $this->Subject = $this->session->get('organisationNameShort').' - '.$subject;
        $this->SetFrom($this->session->get('organisationEmail'), $this->session->get('organisationName'));
    }
```

rossdotparker: Right, makes sense! Thanks.

PR #1092, opened 2020-08-27
Attendance: add an Offsite - Late option for attendance codes

SKuipers:
This one goes under "things you never knew you needed until 2020" 😬 

**Description**
Adds an "Offsite - Late" option for attendance codes and updates scripts so this attendance type is coloured in orange to match Onsite - Late.

**Motivation and Context**
With campus closures, some schools are still taking attendance for online classes. In this case, the Onsite - Late option doesn't work as it will record the student as on campus for purposes of reports, etc. The Offsite - Late option helps fill the need for tracking lates for online classes.

**How Has This Been Tested?**
Tested locally.

**Screenshots**
Screenshot 2020-08-27 at 9 06 17 AM


rossdotparker:
Looking good, and seems like you've updated the same logic in many places..hopefully all of them, but I can't say for sure. We'll know after some beta testing!

PR #1095, opened 2020-09-02
System Admin: add an Email Templates section to customize email text

SKuipers:
Adds an Email Templates page to System Admin, as well as a `gibbonEmailTemplate` table in the database. Templates can be edited using Twig template syntax. I've added the Sent Reports to Parents template as an example, more to come.

The template editing screen includes a list of any variables available for the template, as well as an option to send a test email (which is formatted with fake data).

**Motivation and Context**
Many schools prefer to edit the text that is sent to parents and other users of the system, but currently that requires editing the codebase. These templates should offer schools a flexible way to customize the emails they send. This can be applied to other areas that we often get requests to customize, including acceptance emails.

**How Has This Been Tested?**
Tested locally.

**Screenshots**
Screenshot_2020-09-02 DEMO - Gibbon - System Admin(1)

Screenshot_2020-09-02 DEMO - Gibbon - System Admin


rossdotparker:
Another massive leap forward in terms of flexibility and customisation! A headline feature for v21...

SKuipers:
Thanks! 😃 Yep, hopefully the twig syntax will be simple enough for people to use but powerful enough for us to enable them to really customize these emails to their school's needs.

PR #1097, opened 2020-09-02
Discussion Twig Timestamp

powertashton:
**Description**
Adding timestamps to the timetable twig template

**Motivation and Context**
Gibbon Helpdesk Module is currently using the discussion template for it's discussion pages, and a timestamp felt like an essential add.

**How Has This Been Tested?**
Localhost environment, Mac OS Catalina MAMP with PHP 7.1.0 and Apache

**Screenshots**
Previously, using the type tag for the timestamp within the helpdesk module:

image

Now with the type tag used to indicate type, and the timestamp in it's own spot:
image

Additional example of the Free Learning Module discussion with a timestamp.
image

Hopefully this time the pull request is to the right branch...


SKuipers:
Hi Ash, great to see the timestamp in there. I wonder if we can borrow some UI styles from GitHub for the placement of the time. This is an example of how the staff coverage handles it (I notice now that I need to add the little arrows back in):

Screenshot 2020-09-03 at 12 20 01 PM

Template wise, it could accept `action` and `timestamp` variables, and if there's no action, default it to a single ASCII mid-dot (twitter-style). What do you think? Its a bit more succinct, and closer to what we're familiar with for other apps.

powertashton:
image
image
Something along the lines of this? 
I think instead of defaulting to a single ASCII mid-dot (twitter-style, as you say), it actually still works when it's just the timestamp by itself, thanks to the whitespace-no-wrap class, but that is just my opinion of it.

I also went and enabled the arrows to show again, they would've disappeared (for users without a user profile image) when you added the fallback image, this way there's always an arrow, as there's always a profile image (even if it's a fallback one)

SKuipers (approved):
Nice! Looks great, going to merge and test! 👍

PR #1098, opened 2020-09-03
Reports: add a Send Reports tool to bulk-send templated emails with download links

SKuipers:
This is a big one 😅 This PR adds a tool for sending reports via email to parents and students. This can be useful for students who have left the school, or just as an alternate means of publishing reports.

Features:
- Reports sent via email include a link to directly download the report from Gibbon using a unique access token. The link expires after one week. Re-sending the report resets the expiry date.
- Only reports that are generated as Final and in an archive accessible by parents & students can be sent via email.
- Reports sent via email use the new Email Templates to enable customizing the email content. Eg:
  Screenshot 2020-09-03 at 9 47 48 AM

- Reports can be sent to parents or students and use a background process for sending. The email will only include a View Online option if the user is Full and had Can Login access.
- Reports downloaded via email still update the Read indicator, so you can see that a parent has access the report either in Gibbon or via email.
- The Send Reports page indicates Sent and Read status with checkmarks and tooltips: Green checkmark for sent, Double check for sent and read, and faded check for already read online. This helps admin target users who have not received or read their reports.
  Screenshot 2020-09-03 at 10 48 34 AM

This PR also:
- Updates the familyContacts template so it can be used without displaying phone numbers.
- Updates the email.twig.html template to add an option for a second button.
- Adds new return codes for expired links and background processes, which can be used globally by any relevant process.
  Screenshot 2020-09-03 at 11 41 59 AM
  Screenshot 2020-09-03 at 10 28 42 AM

Notably:
- This PR also tweaks the Login form so that it can pass along GET variables after logging in. This is useful because any user clicking a link to a page and then needing to login will usually see an error after login because only the q parameter was passed along and any ID numbers from that link have been lost. I have tested and pondered this and I don't see any issues, as the http_build_query will escape values. Let me know what you think. This also addresses errors that users see after clicking a notification link when they're not logged in.

**How Has This Been Tested?**
Locally.


rossdotparker:
Hi Sandra, this looks great. I am just about to dig in. I think the login parameter passing sounds good. Does this apply to Sign in with Google? I believe this does not even pass q at present...

PR #1099, opened 2020-09-03
Added Header Actions to Details Tables

raynichc:
**Description**
Implemented Header Action rendering for Details Tables. Detail Tables without Header Actions should look the same.

**How Has This Been Tested?**
Locally on Detail Tables with and without Header Actions, in both core and Help Desk module.

**Screenshots**
Example of simple details table with Header:
image
Example of simple details table without Header (i.e. no change):
image


SKuipers (approved):
Nice, thanks Ray, looks great!

Btw, you can add the `displayLabel` method to your header actions if you'd like both the icon and the name to show up.

PR #1102, opened 2020-09-09
Tests: Simple bug fixes and fallback environment variables

powertashton:
**Description**
Added fallback environment variables if .env is not defining DB details
Changed HTTP to HTTPs, so SSL enabled environments don't fail
Added Users group and changed reference in acceptance.suite.yml so that the acceptance suite does not fail

**Motivation and Context**
These changes are required for tests to run with minimal setup, otherwise one would need to do additional setup that may take hours to figure out (Source: it took me hours)

**How Has This Been Tested?**
localhost, mac os catalina, MAMP environment using PHP 7.1.0 and Apache, with phpunit and codeception installed globally.
Ran the install suite to set up gibbon, and everything worked. Acceptance tests run with some minimal failures/errors, but presumably that's an issue with the tests or source code, and not my changes.
**Screenshots**

Install suite success and verbose process, using `codecept install run --env env -vvv` 
image


SKuipers (approved):
Thanks! These look good, I'll merge and test them out.

PR #1111, opened 2020-09-14
Tests: Simple fixes

powertashton:
**Description**
Slight changes so that the tests all pass, including changing the password to be compliant with the higher password requirement setting necessitating a non alphanumeric symbol that is set in an earlier test, and fixing some selectOptions

**Motivation and Context**
It wasn't working before

**How Has This Been Tested?**
Locally, using phpunit and codeception to run tests using the default phpbrowser.

**Screenshots**
All tests passing :) ;
image


SKuipers (approved):
Thanks, great fixes! 👍

PR #1112, opened 2020-09-17
Breadcrumb design pattern - W3C

leofreitas:
**Description**
Using W3C WAI-ARIA Design Patterns to make breadcrumbs more accessible.

**Motivation and Context**
Accessibility

**How Has This Been Tested?**
Local


yookoala:
I believe the `<nav>` tag is usually used as container of navigation menu. I've never seen breadcrumb being wrapped with `<nav>` before. Are you sure this is a preferable approach?

leofreitas:
Official W3C Example (https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html)

SKuipers (approved):
Thanks Leo! I've checked out the W3C article you shared and learned something new! 😃 I'm going to merge this now and test it out 👍

PR #1114, opened 2020-09-19
Space form factory

### LINKED ISSUE #1108: Facility (Space): Database Form Factory; createSelectSpace does not allow for sorting by type
**Is your development discussion related to a problem?**
I use the "type" column of gibbonSpace to define the building that a location is within, as we have a lot of different facilities and it is easiest to sort each location, by their parent location. However selects that use the database form factory's ->addSelectSpace do not allow for any organisation beyond just listing all the locations out, which could be quite frustrating.

**Describe the solution you're proposing**
```php 
$facilityGateway = $container->get(FacilityGateway::class);
    $criteria = $subcategoryGateway->newQueryCriteria(true)
            ->sortBy(['type', 'name'])
            ->fromPOST();
    $facilityData = $facilityGateway->queryFacilities($criteria);
    
     $row = $form->addRow();
        $row->addLabel('gibbonSpaceID', __('Facility'));
        $row->addSelect('gibbonSpaceID')
            ->fromDataSet($facilityData, 'gibbonSpaceID', 'name', 'type')
            ->placeholder()
            ->isRequired();
```
This is the current workaround I've used (In the v1.3 version of the HelpDesk Module), which uses a regular select instead, but I know (for example) addSelectStudent does allow for the same sort of sorting from the form factory.

**Describe alternatives you've considered**
I've also considered instead of using grouped sorts, the addSelectSpace could be two selects chained together, however the implementation of this may require additional work.

**Additional context**
Without sorting
image
With sorting
image

powertashton:
**Description**
Adds the ability for grouping by type in the createSelectSpace, as a default.

**Motivation and Context**
Fixes #1108

**How Has This Been Tested?**
Locally on both the library module and helpdesk module selectSpaces

**Screenshots**
image


SKuipers (approved):
Looks good, I'm going to merge and test 👍  Thanks!

PR #1115, opened 2020-09-20
Staff: removed roles-as-type in Manage Staff

rossdotparker:
**Description**
This PR restricts the Staff type field choice back to only Teaching and Support, setting all non-Teaching users to Support. It may disruptive for some schools who have come to use roles as types.

**Motivation and Context**
At some point (many version ago) we added roles with category staff to the list of staff types: this seemed like a good idea at the time, but means that you can't say categorically who should get access to certain teaching-related functionality. It is important, in many schools, to know this, and so we are reverting to the old schema.

**How Has This Been Tested?**
Locally


SKuipers (approved):
This is looking good, I'm going to merge and test 👍

PR #1116, opened 2020-09-23
Data Updater: auto-accept data updates with no changes

SKuipers:
This PR changes the data update process so that a data update with no changes is automatically accepted and does not generate a notification. It does this by comparing the new values to the old values when processing the request.

**Motivation and Context**
It is good to encourage people to submit a data update even if there are no changes, so that the system knows their data is up to date. However, accepting all these changes can be tedious. This PR aims to save some time for admin.

**How Has This Been Tested?**
Tested locally.


rossdotparker:
Looking good @SKuipers!

PR #1117, opened 2020-09-25
System: fix various PHP Notices and Warnings

SKuipers:
This PR fixes a number of warnings, which helps to clean up the error log so the actual errors are more visible. Most warnings were fixes by a null coalesce or an empty() check.


rossdotparker:
It's great to see Gibbon getting more concise with these newer PHP shorthands.

PR #1118, opened 2020-09-30
Timetable Admin: add a dateEnrolled and dateUnenrolled to class enrolments

SKuipers:
This PR adds dateEnrolled and dateUnenrolled fields to the gibbonCourseClassEnrolment table. It updates these dates when enrolling as well as syncing or auto-enrolling. The dates 

**Motivation and Context**
We've discovered a few instances where it is helpful to know when a student joined or left a class, for the purpose of filtering relevant lessons and homework for that student.

**How Has This Been Tested?**
Locally.

**Screenshots**
Screenshot 2020-09-30 at 2 33 30 PM

### DIFF
--- a/modules/Students/studentEnrolment_manage_editProcess.php
+++ b/modules/Students/studentEnrolment_manage_editProcess.php
@@ -114,6 +114,17 @@ if ($gibbonStudentEnrolmentID == '' or $gibbonSchoolYearID == '') { echo 'Fatal 
                             exit;
                         }
 
+                        // Update existing course enrolments for new Roll Group
+                        $data = array('gibbonStudentEnrolmentID' => $gibbonStudentEnrolmentID, 'dateEnrolled' => date('Y-m-d'));
+                        $sql = "UPDATE gibbonCourseClassPerson 
+                                JOIN gibbonStudentEnrolment ON (gibbonCourseClassPerson.gibbonPersonID=gibbonStudentEnrolment.gibbonPersonID)
+                                JOIN gibbonCourseClassMap ON (gibbonCourseClassPerson.gibbonCourseClassID=gibbonCourseClassMap.gibbonCourseClassID 
+                                    AND gibbonCourseClassMap.gibbonRollGroupID=gibbonStudentEnrolment.gibbonRollGroupID)
+                                SET gibbonCourseClassPerson.role='Student', gibbonCourseClassPerson.dateEnrolled=:dateEnrolled, reportable='Y'
+                                WHERE gibbonStudentEnrolment.gibbonStudentEnrolmentID=:gibbonStudentEnrolmentID
+                                AND gibbonCourseClassPerson.gibbonCourseClassPersonID IS NOT NULL";
+                        $pdo->executeQuery($data, $sql);
+
                         // Add course enrolments for new Roll Group
                         $data = array('gibbonStudentEnrolmentID' => $gibbonStudentEnrolmentID, 'dateEnrolled' => date('Y-m-d'));
                         $sql = "INSERT INTO gibbonCourseClassPerson (`gibbonCourseClassID`, `gibbonPersonID`, `role`, `dateEnrolled`, `reportable`)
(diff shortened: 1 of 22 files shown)


rossdotparker:
Enrolled or Enroled? I am never sure! I understand the former us US, the latter UK. Gibbon tends to use UK behind the scenes, and so in most places in the database it is enrol.

modules/Students/studentEnrolment_manage_editProcess.php
```diff
@@ -115,9 +115,9 @@
                         }
 
                         // Add course enrolments for new Roll Group
-                        $data = array('gibbonStudentEnrolmentID' => $gibbonStudentEnrolmentID);
-                        $sql = "INSERT INTO gibbonCourseClassPerson (`gibbonCourseClassID`, `gibbonPersonID`, `role`, `reportable`)
-                                SELECT gibbonCourseClassMap.gibbonCourseClassID, gibbonStudentEnrolment.gibbonPersonID, 'Student', 'Y'
+                        $data = array('gibbonStudentEnrolmentID' => $gibbonStudentEnrolmentID, 'dateEnrolled' => date('Y-m-d'));
+                        $sql = "INSERT INTO gibbonCourseClassPerson (`gibbonCourseClassID`, `gibbonPersonID`, `role`, `dateEnrolled`, `reportable`)
```
rossdotparker, line 119: When a student is enroled, should it set dateUnenrolled to null, just in case it is the second time a student is enrolled in the class (happens, but not very often), and so the old unenroll ends up older than the new enroll?

modules/Timetable Admin/courseEnrolment_manage_byPerson_edit_editProcess.php
```diff
@@ -64,9 +65,11 @@
                     header("Location: {$URL}");
                 } else {
                     //Write to database
+                    $dateEnrolled = $role != $values['role'] && stripos($role, 'Left') === false ? date('Y-m-d') : $values['dateEnrolled'];
```
rossdotparker, line 68: Can you explain the relationship between role and un/enroll date? Not sure I get it!

modules/Timetable Admin/course_rollover.php
```diff
@@ -193,17 +193,15 @@
...
-            } catch (PDOException $e) {
-                echo "<div class='error'>".$e->getMessage().'</div>';
-            }
-            if ($resultNext->rowCount() == 1) {
-                $rowNext = $resultNext->fetch();
-            }
+            
+            $dataCurrent = array('gibbonSchoolYearID' => $_SESSION[$guid]['gibbonSchoolYearID']);
+            $sqlCurrent = 'SELECT * FROM gibbonSchoolYear WHERE gibbonSchoolYearID=:gibbonSchoolYearID';
+            $rowCurrent = $pdo->selectOne($sqlCurrent, $dataCurrent);
```
rossdotparker, line 199: Does this not have a deleted section somewhere about? I can see where $dataNext is deleted from, but not $dataCurrent?

SKuipers:
> Enrolled or Enroled? I am never sure! I understand the former us US, the latter UK. Gibbon tends to use UK behind the scenes, and so in most places in the database it is enrol.

I wondered that too! I think by adding the ed, even when the root word is enrol, the past tense is still enrolled:

Screen Shot 2020-10-04 at 3 36 21 PM

modules/Timetable Admin/course_rollover.php
SKuipers: No deleted section for this one. This was added to grab the school year info to use the lastDay value, which as it happens I removed because setting all the classes to unenrolled first didn't seem useful in the end. I'll tidy it u and remove this section.

modules/Timetable Admin/courseEnrolment_manage_byPerson_edit_editProcess.php
SKuipers: It's looking to see if the role has changed when submitting the form, eg: to prevent changing the enrol date when only the reportable flag has been changed. For this logic: if the role has changed and the new role is not left, then update the enrolled date, otherwise keep it the same. This happens when setting a left enrolment back to non-left.

modules/Students/studentEnrolment_manage_editProcess.php
SKuipers: Hmm, yep most of the cases are doing that, using an UPDATE statement. Looks like this case of auto-enrol hasn't been checking for existing enrolment, so I'll add that and set it to reset the enrolment date.

PR #1120, opened 2020-10-09
Format: nameLinked

powertashton:
**Description**
This PR adds the formatter function for a linked profile

**Motivation and Context**
Discussed on slack with @SKuipers after encountering the need for a name to link to a profile. Current method of doing so involves nesting a format::name in a format::link with a few extra steps to define the URL etc, this automatically does it based off of the role type in one function.

**How Has This Been Tested?**
Locally, using the development helpdesk module as a playing ground


SKuipers (approved):
Nice! Looks like a handy utility method 👍

PR #1121, opened 2020-10-09
Login: Custom Loader

powertashton:
**Description**
Adds a custom loader on the login page, so as to allow schools to add additional login handling methods such as LDAP connections

**Motivation and Context**

My setup involves an LDAP login to facilitate users having the same password across all systems, which makes upgrading a slightly tricky business as I'd hard coded the code into login.php previously and would have to remerge these changes each upgrade

**How Has This Been Tested?**

Locally, using a custom LDAP authentication that had previously been merged into the main login.php,


login.php
```diff
@@ -46,8 +49,23 @@
...
+    // Custom login loader
+    if (!$session->exists('login_custom.php')) {
+        $globals = [
+            'guid'        => $guid,
+            'connection2' => $connection2,
+            'username' => $username,
+            'password' => $password
+        ];
+
+        $session->set('login_custom.php', $page->fetchFromFile('./login_custom.php', $globals));
```
SKuipers, line 62: I wonder if `fetchFromFile` may not work as you're expecting for this situation. It's primary use is for templating, as it runs some code and returns the string output. This line of code would run the PHP in login_custom.php and only cache the output result. Then line 66 would echo the cached output without running the PHP code.

Since your aim is likely to run the PHP code for every login execution, a `require_once` may be better in this instance.

powertashton: That would make a lot more sense, you're right. Pushing the changes now!

powertashton:
Travis :( 
Will take a look at what's going wrong with him tomorrow when I actually have access to my LDAP server.

login.php
```diff
@@ -46,8 +49,16 @@
     header("Location: {$URL}");
     exit;
 }
-//VALIDATE LOGIN INFORMATION
 else {
+    // Custom login loader
+    if (!$session->exists('login_custom.php')) {
```
SKuipers, line 54: Because there's nothing to cache, using the session here doesn't make sense anymore. This may work better as a file_exist check, then require the login script if it does.

Btw, how does the login scrips validate/invalidate login access if it's not interacting with the rest of the script? 🤔

login.php
```diff
@@ -46,8 +49,16 @@
...
+        require_once './login_custom.php';
+    }
+
+    if ($session->has('login_custom.php')) {
+        $page->write($session->get('login_custom.php'));
```
SKuipers, line 59: Also here, as there isn't necessarily anything being cached or written to the page, right?

SKuipers (commented):
Getting there!

login.php
powertashton: The login script I'm using validates by taking the $username and $password variables from the main login script, and then processes them to then create an account (if the user doesn't exist already) or update the password, so;
```if ($valid) {
    $data = array('username' => $username);
    $sql = "SELECT gibbonPerson.username FROM gibbonPerson WHERE (username=:username)";
    $result = $connection2->prepare($sql);
    $result->execute($data);
        // add a user if not there so that the rest of the login code will work
    if ($result->rowCount() == 0) {
        $salt = getSalt();
        $passwordStrong = hash('sha256', $salt.$password);
        $sql = "INSERT INTO gibbonPerson (username, passwordStrong, passwordStrongSalt, gibbonPersonID, firstName, surname, title, officialName, preferredName, nameInCharacters, gender, passwordForceReset, status, canLogin, gibbonRoleIDPrimary, gibbonRoleIDAll, lastIPAddress, address1, address1District, address1Country, address2, address2District, address2Country, phone1Type, phone1CountryCode, phone1, phone3Type, phone3CountryCode, phone3, phone2Type, phone2CountryCode, phone2, phone4Type, phone4CountryCode, phone4, website, languageFirst, languageSecond, languageThird, countryOfBirth, birthCertificateScan, ethnicity, citizenship1, citizenship1Passport, citizenship1PassportScan, citizenship2, citizenship2Passport, religion, nationalIDCardNumber, nationalIDCardScan, residencyStatus, profession, employer, jobTitle, emergency1Name, emergency1Number1, emergency1Number2, emergency1Relationship, emergency2Name, emergency2Number1, emergency2Number2, emergency2Relationship, studentID, lastSchool, nextSchool, departureReason, transport, transportNotes, calendarFeedPersonal, viewCalendarSchool, viewCalendarPersonal, viewCalendarSpaceBooking, lockerNumber, vehicleRegistration, personalBackground, googleAPIRefreshToken, receiveNotificationEmails, fields) values ('$username', '$passwordStrong', '$salt', 0, '$username', '', '', '$username', '$username', '$username', 'Unspecified', 'N', 'Full', 'Y', $roleid, $roleid, '', '', '', '', '', '', '', 'Other', '', '', 'Other', '', '', 'Other', '', '', 'Other', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', '', 'Y', 'Y', 'N', '', '', '', '', 'Y', '');";
        try {
            $connection2->query($sql);
        } catch (PDOException $e) {
            error_log(print_r($e->getMessage(), true));
        }
    } else {
            // existing user so update the password just in case
        $salt = getSalt();
        $passwordStrong = hash('sha256', $salt.$password);
        $sql = "UPDATE gibbonPerson set passwordStrong = '$passwordStrong', passwordStrongSalt = '$salt' WHERE username = '$username';";
        try {
            $connection2->query($sql);
        } catch (PDOException $e) {
            error_log(print_r($e->getMessage(), true));
        }
    }
```
and then it can proceed with the rest of the normal login process. If the user exists on gibbon but not on the ldap server I'm authenticating, or if the password wasn't the correct LDAP password, the system still runs through the normal login process to validate it that way (Useful for my admin account that shares the same username as an LDAP account but uses a different password stored in Gibbbon)

powertashton:
thank you travis :)

SKuipers (approved):
Looking good! A small tweak, but hopefully useful for those looking to implement the LDAP login. Thanks! 😃

PR #1124, opened 2020-10-12
Trip Planner: trips_manage.php

jimbo8098:
OOifying trips_manage.php. Wasn't sure if this should be done on main core or within https://github.com/raynichc/Trip-Planner. Whilst it's hosted on there, it's also part of @SKuipers refactoring doc so I'm a little uncertain. If it needs to go there, I'll raise a separate PR. 

- Queries run as gateways now
- Now performing some previously in-code checks in SQL instead which should provide some minor benefits when it comes to efficiency.

Would suggest paying close attention to the $eutFilter variable's effect on the output as I'm not 100% on the intention with that one. I think my suspicion is correct. I do note that there may be some parts above which can be removed, if everyone's happy with the functions here I'll clear it up and run code sniffer on it to reformat it.


SKuipers:
Hi Jim, thanks for the PR. Good question! Trip Planner was an additional module created by Ray Clark @raynichc and was maintained from his account https://github.com/raynichc. Since it's a module that has been in use at our schools here for some time, we've just chatted with Ray and transferred the repo into the GibbonEdu organization so the community can help maintain it. The new repo address is https://github.com/GibbonEdu/module-tripPlanner Can you re-submit your PR to that repo? Thanks! 😃 

Btw, some codesniffer stuff snuck its way into the PR, you could gitignore it in your local repo by adding a .git/info/exclude file.

PR #1126, opened 2020-10-13
Fix GitLab CI builds

yookoala:
Various GitLab CI fixes.

**Description**
Some long standing maintenance fix to the GitLab CI configurations:
1. Only do checksum check if environment variable `COMPOSER_CHECKSUM` is present.
2. Remove the global codeception and phpunit installation.
3. Refactoed gitlab-ci.yml to use extends.

**Motivation and Context**
1. The previous config hardcoded the SHA checksum of composer, which is unreasonable.
2. Since phpunit and codeception are both in the `require-dev` section of composer.json, there is no need for a global installation.
3. The previous config yml was clumsy and need better styles.

**How Has This Been Tested?**
Tested with my GitLab fork:
* [The repository](https://gitlab.com/yookoala/gibboneducore/)
* [The relevant pipeline result](https://gitlab.com/yookoala/gibboneducore/-/pipelines/202032202)

**Screenshots**
2020-10-13 23-43-56 的螢幕擷圖
2020-10-13 23-44-17 的螢幕擷圖

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,9 +1,9 @@
 before_script:
   # install apt packages
   - apt-get update -qy
-  - apt-get install -qy git zip unzip zlib1g-dev gettext locales
+  - apt-get install -qy gettext git libgd-dev libpng-dev locales unzip zip zlib1g-dev
   # install php extensions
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Looking good, thanks Koala for maintaining these files 👍

PR #1127, opened 2020-10-14
Fix installer config rendering

### LINKED ISSUE #1040: Mysql password cannot contain "/" (V 19.0.00)
**Describe the bug**
We get an error invalid input error in the initial install section (`/installer/install.php?step=1`) if the database password contains a "/".

**To Reproduce**
1. Create a database user with a password that contains a slash.
2. Go to the Gibbon URL and start the installer
3. Input the database config
4. On submit we get a redirect to `/install.php?step=2` with the error message that the input is invalid.

**Expected behavior**
The site should handle all character for password (probably need to escape the string).

**Screenshots**
image

image

**Your System**
 - Gibbon v19.0.00
 - Browser Firefox 76.0 (64-bit)

yookoala:
**Description**
1. Properly sanatize the config variables before printing them in config file.
2. Remove the "bad character" check for the database password (for it is no longer needed).
3. Resolve #1040.

**Motivation and Context**
The config file is rendered from [config.twig.html](https://github.com/GibbonEdu/core/blob/3f27cf3e5ee857a2d4936a3210671b2d7b860c9f/resources/templates/installer/config.twig.html), which would manually wrap the variable with single quote `'` to make them string literals. This is approach is susceptible to injection attack. So the installer has some [bad-character check inplace](https://github.com/GibbonEdu/core/blob/3f27cf3e5ee857a2d4936a3210671b2d7b860c9f/installer/install.php#L123-L128) to mitigate the issue.

While this would work in many cases, this also create some strange limitation to the installation. For example, those "bad character" are actually valid character for a database password (hence #1040).

**How Has This Been Tested?**
It has been tested both locally and in Travis CI. Unit tests are also written for function `config_file_template()` to test:
1. Basic normal configuration.
2. If any of the variables are missing from the array by mistake.
3. Bad characters in all the config values.

### DIFF
--- a/installer/install.php
+++ b/installer/install.php
@@ -123,7 +123,7 @@ if ($step >= 1) {
 // Check config values for ' " \ / chars which will cause errors in config.php
 $pattern = '/[\'"\/\\\\]/';
 if (preg_match($pattern, $databaseServer) == true || preg_match($pattern, $databaseName) == true ||
-    preg_match($pattern, $databaseUsername) == true || preg_match($pattern, $databasePassword) == true) {
+    preg_match($pattern, $databaseUsername) == true) {
     $isConfigValid = false;
 }
 
--- a/installer/install.php
+++ b/installer/install.php
@@ -291,8 +291,9 @@ if ($canInstall == false) {
         echo '</div>';
     } else {
         //Set up config.php
+        include './installerFunctions.php';
         $configData = compact('databaseServer', 'databaseUsername', 'databasePassword', 'databaseName', 'guid');
-        $config = $page->fetchFromTemplate('installer/config.twig.html', $configData);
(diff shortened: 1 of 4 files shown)


yookoala:
We can also remove all the other checks in database hostname, database name, and username. Or even better, rewritten to actually validate the strings to their purpose. The error message can be much clearer than "Your request failed because your inputs were invalid". But this seems link another PR.

SKuipers:
Hmm, it's still returning an error, just a different error this time 🤔

yookoala:
Hopefully, its all fixed this time

yookoala:
Yay

SKuipers (approved):
Thanks Koala for digging into this and creating not only a fix, but a unit test for it as well 😃 All the tests are passing, so I'm going to merge this in and check it out. Much appreciated!

PR #1132, opened 2020-10-20
Replacing session variables and adding null coalesce operators

fvlasie:
Replacing session variables and adding null coalesce operators.

I am not sure about the isset() replacements but I did my best! :)


yookoala:
A quick hint for using GitHub, you may simply [change the base](https://help.github.jp/enterprise/2.11/user/articles/changing-the-base-branch-of-a-pull-request/) of your pull request instead of re-creating the whole PR. This would be a time-saver for you. :-)

PR #1133, opened 2020-10-20
Replacing session variables and adding null coalesce operators

fvlasie:
Replacing session variables and adding null coalesce operators.

I am not sure about the isset() replacements but I did my best! :)

### DIFF
--- a/src/UI/Components/Header.php
+++ b/src/UI/Components/Header.php
@@ -46,7 +46,7 @@ class Header
 
         $return .= '<div class="flex flex-row-reverse mb-1">';
 
-        if $gibbon->session->get('username') = $_GET[$gibbon->session->get('username')] ?? '' != false) {
+        if (null !== $gibbon->session->get('username')) {
             //MESSAGE WALL!
             if (isActionAccessible($guid, $connection2, '/modules/Messenger/messageWall_view.php')) {
                 $return .= "<div id='messageWall' class='relative'>";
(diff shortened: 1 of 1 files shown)


yookoala:
Thanks for helping. There are something that I'd like to address so your contribution can be merged into the codebase:

## General Advices

1. Please separate the session refactor and null coalesce operator refactoring into 2 different PR. Each pull request should have a clear topic on its own. It would be easier for the reviewer to review the works.

2. I noticed that there are some syntax problem with the changes here. Please at least validate the file syntax with the PHP CLI or any IDE extension that does the job.

## Null Coalesce Operator

The null coalesce operator refactoring is only applicable for the *variable assignments* like this:

```php
if (isset($some_array['some key'])) {
  $variable = $some_array['some key'];
} else {
  $variable = "some default value";
}
```

Where "some default value" can be `null` or any other value for assignement.

This is usually written with [Ternary Operator](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary) in the old codebase like this:

```php
$variable = isset($some_array['some key']) ? $some_array['some key'] : "some default value";
```

If you found assignments like this, you can rewrite it with the [null coalesce operators](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op):

```php
$variable = $some_array['some key'] ?? "some default value";
```

fvlasie:
Thank you for the very helpful instructions  @yookoala !

I have a reputation of doing more harm than good when I touch Gibbon code but I will see if I can get this fixed!

:)

SKuipers (approved):
Hi FV. This particular file is an object rather than a plain php file, so it does not have direct access to the $gibbon object. Luckily, because this object is new-ish and handled by the container, this can be fixed fairly easily. I'm going to merge it in and make a few tweaks post-merge so that it works 👍  Thanks!

SKuipers:
Btw, the session class has some handy functions that can help for refactoring and improving the clarity of the code.

For example, an `isset($_SESSION[$guid]['someVar'])` call can be replaced with `$gibbon->session->exists('someVar')`.

A `!empty($_SESSION[$guid]['someVar'])` is the equivalent of `$gibbon->session->has('someVar')` (which checks to see that the value is set as well as is not empty).

The get method takes a second param as a default value, so `$gibbon->session->get('someVar') ?? ''` can be written as `$gibbon->session->get('someVar', '')`.

fvlasie:
Those are handy!

PR #1142, opened 2020-10-20
billingSchedule_manage.php: Replacing session variables

fvlasie:
There was a ```?>``` at the end of the file. I am assuming that can be deleted...


SKuipers (approved):
Yep, thanks! Removing the trailing `?>` is good practice as it helps prevent any whitespaces that may mess up the http headers.

PR #1155, opened 2020-10-21
moduleFunctions.php: Replacing session variables

fvlasie:
Working on this file brought back some memories!


SKuipers:
Hi FV. Unfortunately the $gibbon object isn't available inside these function calls, so they can't be refactored the same way as other areas of the codebase. One possibly workaround is to use $gibbon as a global, but it's hack-ish, and in in the interest of refactoring it'll be best to revisit these functions in the future and turn them into proper objects. I'm going to close this PR for now, but I absolutely appreciate your refactoring efforts 😃

PR #1159, opened 2020-10-21
Installer: fix gibbon.php include path

yookoala:
**Description**
In `gibbon.php`, used `__DIR__` to resolve the absolute path to the `error.php` instead.

**Motivation and Context**
When `gibbon.php` is run within the context of `installer.php`, the statement "include './error.php'" would resolve an incorrect path and hence unable to display the error message.

**How Has This Been Tested?**
Tested locally.


SKuipers (approved):
Thanks, well caught! 👍

PR #1160, opened 2020-10-21
Installer: fix gibbon.php include path

yookoala:
**Description**
In gibbon.php, use `__DIR__` to resolve the absolute path to the `error.php` instead of relative path.

**Motivation and Context**
When gibbon.php is included from installer.php, the statement "include './error.php'" would be resolved incorrectly and hence unable to include and display the error message.

**How Has This Been Tested?**
Locally


yookoala:
Sorry. Was merged.

PR #1161, opened 2020-10-21
Installer: extract inline-javascript as much as possible

yookoala:
**Description**

* Separate the code that expose environment data to javascript for the javascript that uses the data.
* Extract majority of `install.php` javascript into `installer.js` for easier development and debug. Leave only the variable exposing part for inline.

**Motivation and Context**
* Make the installer easier to read and maintain.
* A step to gradually improve the installer.

**How Has This Been Tested?**
Tested locally as well as with Travis.

### DIFF
--- a/installer/assets/installer.js
+++ b/installer/assets/installer.js
@@ -1,0 +1,54 @@
+(function () {
+    $(document).ready(function() {
+
+        // If an element with id "status" is found, do the version check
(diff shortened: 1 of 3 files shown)


SKuipers (approved):
Looks good, I'm going to merge and test. Thanks!

PR #1163, opened 2020-10-22
Exchanged chars in System Overview template

leofreitas:
**Description**
Exchanged chars in System Overview template

**How Has This Been Tested?**
Local


SKuipers (approved):
I like "exchanged chars", it's a very graceful way to say "fixed typos" 😂

PR #1167, opened 2020-11-05
Planner: final form ooification

SKuipers:
This is the last piece of form ooification in the core 🎉  (ignoring a couple small forms in the timetable, which will be templatified).

This PR makes a couple visual tweaks to the lesson content section, so that the padding is not as squished. It ooifies the attendance sidebar, guest list, and homework submission forms. It also removed the previous front-end editing of smart blocks in favour of adding a link to the smart block section of the Edit Planner form.

**How Has This Been Tested?**
Locally

**Screenshots**
Attendance:
Screenshot 2020-11-05 at 9 26 52 AM

### DIFF
--- a/modules/Planner/planner_view_full_myHomeworkProcess.php
+++ b/modules/Planner/planner_view_full_myHomeworkProcess.php
@@ -63,8 +63,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/planner_view_full.
         } else {
             //Get variables
             $homework = $_POST['homework'];
-            if ($_POST['homework'] == 'Yes') {
-                $homework = 'Y';
+            if ($homework == 'Y') {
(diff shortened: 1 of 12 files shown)


modules/Planner/planner_view_full_myHomeworkProcess.php
```diff
@@ -63,8 +63,7 @@
         } else {
             //Get variables
             $homework = $_POST['homework'];
-            if ($_POST['homework'] == 'Yes') {
-                $homework = 'Y';
+            if ($homework == 'Y') {
```
rossdotparker, line 66: The conditional that has been removed uses "Yes" but the new one uses "Y". I guess there is a corresponding change in the front end?

rossdotparker:
This is a massive milestone. Well done : )

PR #1168, opened 2020-11-06
Student birthday option

rossdotparker:
**Description**
Given the right permissions, students can see birthday icon alerts for classmates in the class and form group table views, but not in the Planner lesson view. This PR adds the birthday icon to the Planner lesson view using the same logic as the other two locations.

**Motivation and Context**
To resolve this inconsistency in order to deliver the same behaviour across the board.

**How Has This Been Tested?**
Locally


SKuipers (approved):
Looking good, going to merge & test 👍

PR #1172, opened 2020-11-13
System Admin: add Server Info and Cache Manager pages

SKuipers:
General improvements to System Admin to help people troubleshoot and debug their system.

- Adds a dedicated page for managing cache. 
- Scan the cache directory for non-writeable files, which will cause errors in the system.
- Enable customizing the location of the cache folder.
- Adds a Server Info page which displays the contents of phpinfo and makes it easier to find the error log location.
- Adds a check for exec() access to the System Check page.

**How Has This Been Tested?**
Locally.

**Server Info**
Screenshot 2020-11-13 at 3 54 16 PM

**Cache Manager**
Screenshot 2020-11-13 at 3 54 30 PM


rossdotparker:
These are some more excellent back-end features, which will definitely help with troubleshooting and development : )

PR #1174, opened 2020-11-18
System Admin: add a Purge Logs action to the View Logs page

SKuipers:
Adds the ability to delete logs by type and before a specified cutoff date.

**Motivation and Context**
The logs can fill up in large schools.

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2020-11-18 at 10 53 05 AM

### DIFF
--- a/modules/System Admin/logs_view_purge.php
+++ b/modules/System Admin/logs_view_purge.php
@@ -1,0 +1,56 @@
+<?php
+/*
+Gibbon, Flexible & Open School System
+Copyright (C) 2010, Ross Parker
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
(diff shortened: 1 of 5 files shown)


modules/System Admin/logs_view_purge.php
```diff
@@ -0,0 +1,56 @@
...
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use Gibbon\Forms\Form;
+use Gibbon\Services\Format;
+
+require_once __DIR__ . '/moduleFunctions.php';
+
+if (isActionAccessible($guid, $connection2, '/modules/System Admin/logs_view.php') == false) {
```
rossdotparker, line 25: In theory this ought to be /logs_view_purge.php, with a related CHANGEDB.php entry. I am not sure it ultimately makes a difference, unless we want to do some kind of audit, or introduce a grouped action.

PR #1176, opened 2020-11-20
System: refactor the update process and add a simple migrations system

SKuipers:
This PR ooifies the Update page and extracts update-related logic into an Updater class. It also adds a system where complex migrations requiring PHP can be run by the updater.

In this PR:
- Some visual improvements to the update page, see below.
- The Inno DB update has been moved into a migration class, and will only display if an update is required.
- Both the System Admin/update.php and the root update.php files now use a common Updater class.
- The updater now includes a migration component, which can be used for complex database changes that cannot be run in a CHANGEDB.php line.
- Migration scripts extend an abstract Migration class. They include the date in the class name, similar to other migration systems, to ensure that they're run in order.
- Migrations that have been run are stored in a new gibbonMigration table, along with the current version and timestamp.

It makes some key changes to the updater logic:
- In a regular update, full version updates that fail will stop after that version and not carry on updating the next version. This helps prevent compound issues where updates fail one after another.
- In a cutting edge update, line-by-line changes will stop after the first error. This enables manually correcting the error, then running the update to continue where it left off. It helps prevent the cuttingEdgeCodeLine number from getting out of sync with the SQL that has actually been run.

**Motivation and Context**
Currently the update process has existed in two locations, copy-pasted. Before adding additional functionality it made sense to refactor it.

**How Has This Been Tested?**
Locally with a regular update, cutting edge full update and cutting edge partial update.

**Screenshots**
Screenshot 2020-11-20 at 2 31 13 PM


rossdotparker:
A substantial change, but great see the removal of some of that repeated code, and a more abstracted and structured approach to updates.

PR #1179, opened 2020-11-25
System: migrate all custom fields from serialized data into json data

SKuipers:
This PR is a follow-up to #1176 which adds a migration for changing serialized custom fields into JSON custom fields. JSON in a database is more stable and has better functionality in more recent versions of MySQL. 

- Adds a migration file in src/Database/Migrations.
- Changes instances of unserialize/serialize to json_decode/json_encode.
- Allows the selectBy to be used with no constraints, returning all rows.

**How Has This Been Tested?**
Locally


SKuipers:
Note: Migrations are triggered by an update, so the CHANGEDB here makes a fairly cosmetic change to the database in order to trigger the migration. This shouldn't be an issue in the future, as the need to run complex PHP migrations is nearly always preceded by a change to the database structure in some way.

rossdotparker:
Looking great...another leap in the right direction! I'll merge and then test locally.

PR #1184, opened 2020-12-10
System Admin: add a Data Retention tool for scrubbing personal data

SKuipers:
This is a big one 😅 Ross laid the ground work and I finalized the functionality.

Functionality wise:
- Adds a Data Retention page to System Admin. This process clears certain fields, rather than removing any database rows.
- Data retention requires a cutoff date, which is used to check which user data is deleted. The action will scrub selected data for all users in the specified category whose status is Left, and whose end date precedes the specified data.
- Data retention cutoff must be before the current school year, otherwise unpleasant things would happen!
- Data retention is divided into a number of categories, to help admins choose which data to clear.
- Handles scrubbing files by removing not only the file path but also the file itself.
- Displays a note on Edit User if the user's data has been scrubbed.

Code wise:
- Adds a DataRetention gateway with a list of scrubbable gateways.
- Adds a gibbonDataRetention table to store a list of all users which have been scrubbed.
- Adds a ScrubbableGateway interface which indicates that a gateway can be scrubbed.
- Adds a Scrubbable trait which implements the scrubbing part of the interface.
- Adds ScrubByPerson, ScrubByTimestamp, and ScrubByFamily traits to get the records that are scrubbable.
- Updates each of the gateways that are scrubbable to indicate which fields need to be scrubbed.

**Motivation and Context**
Gibbon’s core currently retains user data indefinitely. Whilst school’s will generally not be interested in expunging academic records, compliance with privacy laws and best-practice require the ability to remove certain data after a period of time.

**How Has This Been Tested?**
Locally with a few different data sets.

**Screenshots**
Data Retention page:
Screenshot 2020-12-10 at 12 30 19 PM

Example of deleted file:
Screenshot 2020-12-10 at 12 29 52 PM

Edit User:
Screenshot 2020-12-10 at 1 15 54 PM


modules/System Admin/dataRetentionProcess.php
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    // Store a record of which users and which tables were scrubbed
+    foreach ($scrubbedList as $gibbonPersonID => $tables) {
+        if (empty($gibbonPersonID)) continue;
+
+        $data = [
+            'gibbonPersonID'            => $gibbonPersonID,
+            'tables'                    => json_encode(array_values(array_unique($tables))),
+            'status'                    => !empty($tables) ? 'Success' : 'Partial Failure',
```
rossdotparker, line 92: Not sure about the logic here. Can you explain?

src/Domain/Students/MedicalConditionGateway.php
```diff
@@ -19,36 +19,28 @@
...
+    private static $tableName = 'gibbonPersonMedicalCondition';
+    private static $primaryKey = 'gibbonPersonMedicalConditionID';
 
     private static $searchableColumns = [];
 
-    /**
-     * @param QueryCriteria $criteria
-     * @return DataSet
-     */
-    public function queryMedicalConditions(QueryCriteria $criteria)
```
rossdotparker, line 43: I'm interested to know why this one was removed. Is it inherited? in some form?

modules/System Admin/dataRetentionProcess.php
SKuipers: Sure, this sets up the data to insert into the gibbonDataRetention table, and if there's already a record, it merges the list of tables that have been cleared. I figured we didn't need a timestamp for every time the data has been cleared, just the most recent time, and a record of everything cleared up until that point.

src/Domain/Students/MedicalConditionGateway.php
SKuipers: It wasn't necessarily removed, that's an artifact of the version comparison. I realized our naming conventions for the new MedicalConditionGateway overlapped with the need for a MedicalConditionGateway for individual users. So, I moved the gateway that manages the list of potential conditions to School Admin, where it's managed, so I could add the gateway for students to the Student namespace. The other option was finding and renaming all the existing MedicalGateway uses to PersonMedicalGateway. The first option seemed better, since that gateway was just introduced this version.

src/Domain/Traits/ScrubByFamily.php
```diff
@@ -0,0 +1,68 @@
...
+        // Apply the context based on user role
+        if (!empty($context)) {
+            $query->innerJoin('gibbonRole', 'gibbonRole.gibbonRoleID=gibbonPerson.gibbonRoleIDPrimary')
+                ->where('FIND_IN_SET(gibbonRole.category, :roleCategories)')
+                ->bindValue('roleCategories', implode(',', $context));
+        }
+
+        // Only get users whose dateEnd is before the cutoff, falling back to using the lastTimestamp
+        $query->where('((gibbonPerson.dateEnd IS NOT NULL AND gibbonPerson.dateEnd < :cutoffDate) 
+            OR (gibbonPerson.lastTimestamp IS NOT NULL AND gibbonPerson.lastTimestamp < :cutoffDate)
```
rossdotparker, line 56: Should this line also contain a gibbonPerson.dateEnd IS NOT NULL, to ensure it only looks at last login if the end date is not set? I guess in theory there should not be a last login date after the end date, unless the end date was retroactively entered.

SKuipers: Sure, I'll add that to be on the safe side.

modules/System Admin/dataRetentionProcess.php
rossdotparker: Sorry, should have been more specific, looking at line 92. It seems to set Success if there are any tables, otherwise Partial Failure. What happens if some fail and some succeed, would $tables not be empty?

PR #1189, opened 2021-01-08
Refactoring - Attendance

hmerrettICHK:
**Description**
Refactoring process files in Attendance module - adding null coalesce

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


modules/Attendance/attendance_take_byCourseClassProcess.php
```diff
@@ -26,14 +26,14 @@
...
 require_once __DIR__ . '/moduleFunctions.php' ;
 
-$gibbonCourseClassID=$_POST["gibbonCourseClassID"] ;
-$currentDate=$_POST["currentDate"] ;
+$gibbonCourseClassID=$_POST["gibbonCourseClassID"] ?? '';
+$currentDate=$_POST["currentDate"] ?? '';
 $today=date("Y-m-d");
 
-$moduleName = getModuleName($_POST["address"]);
+$moduleName = getModuleName($_POST["address"]) ?? '';
```
SKuipers, line 33: Hi Harry, this is one spot where the null coalesce would need to be added to the variable itself rather than the function call, see my example below:

```suggestion
$moduleName = getModuleName($_POST["address"] ?? '');
```

modules/Attendance/attendance_take_byPerson_editProcess.php
```diff
@@ -23,9 +23,9 @@
 //Module includes
 include './moduleFunctions.php';
 
-$gibbonAttendanceLogPersonID = isset($_POST['gibbonAttendanceLogPersonID'])? $_POST['gibbonAttendanceLogPersonID'] : '';
-$gibbonPersonID = isset($_POST['gibbonPersonID'])? $_POST['gibbonPersonID'] : '';
-$currentDate = isset($_POST['currentDate'])? $_POST['currentDate'] : dateConvertBack($guid, date('Y-m-d'));
+$gibbonAttendanceLogPersonID = $_POST['gibbonAttendanceLogPersonID'] ?? '';
+$gibbonPersonID = $_POST['gibbonPersonID'] ?? '';
+$currentDate = $_POST['currentDate'] ?? dateConvertBack($guid, date('Y-m-d'));
```
SKuipers, line 28: Another spot. See if you can examine the syntax and come up with a way to write this as a null coalesce.

SKuipers (changes requested):
I've added some items to look at.

modules/Attendance/attendance_take_byPerson_editProcess.php
hmerrettICHK: Ross and I discussed this with you, and we decided that it's okay.

SKuipers (approved):
👍

PR #1195, opened 2021-01-16
Hash deleted in code, because the alertColor attribute has hash in the value

leofreitas:
**Motivation and Context **
The pages doesn't show the color correctly

**Description**
- Hash deleted from the code, because the alertColor attribute has the value with hash
- alertColor attribute fixed in template medicalConditions

**How Has This Been Tested?**
 Local and Travis


SKuipers (approved):
Thank you Leo! Well caught on these fixes, I'm merging them in now 👍

PR #1198, opened 2021-01-18
Fixed null reference when row modifier returned null

raynichc:
**Description**
Fix for a bug that occurs when two (or more) row modifiers are added to a table. If any (except the last) row modifier returns null, an error will be thrown as all subsequent modifiers will be given the null variable.

**Motivation and Context**
Expandable columns use rowModifiers, however, if a custom row modifier that can (and does) return null is used, an error will be thrown and the table will not render.

**How Has This Been Tested?**
Locally on the Trip Planner (where the bug existed) and on the Help Desk (where an existing table with a row modifier exists)

**Screenshots**
image


SKuipers (approved):
Wow, well caught! Thanks for the fix Ray, merging & testing 👍

PR #1201, opened 2021-01-19
Fixed Custom Block Labels being overriden

raynichc:
**Description**
Fixes a bug that causes labels in custom blocks to be overriden when blocks are preloaded.

**How Has This Been Tested?**
Locally on the Trip Planner and Planner modules


SKuipers (approved):
Thanks Ray, well caught!

PR #1202, opened 2021-01-19
Password force reset fixed

leofreitas:
**Description**
 - moved item returns['successa'] from preferences.php to functions.php because when the request is successfully completed, the user is redirected to index.php
- in preferencesPasswordProcess.php, if passwordForceReset is true and a PDO exception occurs, the URL should be preferences.php and not index.php

**Motivation and Context**
- password force reset fixed

**How Has This Been Tested?**
Local and Travis


SKuipers (approved):
Thanks! Well caught, I'm going to merge and test now 👍

PR #1203, opened 2021-01-19
addUser: translation support

leofreitas:
**Description**
Translation support for the Primary Role field

**Motivation and Context**
Translation support 

**How Has This Been Tested?**
Local and Travis


SKuipers (approved):
Thanks! One of the final fixes for v21 😄

PR #1207, opened 2021-01-21
resources/build: remove obsoleted files and folders

yookoala:
**Description**
Remove the leftover in the build folder.

**Motivation and Context**
The build folder previously migrated as a submodule. The leftover files and folders are obsoleted and should be removed.

**How Has This Been Tested?**
Tested locally:
* When cloning the branch recursively, both i18n and resource/build folder are fetched as expected.
* With this change, the shallow clone size went down from 906MB to 703MB.


SKuipers (approved):
Thanks! Somehow the merge into main messed up the submodule and the whole node_modules folder tagged along 😬 This fix looks good, going to merge and test now.

PR #1123, opened 2020-10-12
Replace "/vendor" with automatic installation bundle in release (Travis CI)

yookoala:
Instead of keeping a `/vendor` folder in the repository, we tell Travis CI to build and submit additional bundle for easy installation.

**Description**

1. Add a GitHub asset release script for uploading arbitrary files and attach to GitHub release.
2. Add a "[deploy](https://docs.travis-ci.com/user/deployment/)" section in the .travis.yml to have Travis run `composer install`, build an archive with everything and submit to GitHub as part of a tagged release (see "[Script Deployment](https://docs.travis-ci.com/user/deployment/script/)" for more details).
3. Remove "/vendor" folder from the repository (this change messed with the diff here, but you can checkout my branch and diff the individual commits there).

**Motivation and Context**

The current GibbonEduCore includes the `/vendor` in the repository so that the release "Source code" archives would have all the files and folders needed for installation. This is for the ease of installation to school system admin who are not familiar with composer. However, this is considered an anti-pattern in modern PHP source code management (as discussed before in threads like #462).

To remove the "/vendor" folder while still supporting system admin to install / upgrade their Gibbon, we need a way to release archive files that has the proper "/vendor" folder after it is removed from the source.

Since Gibbon is using Travis CI for continuous testing, the logical approach is to use Travis to build such archive and release it.

This PR add a generic GitHub asset releasing script "cli/github-release.sh" for releasing the "Install Bundle" (the archive file(s) for easy installation). Then add some instructions to Travis so it would build and release the archives at appropriate triggers (i.e. Tag, Environment variables, branch, etc).

**How Has This Been Tested?**

The approach is tested in [my fork of GibbonEduCore](https://github.com/yookoala/GibbonEduCore/) for some times. It works:

* [Travis CI build record](https://travis-ci.org/github/yookoala/GibbonEduCore/jobs/734781546)
* [Result (buildtest3)](https://github.com/yookoala/GibbonEduCore/releases/tag/v21.0.00-buildtest3)

**Screenshots**
2020-10-12 11-55-09 的螢幕擷圖

PR #1209, opened 2021-01-21
Replace "/vendor" with automatic installation bundle in release (GitHub Actions)

yookoala:
Attach an additional release bundle to all releases.

**Description**
Another take of #1123. But this time, uses GitHub Actions instead of Travis CI.

The approach is a bit different than before. The GitHub Action described in "release.yml" is triggered when a Release is created (not when a tag is pushed). The installation bundles will be attached to the newly created release. It won't create the release for you, so it give more flexibility for releasing.

**Motivation and Context**
Same as #1123.

**How Has This Been Tested?**
On my fork of GibbonEdu.


yookoala:
## Way to Test This
1. Create a fork (or use existing fork) of GibbonEdu/core. Tag any commit in it. Remember to push the tag to your fork repository.
2. Create a Draft Release. Select the newly created tag. And fill in the Title / Description as you please.
    2021-02-04 17-27-13 的螢幕擷圖
    2021-02-04 17-27-49 的螢幕擷圖
    2021-02-04 17-32-36 的螢幕擷圖
3. Publish the Release. The newly created release will trigger GitHub Action to run:
    2021-02-04 17-58-57 的螢幕擷圖
4. After the release action is done, the bundle zip and tar.gz file will be added to the newly created release:
    2021-02-04 18-01-35 的螢幕擷圖
    2021-02-04 18-01-55 的螢幕擷圖

SKuipers (approved):
Thanks for your eternal patience on this one! I think we're finally ready to make this switch. We'll still need to inform our developers so they're aware of how to install and manage composer, but this is a worthwhile step to finally be able to take advantage of the whole point of a package manager 😄  Having #1267 in place should help with that too.

Amazing to see 1.3 million lines of code disappearing 🎉 Thanks!!!

PR #1211, opened 2021-01-22
Feature: connector debug in installer

yookoala:
**Description**

* Add a backward-compatible change to connector to allow it to rethrow PDO connection error.
* Improves the installer error message with database connection issue(s).
* Probably fixed an installer bug. Instead of checking `empty($pdo)`, which is `false` instead of `null` when error, the installer should either check something like `$pdo instanceof \PDO` or `$pdo !== false`.

**Motivation and Context**

It's really hard to tell what went wrong if the installer failed to connect to database. Especially if you're debugging CI build errors. It would be really helpful to show some internal error message on the page. Then the test artifact would be more helpful to debug with.

**How Has This Been Tested?**

Tested locally.

### DIFF
--- a/installer/install.php
+++ b/installer/install.php
@@ -20,6 +20,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 use Gibbon\View\Page;
 use Gibbon\Forms\Form;
 use Gibbon\Data\Validator;
+use Gibbon\Database\Connection;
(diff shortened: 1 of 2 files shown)


yookoala:
@SKuipers: Sorry. Some VSCode auto formatting fix slipped in...

SKuipers (approved):
Looks like a useful improvement for debugging installation issues, thanks Koala!

Issue #1210, opened 2021-01-21
Make Gibbon PHP 8 ready


yookoala:
**Is your development discussion related to a problem?**

PHP 8.0 is out. It's time to make Gibbon forward-compatible with PHP 8.x.

**Describe the solution you're proposing**
1. Add PHP 8 as a test target.
2. Handle whatever error we encounter.
3. Hopefully we can get both PHP 7.x and PHP 8.x happy.

**Describe alternatives you've considered**
I don't see any.

**Additional context**
Already encountered some compatibility issues in `Gibbon\Contract\Database\Connection` and `Gibbon\Contract\Database\Result` that would break the installer. Haven't started looking into the modules yet. Hopefully not much needs to be changed.

SKuipers:
Hi Koala, thanks for getting the ball rolling. Yep, PHP 8 compatibility is a good goal for v22. Lets keep this discussion open and use it to track our progress in this area 👍

SKuipers:
I've pushed an initial fix for the database issues, which means that pages will now load and the installer will run, albeit with some deprecation messages related to Twig. https://github.com/GibbonEdu/core/commit/805b0af191a3abe4b0ae279f5299f99b12146951 

I'll work on a few more compatibility issues today, but I suspect the majority of them will be uncovered through testing and general use.

abdirooble:
<----I'm posting this for the second time, hoping to get faster answer ;).----/>

hey guys!
I have two questions.
can someone tell me which stacks Gibbon has?
and do you use a framework for PHP (laravel) or only PHP?

yookoala:
> <----I'm posting this for the second time, hoping to get faster answer ;).----/>
> 
> hey guys!
> I have two questions.
> can someone tell me which stacks Gibbon has?
> and do you use a framework for PHP (laravel) or only PHP?

OT. Please see my reply in your other comment.

rossdotparker:
Hi abdirooble, whilst I definitely appreciate your enthusiasm, multiple posts and emails acts to slow us all down. We work across multiple time zones, offering volunteer support to help as many people as we can. Resources are limited. If you make a single post ([to the forums](https://ask.gibbonedu.org/) as Koala pointed out in his other reply), we'll get back to you there as soon as we can. Cheers : ) Ross

SKuipers:
[...]

PR #1213, opened 2021-01-23
github-actions: add experimental php8 test

yookoala:
**Description**
* Add PHP 8.0 test to CI.
* Mark it as "experimental" and do not treat failure as an error.
* Failure will still have message in the CI result. And test artifacts will still be available for study.

**Motivation and Context**
As as step to gradually resolve issue #1210 without disturbing other PR and updates.

**How Has This Been Tested?**
On this PR, with GitHub Actions.

**Screenshot**
2021-01-23 16-05-02 的螢幕擷圖

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,15 +6,17 @@ on:
 
 jobs:
   test:
-    name: CI Test (PHP ${{ matrix.php-version }})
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Interesting! Its handy that we can enable this without immediately causing all our tests to error. Thanks for setting this up 👍

PR #1215, opened 2021-01-23
Activities: fix count error with null in php 8

yookoala:
* The variable $staff (from $_POST['staff']) should be an array
  from the form input. The previous fallback was to null, which
  is not Countable nor array. In PHP 8, using count() on anything
  other than the 2 is a TypeError.

* Instead, changed the fallback to empty array would fix the issue.

* In addition, add extra sanitation step to make sure it is an
  array.

**Description**

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


SKuipers (approved):
Looks good.

PR #1216, opened 2021-01-25
Refactoring Job Openings - moving code to Twig template

hmerrettICHK:
**Description**
Refactoring Job Openings - moving code to Twig template

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


modules/Staff/applicationForm_jobOpenings_view.php
```diff
@@ -67,10 +67,10 @@
         echo "<a href='".$_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/'.$_SESSION[$guid]['module']."/applicationForm.php'>".__('Submit Application Form')."<img style='margin-left: 5px' title='".__('Submit Application Form')."' src='./themes/".$_SESSION[$guid]['gibbonThemeName']."/img/plus.png'/></a>";
         echo '</div>';
```
yookoala, line 68: Please also move these to the template.

yookoala:
Please also rewrite the error messages with `$page->addError`. You can see these lines for reference:
https://github.com/GibbonEdu/core/blob/cbaa8a8bca8616f7dca8f565ed18d152c1dbd6ca/index.php#L147-L154

yookoala (changes requested):
Please also rewrite the error messages with `$page->addError`. You can see these lines for reference:
https://github.com/GibbonEdu/core/blob/cbaa8a8bca8616f7dca8f565ed18d152c1dbd6ca/index.php#L147-L154

SKuipers (changes requested):
Hi Harry, the  resources/templates/ui/activityView.twig.html file accidentally ended up in this PR. Can you double-click to check out your jobOpenings branch locally, then remove this file and commit the change, then push your change. That will remove it from this PR. Thanks!

SKuipers (approved):
Thanks Harry, I'm going to merge this one and test it 👍  There are eventually more types of refactorings you can do, but we'll do one at a time for now. All in good time!

modules/Staff/applicationForm_jobOpenings_view.php
SKuipers: Thanks Koala. Interestingly, this is one of the few areas where an Action link occurs without a form or table. Ideally I'd like them to use the same object, so we'll keep this one out of the template for now and look at a more object-oriented solution.

PR #1218, opened 2021-01-26
Tests: revamp grouping

yookoala:
**Description**
* Fix previously obsoleted grouping config.
* Regroup the tests by functionality.

**Motivation and Context**
* According to documentations, the [groups](https://codeception.com/docs/07-AdvancedUsage#group-files
) directive in configuration serves to specify files and folders so user can run tests selectively. But the directive has to specify files and folders that exists.
* The current configuration seems to be generated by `codecept init`. All previous iterations of this file don't seems to match with any specific subfolder in the `tests/` folder:

  https://github.com/GibbonEdu/core/blob/7272bd60218ac59dc5e259556f923ee8d2dc6eb4/tests/acceptance.suite.yml#L17-L18
  https://github.com/GibbonEdu/core/blob/adf8f0f5b5556219024413c790db0eb3d0c785ca/tests/acceptance.suite.yml#L33-L34

* In codecption 3, the directive is not checked so there is no issue with it. In codectpion 4, however, this will cause ConfigurationException:
  2021-01-26 13-38-36 的螢幕擷圖

**How Has This Been Tested?**
Tested locally and in CI environment.

### DIFF
--- a/tests/acceptance.suite.yml
+++ b/tests/acceptance.suite.yml
@@ -31,4 +31,17 @@ modules:
             cleanup: false
             reconnect: true
 groups:
-    users: [/Users]
+    users:
+      - acceptance/User Tests
+      - acceptance/User Admin
+    system:
+      - acceptance/System Admin
+    modules:
+      - acceptance/Activities
+      - acceptance/Data Updater
+      - acceptance/Finance
+      - acceptance/Markbook
+      - acceptance/Planner
+      - acceptance/School Admin
+      - acceptance/Students
+      - acceptance/Timetable Admin


SKuipers (approved):
Nice, looks good. Does this configuration have a wildcard option? The one thing with adding the names of the folders manually is that we'll have to remember to update them when we add new tests.

yookoala:
> Nice, looks good. Does this configuration have a wildcard option? The one thing with adding the names of the folders manually is that we'll have to remember to update them when we add new tests.

Yes, you can use wildcard here. Also grouping is optional. You can remove the `groups` directive entirely and the tests would still work.

Issue #1214, opened 2021-01-23
PHPUnit / Codeception errors in PHP 8


yookoala:
**Describe the bug**

Ran into this error in PHP 8 testing:
```
ParseError: syntax error, unexpected token "match", expecting variable
```
The error is very similar to doctrine/dbal#4197, which is caused by sebastianbergmann/phpunit#4373. In short, an internal programming interface is using the keyword "match". Since PHP 8 introduced the [match expression](https://www.php.net/manual/en/control-structures.match.php), it has became a [reserved word](https://www.php.net/manual/en/reserved.keywords.php). The solution for PHPUnit is to replace it with something else. And the fix was published in PHPUnit 9.3.

As shown in PHPUnit's [compatibility table](https://phpunit.de/supported-versions.html):
![2021-01-23 16-30-14 的螢幕擷圖](https://user-images.githubusercontent.com/91274/105573405-4fca4400-5d98-11eb-9190-16863b48c2f1.png)

PHPUnit 9 supports PHP 7.3+, which seems to be the targeted lowest supported PHP version now (?). And if I read it correctly, Codeception/Codeception#6000 is merged into [Codeception 4.1.9](https://codeception.com/changelog) so we can expect it to be safe for PHP 8.

**To Reproduce**
1. Run the acceptance test suite in PHP 8.0.x environment.
2. See the error(s).

**Expected behavior**
All tests should be able to run in PHP 8 environment.

**Screenshots**
![2021-01-23 16-22-17 的螢幕擷圖](https://user-images.githubusercontent.com/91274/105573283-602def00-5d97-11eb-9b2d-4e55e8c568d0.png)

![2021-01-23 16-22-04 的螢幕擷圖](https://user-images.githubusercontent.com/91274/105573285-61f7b280-5d97-11eb-8537-d73ed1b4e547.png)

**Your System**
 - PHP 8

**Additional context**
Blocks #1210.

yookoala:
Changing the codeception and phpunit version seems to require changes to a lot of libraries in vendor folder. It would be better if #1123 is merged before touching this subject.

SKuipers:
Great work! Yes, with our minimum version moving up to PHP 7.3, it'll be great to update our test suite to make use of newer PHPUnit and Codeception versions. I've merged your PRs and everything is looking good, all tests are green 👍 A massive thank you for doing the leg-work on these upgrades.

PR #1221, opened 2021-01-26
Tests: fix TestCase for phpunit 9

yookoala:
**Description**
* Fix forward all unit test compatibiliy with phpunit 9 by updating setUp and tearDown function signatures.

**Motivation and Context**
* Newer PHPUnit TestCase setUp and tearDown method needs to specifically hint to return "void".
* The change is backward compatible. Loose parent methods are compatible with strict children implementations.
* This fixed a few errors that block #1214.

**How Has This Been Tested?**
On CI platform.

### DIFF
--- a/tests/unit/Data/UsernameGeneratorTest.php
+++ b/tests/unit/Data/UsernameGeneratorTest.php
@@ -29,7 +29,7 @@ class UsernameGeneratorTest extends TestCase
 {
     private $usernameGenerator;
 
-    public function setUp()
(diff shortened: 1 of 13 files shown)


SKuipers (approved):
Thanks! You're really streamlining the test process. I know we were frozen on the same PHPUnit/Codeception versions for a while to try and keep things from breaking, great to see you updating these and untangling the issues. Hugely appreciated 😃

PR #1222, opened 2021-01-26
Tests: fix test Domain/QueryCriteriaTest

yookoala:
**Description**
* assertContains should be only used on iterables (e.g. array) in newer phpunit. To check if a string contains another string, use assertStringContainsString or assertStringNotContainsString instead (if available).

**Motivation and Context**
* Make the test forward compatible with phpunit 9
* This fixed an error that block #1214.

**How Has This Been Tested?**
CI platform.


SKuipers (approved):
Aha, good to know for future tests too, thanks!

PR #1229, opened 2021-01-27
Documentation: better docblock for translation functions

yookoala:
**Description**
* Improve docblock comment for __() and Gibbon\Locale::translate().

**Motivation and Context**
* It's not obvious how to use the `Locale::translate` or the `__()` function. Improve the docblock can improve the translation strings written in long term (i.e. the usage of string placeholder).

**How Has This Been Tested?*
Documentation don't seems to have a way to test.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -59,9 +59,11 @@ function emailBodyConvert($body)
 /**
  * Custom translation function to allow custom string replacement
  *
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
Thanks! I agree, it's good to clarify the usage of these functions/methods 👍

PR #1230, opened 2021-01-27
Fix Gibbon\Forms\DatabaseFormFactory database methods

yookoala:
**Description**
* Fix a mistake in reusing variable.

**Motivation and Context**
* In createSelectStudent
  * When the query is both byRoll and byName and is neither allStudents
    nor activeStudents, the $data defined in the first query will affect
    that of the second query.

    And in some MySQL / MariaDB / PDO configuration or version, this
    will cause error: "SQLSTATE[HY093]: Invalid parameter number: number
    of bound variables does not match number of tokens".

  * The solution is to re-initialize $data (and $value) before the
    second condition.

* In createSelectUsers
  * A query without placeholder is supplied with a non-empty data
    array, causing "SQLSTATE[HY093]: Invalid parameter number: number
    of bound variables does not match number of tokens". Simply change
    to query with empty array would fix it.

* In modules/User Admin/user_manage_editProcess.php
  *  A very similar issue of SQLSTATE[HY093].

**How Has This Been Tested?**
On GitLab CI.

**Screenshots**
Relevant error messages in CI:

2021-01-28 02-55-42 的螢幕擷圖
2021-01-28 02-55-54 的螢幕擷圖
2021-01-28 02-56-10 的螢幕擷圖
2021-01-28 02-56-21 的螢幕擷圖

### DIFF
--- a/src/Forms/DatabaseFormFactory.php
+++ b/src/Forms/DatabaseFormFactory.php
@@ -489,6 +489,10 @@ class DatabaseFormFactory extends FormFactory
             }
         }
 
+        //Clearn all values
+        $values = [];
+        $data = [];
+
         //Add students by name
         if ($params["byName"]) {
             if ($params["allStudents"]) {
--- a/src/Forms/DatabaseFormFactory.php
+++ b/src/Forms/DatabaseFormFactory.php
@@ -375,7 +375,7 @@ class DatabaseFormFactory extends FormFactory
                     JOIN gibbonStaff ON (gibbonPerson.gibbonPersonID=gibbonStaff.gibbonPersonID)
                     WHERE gibbonPerson.status='Full'
                     ORDER BY gibbonPerson.surname, gibbonPerson.preferredName";
-            $result = $this->pdo->executeQuery([], $sql);
+            $result = $this->pdo->select($sql);
             if ($result->rowCount() > 0) {
                 $users[__('Staff')] = array_reduce($result->fetchAll(), function ($group, $item) {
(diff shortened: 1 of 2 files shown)


src/Forms/DatabaseFormFactory.php
```diff
@@ -370,12 +370,12 @@ public function createSelectUsers($name, $gibbonSchoolYearID = false, $params =
...
         if ($params['includeStaff'] == true) {
             $data = array('gibbonSchoolYearID' => $gibbonSchoolYearID, 'date' => date('Y-m-d'));
-            $sql = "SELECT gibbonPerson.gibbonPersonID, preferredName, surname, username 
+            $sql = "SELECT gibbonPerson.gibbonPersonID, preferredName, surname, username
                     FROM gibbonPerson
                     JOIN gibbonStaff ON (gibbonPerson.gibbonPersonID=gibbonStaff.gibbonPersonID)
                     WHERE gibbonPerson.status='Full'
                     ORDER BY gibbonPerson.surname, gibbonPerson.preferredName";
-            $result = $this->pdo->executeQuery($data, $sql);
+            $result = $this->pdo->executeQuery([], $sql);
```
SKuipers, line 378:
```suggestion
            $result = $this->pdo->select($sql);
```
Small tweak. `executeQuery` is the old method. We can use the newer `select`, which has data as a second optional param

SKuipers (approved):
Looks good, thanks for catching these! 👍

PR #1231, opened 2021-01-28
Twig: fix template forward compatibility

yookoala:
**Description**
* Twig 2.x used to support using condition as a part of the for
  statement. But 3.x no longer support this syntax. Using it will
  result in "syntax error" exception from Twig.

  Reference:
  https://twig.symfony.com/doc/2.x/tags/for.html#adding-a-condition
  https://twig.symfony.com/doc/3.x/tags/for.html

* To future-proof the templates, simply extract the if statement
  from the for loop and wrap it on outside.

**Motivation and Context**
* Stumbled upon this error when trying to upgrade Twig from 2 and 3.
* We might not upgrade to Twig 3, we can still future proof our templates.

**How Has This Been Tested?**
Locally and In CI testing platform.

### DIFF
--- a/resources/templates/menu.twig.html
+++ b/resources/templates/menu.twig.html
@@ -13,19 +13,21 @@ Main Menu: Displays the top-level categories and active modules.
         <a class="block uppercase font-bold text-sm text-gray-800 hover:text-purple-500 no-underline px-2 py-3" href="{{ absoluteURL }}/index.php">{{ __('Home') }}</a>
     </li>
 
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
Aha, good to know! Thanks for updating these. It'd be nice to update to Twig 3. Are there any other breaking changes to consider?

yookoala:
> Aha, good to know! Thanks for updating these. It'd be nice to update to Twig 3. Are there any other breaking changes to consider?

Not that I see.

PR #1241, opened 2021-02-01
Core: Upgrade league/container

### LINKED ISSUE #1226: phpleague ReflectionContainer uses deprecated ReflectionParameter::getClass
**Describe the bug**
The ArgumentResolverTrait in phpleague 2.4.1 still uses the deprecated ReflectionParameter::getClass method. This issue is fixed in [eb23e27](https://github.com/thephpleague/container/commit/75f01de3d99d8a36bd5a5567090540efe245bcb6#diff-c8fbc16c578d79040b51d66616a591a41fa8993f58ff07217c3e82a4827a4957), which got merged into [v3.3.3](https://github.com/thephpleague/container/releases/tag/3.3.3).

As for the 2.x branch, the last update was 2016. And the getClass call is still there:
https://github.com/thephpleague/container/blob/7b57d9455cac5fd1107ba1836a293cda64371fa3/src/Argument/ArgumentResolverTrait.php#L54

The most obvious way to fix this is to upgrade the "phpleague/container" package. It's 3.x branch claims to support PHP 7.0 - PHP 8.0 (experimental) while its 2.x branch supports PHP up to 7.1 only. One difference between this and #1214 is that it is not a development dependency. Changing the container library would change a bunch of files in /vendor unless either #1123 or #1209 is merged first.

**To Reproduce**
1. Checkout GibbonEdu latest source commit (e.g. 18d25c8)
2. Using the source folder as docroot, start a PHP 8 test server with `display_error=1` and `error_reporting=E_ALL` (i.e. showing deprecated warnings). For example with this command:
    ```
    php --server 127.0.0.1:8888 \
            --docroot . \
            --define display_startup_errors=1 \
            --define display_errors=1 \
            --define error_reporting=E_ALL
    ```
3. Visit the site with browser. The message should be easily seen.

**Expected behavior**
Do not see this kind of message at all.

**Screenshots**
2021-02-02 12-41-45 的螢幕擷圖
2021-01-27 23-45-40 的螢幕擷圖
2021-01-28 00-02-05 的螢幕擷圖

**Your System**
 - Gibbon version: v22.0.00 latest
 - Browser: PhpBrowser of Codeception

yookoala:
**Description**
* Fix reference to deprecated ReflectionParameter::getClass method.
* Only change the composer.json and composer.lock. The changes to the `/vendor` folder are not included here, so this will not work without a local `composer install`. Users who don't know composer will have to rely on #1209 bundle for simple installation.

**Motivation and Context**
* Fix #1226
* Depends on #1209.

**How Has This Been Tested?**
Locally and in CI environment

Note: CI test runs `composer install` before tests so things works there actually means the library upgrade works.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -41,7 +41,7 @@
         "ext-gettext": "*",
         "ext-PDO": "*",
         "google/apiclient": "v2.2.0",
-        "league/container": "^2.4",
+        "league/container": "^3.3.3",
         "aura/sqlquery": "3.*-dev",
         "tecnickcom/tcpdf": "6.0.038",
         "twig/twig": "^2.0",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "742368e59c40f1eb9b7d8f116f7af49d",
+    "content-hash": "fe4abccf405facac24e05de854d764a6",
     "packages": [
         {
             "name": "aura/sqlquery",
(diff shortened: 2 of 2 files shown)


yookoala:
## Way to Test This
1. Create a fork (or use existing fork) of GibbonEdu/core.
2. Tag any commit in it. Remember to push the tag to your fork repository.
3. Create a Draft Release
    2021-02-04 17-27-13 的螢幕擷圖
4.  Select the newly created tag. And fill in the Title / Description as you please.
    2021-02-04 17-27-49 的螢幕擷圖
    2021-02-04 17-32-36 的螢幕擷圖
5. Publish the Release.
6. The newly created release will trigger GitHub Action to run:
    2021-02-04 17-58-57 的螢幕擷圖
7. After the release action is done, the bundle zip and tar.gz file will be added to the newly created release:
    2021-02-04 18-01-35 的螢幕擷圖
    2021-02-04 18-01-55 的螢幕擷圖

SKuipers (approved):
Another great step forward! 👍  We're no longer stuck supporting PHP 5, so we can definitely start updating these packages more frequently.

PR #1242, opened 2021-02-01
Fixed MultiSelect buttons not stacking and not expanding dynamically

raynichc:
**Description**
With the new theme, the MultiSelect takes up more space and as a result the Add and Remove buttons do not properly stack. They also do not properly expand. 

**How Has This Been Tested?**
Locally with both the original and new theme.

**Screenshots**
**Before:**
image
**After:**
image
**Before (old theme):**
image
**After (old theme):**
image


SKuipers (approved):
Looks good! This is a handy form element, we need to start using it a bit more in the core 👍

PR #1243, opened 2021-02-01
OOification: Return Process

raynichc:
**Description**
OOification of Return Processing. This PR deprecates the old returnProcess method and implements a new ReturnMessage class which is handled by Page to process and render return messages. ReturnMessage has methods to set the edit link for a page and add custom returns for processing. For backwards compatibility, the returnProcess method calls the new ReturnMessage methods to set the relevant details.

**How Has This Been Tested?**
Locally on several modules with untouched returnProcess methods and pages altered to use new system.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -1777,100 +1777,11 @@ function isCommandLineInterface()
 }
 
 /**
- * Easy Return Display Processing. Print out message as appropriate.
- * See returnProcessGetAlert() for more details.
- *
- * @param string $guid
- *      The guid of your Gibbon Install.
- * @param string $return
- *      The return value of the process.
- * @param string $editLink
- *      (Optional) This should be a link. The link will appended to the end of a success0 return.
- * @param array $customReturns
- *      (Optional) This should be an array. The array allows you to set custom return checks and
- *      messages. Set the array key to the return name and the value to the return message.
- *
- * @return void
+ * @deprecated in v22. Use Page's ReturnMessage.
  */
 function returnProcess($guid, $return, $editLink = null, $customReturns = null)
 {
-    $alert = returnProcessGetAlert($return, $editLink, $customReturns);
-
-    echo !empty($alert)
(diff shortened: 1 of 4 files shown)


SKuipers:
Looking good, nice to see another function removed from functions.php, and the oo approach is nice and clear. 

I notice a switch in logic from `!$session->has('address')` to `$session->has('address')`. I think the former code was in there to handle the dashboards and welcome pages, whereas the returnProcess has been handled individually on pages in the past. I'd like to ideally remove all of the return code from pages, and I think your ooified code can help with this, if we get rid of the `$session->has('address')` check logic entirely. Then, we'll need to do a mass find-replace to delete the existing returnProcess calls, and only add in the `$page->return->addReturns` and `$page->return->setEditLink` calls as needed for those cases. 

It looks like, if I merge the PR currently, we'll see duplicate return messages on all the non-dashboard pages. We may need to get some parallel PRs ready to remove the returnProcess function calls from the codebase, so we can apply these changes at the same time 🤔

raynichc:
Hi Sandra, thanks for having a look. Though I am a little confused by what you mean. As part of the PR, I've disabled the returnProcess method, so instead of printing out a return message it sets the appropriate data using the new methods and rendering is handled by the new system. So there should be no duplicates, right?
I'll also remove the address logic.

SKuipers (approved):
Aha, thanks for clarifying, I miss-read how the old function was being handled. Looks great! Going to merge and test 👍

SKuipers:
I've used some regex searches to remove/replace the existing returnProcess calls. You can see the commits [here](https://github.com/GibbonEdu/core/commit/c7b24d9d4983a1610f3b4837d47fc645601b69d5) and [here](https://github.com/GibbonEdu/core/commit/86730c8a273c072263fc0845a7ae2bfe0623fb94).

raynichc:
Brilliant, thanks Sandra!

PR #1244, opened 2021-02-04
Fixed errors when DataSet is empty

raynichc:
**Description**
When an empty DataSet is given to a select with the function from DataSet, a warning and a notice will pop up.

**How Has This Been Tested?**
Locally.

**Screenshots**
image


yookoala:
I think the better approach would be to simply add a line like this before line 124:

```php
$options = [];
```

SKuipers (approved):
Thanks! Looks good.

PR #1248, opened 2021-02-14
OOification: Set Log

raynichc:
**Description**
Implemented setLog function into LogGateway to replace setLog in functions.php. The new function doesn't require a gibbonModuleID, but instead the module name, making logging easier.

**How Has This Been Tested?**
Locally with the Help Desk.

### DIFF
--- a/src/Domain/System/LogGateway.php
+++ b/src/Domain/System/LogGateway.php
@@ -114,7 +114,7 @@ class LogGateway extends QueryableGateway
         return $this->db()->delete($sql, $data);
     }
 
-    public function setLog($gibbonSchoolYearID, $module, $gibbonPersonID, $title, $array = null, $ip = null)
+    public function addLog($gibbonSchoolYearID, $module, $gibbonPersonID, $title, $array = null, $ip = null)
     {
         $serialisedArray = is_array($array) ? serialize($array) : null;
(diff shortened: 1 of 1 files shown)


src/Domain/System/LogGateway.php
```diff
@@ -113,4 +113,29 @@ public function purgeLogs($title, $cutoffDate)
 
         return $this->db()->delete($sql, $data);
     }
+
+    public function setLog($gibbonSchoolYearID, $module, $gibbonPersonID, $title, $array = null, $ip = null)
```
yookoala, line 117: This function is adding a new log entry to the database, but using the verb "set" seem to imply a key-value storage. Wouldn't it better named as "appendLog", "addLog" or "log"?

raynichc: I was thinking that, but I kept it as is to keep inline with the current naming convention. Perhaps a change is for the better though.

SKuipers: Thanks Ray and Koala. I agree, updating the method name to align with conventions makes sense. I think `addLog` would work here.

raynichc: Sounds good, I've updated the name accordingly

SKuipers (approved):
Looks good! 👍

PR #1249, opened 2021-02-15
fix GridView

leofreitas:
**Description**
* Added Criteria instance and id to element dataTable in GridView

**Motivation and Context**
* Element dataTable needs a Criteria instance

**How Has This Been Tested?**
 Local and Travis

**Screenshots**

departments

### DIFF
--- a/modules/Departments/departments.php
+++ b/modules/Departments/departments.php
@@ -35,9 +35,14 @@ if (isActionAccessible($guid, $connection2, '/modules/Departments/departments.ph
 
     $departmentGateway = $container->get(DepartmentGateway::class);
     
+    // QUERY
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Thanks for updating this! Hmm. The grid view is not paginated by default, so I don't think it should require a $criteria object. I'll check the scripts, there may be something in there to fix to address the javascript error.

leofreitas:
> Thanks for updating this! Hmm. The grid view is not paginated by default, so I don't think it should require a $criteria object. I'll check the scripts, there may be something in there to fix to address the javascript error.

Ok, but I don't think that should be a problem. we can keep required $criteria object

PR #1251, opened 2021-02-16
Rewrite all setLog calls to LogGateway::addLog

yookoala:
**Description**
Rewrite all setLog calls to LogGateway::addLog

**Motivation and Context**
* Adapt the change from #1248 and 37185b3.

**How Has This Been Tested?**
CI platform.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -20,6 +20,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 use Gibbon\Forms\Form;
 use Gibbon\Services\Format;
 use Gibbon\Contracts\Comms\Mailer;
+use Gibbon\Domain\System\LogGateway;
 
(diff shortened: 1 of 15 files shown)


SKuipers (commented):
Hi Koala, this is looking good, thanks for implementing these changes. I wonder about using `$logGateway` instead of `$logger`, for clarity and consistency. This would help differentiate it from any system-level logging classes (such as `monolog` and the `Psr\Log\LoggerInterface`), which could be useful for the future.

yookoala:
@SKuipers: Updated.

I was thinking of rewriting the log instances into a PSR-3 logger implementation (i.e. `$logger->info()` instead of `$logGateway->addLog()`) with customer handler that use `addLog()` underneath. Maybe later.

SKuipers (approved):
Looks great, thanks for refactoring these! 👍

PR #1252, opened 2021-02-16
Remove obsoleted getLog and getLogByID

yookoala:
**Description**
* Remove obsoleted getLog and getLogByID in functions.php

**Motivation and Context**
* The functions are never used in the codebase.

**How Has This Been Tested?**
Locally and CI platform.


raynichc:
To my knowledge, getLog was only ever used in the Help Desk. I believe it still does as the current getLog requires additional filter criteria to be defined in order to operate identically.

yookoala:
Right. Checked against my old [modules listing repo](https://github.com/yookoala/GibbonEduModules) and found only 1 instance of usage here:
https://github.com/GibbonEdu/module-helpdesk/blob/7beb6357496479d30618cba6ca0068c278e10dd6/Help%20Desk/helpDesk_technicianStats.php#L81-L82

Should probably rewrite this first.

@SKuipers: What do you think?

SKuipers (approved):
Great to see some unused code removed!

SKuipers:
Oops, moved a bit too quickly. @raynichc Are you or Ash able to update the Help Desk so that it uses the LogGateway instead?

raynichc:
Yep, I'll get to work on that.

PR #1253, opened 2021-02-20
Activities: OOified Edit Activity

raynichc:
**Description**
Refactored Edit Activity to use Gateways. Replaced Time Slot Inputs to use CustomBlocks, allowing them to be displayed, edited and deleted in the same form. Replaced Activity Staff Gateway with DataTable and implemented new staff delete confirmation page.

**How Has This Been Tested?**
Locally with Demo Data.

**Screenshots**
image


SKuipers:
Hi Ray, I've reviewed your changes and it's looking good. Lots of code changes, so it'll be easier to get a sense of how it works by plating with it post-merge. Great to see the new custom blocks, and more gateway classes in use! Before I merge and test, I wanted to double-check that this is ready to go. Any pending tweaks or changes?

Also, I wonder if it's possible to put both the start and end times on the same line, to help make the custom block more compact? 🤔

raynichc:
Good call on the times, I'll change that. After that there should be no pending tweaks.

CHANGEDB.php
```diff
@@ -775,5 +775,6 @@
 DELETE FROM `gibbonSetting` WHERE scope='Messenger' AND name IN ('messengerLastBubble','messageBubbleBGColor','messageBubbleWidthType','messageBubbleAutoHide');end
 INSERT INTO gibbonCountry VALUES ('South Sudan', '211');end
 INSERT INTO `gibbonSetting` (`scope`, `name`, `nameDisplay`, `description`, `value`) VALUES ('User Admin', 'publicRegistrationAllowedDomains', 'Public Registration Allowed Domains', 'Comma-separated list of email address domains allowed when registering. Leave blank for no restriction.', '');end
+ALTER TABLE `gibbonActivitySlot` CHANGE `gibbonSpaceID` `gibbonSpaceID` int(10) UNSIGNED ZEROFILL DEFAULT NULL;end
```
SKuipers, line 778: Hi Ray! I see you merged in the latest changes, thanks. Looks like this line accidentally ended up before the end of the list, which will break the updater. Can you move it to the end? Then it looks like it's ready to merge. Thanks!

SKuipers (changes requested):

CHANGEDB.php
raynichc: Hi Sandra. I've moved it to the end and merged in the most recent changes.

SKuipers (approved):
Thanks Ray for your work on this, I'm going to merge it in and check it out 👍

PR #1255, opened 2021-02-24
WIP: Monolog bridge to LogGateway

yookoala:
**Work in progress**. Please comment on the approach and see if anything needs to be changed.

**Description**
* Create a `Monolog\Handler\HandlerInterface` implementation specifically to bridge old log format to [LogGateway](https://github.com/GibbonEdu/core/blob/77c959110139391c99e67e0cb3df50c1e977f2c6/src/Domain/System/LogGateway.php).
* Update existing [LoggerFactory](https://github.com/GibbonEdu/core/blob/77c959110139391c99e67e0cb3df50c1e977f2c6/src/Services/LoggerFactory.php) to use the new Handler.
* (TODO) Rewrite all `LogGateway::addLog` calls to `Monolog::notice` call with `LogHandler::context`.

**Motivation and Context**
A follow up to #1251. Bridging old logging implementation.
Pave way to use Monolog or even PSR-3 Logger in the future.

**How Has This Been Tested?**
Locally.

### DIFF
--- a/src/Domain/System/LogHandler.php
+++ b/src/Domain/System/LogHandler.php
@@ -1,0 +1,117 @@
+<?php
+/*
+Gibbon, Flexible & Open School System
+Copyright (C) 2010, Ross Parker
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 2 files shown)


yookoala:
Couldn't find an elegant way to do this at the moment. Will revisit this in the future.

PR #1256, opened 2021-02-24
Added Filter Rules to LogGateway

raynichc:
**Description**
Added filter rules for filters previously provided by getLog function.

**How Has This Been Tested?**
Locally with the Help Desk.

### DIFF
--- a/src/Domain/System/LogGateway.php
+++ b/src/Domain/System/LogGateway.php
@@ -74,6 +74,37 @@ class LogGateway extends QueryableGateway
                     ->where('gibbonLog.gibbonPersonID = :gibbonPersonID')
                     ->bindValue('gibbonPersonID', $gibbonPersonID);
             },
+            'module' => function ($query, $module) {
+                return $query
+                    ->where('gibbonModule.name = :module')
+                    ->bindValue('module', $module);
+            },
+            'startDate' => function ($query, $startDate) {
+                return $query
+                    ->where('timestamp >= :startDate')
+                    ->bindValue('startDate', $startDate);
+            },
+            'endDate' => function ($query, $endDate) {
+                return $query
+                    ->where('timestamp <= :endDate')
+                    ->bindValue('endDate', $endDate);
+            },
+            'array' => function ($query, $array) {
+                $array = unserialize($array);
+                if (is_array($array)) {
+                    $count = 0;
+                    foreach ($array as $key => $value) {
+                        $bindKey = 'key' . $count;
+                        $bindValue = 'value' . $count;
+
+                        $query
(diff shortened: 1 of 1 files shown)


src/Domain/System/LogGateway.php
```diff
@@ -74,6 +74,37 @@ public function queryLogs(QueryCriteria $criteria, $gibbonSchoolYearID)
...
+                    $count = 0;
+                    foreach ($array as $key => $value) {
+                        $bindKey = 'key' . $count;
+                        $bindValue = 'value' . $count;
+
+                        $query
+                            ->where('serialisedArray LIKE CONCAT("%", :'.$bindKey.', "%;%", :'.$bindValue.', "%")')
+                            ->bindValue($bindKey, $key)
+                            ->bindValue($bindValue, $value);
+                    }
```
yookoala, line 104: @raynichc: If I read it correctly, you planed to increment the `$count` variable in the foreach loop, right? It seems to miss a `$count++` in your loop.

raynichc: Good catch, I think I accidentally dropped it when I copied the code. I'll fix that up. Thanks!

yookoala (approved):

src/Domain/System/LogGateway.php
```diff
@@ -74,6 +74,39 @@ public function queryLogs(QueryCriteria $criteria, $gibbonSchoolYearID)
...
+            'array' => function ($query, $array) {
+                $array = unserialize($array);
+                if (is_array($array)) {
+                    $count = 0;
+                    foreach ($array as $key => $value) {
+                        $bindKey = 'key' . $count;
+                        $bindValue = 'value' . $count;
+
+                        $query
+                            ->where('serialisedArray LIKE CONCAT("%", :'.$bindKey.', "%;%", :'.$bindValue.', "%")')
```
SKuipers, line 101: Looks like some useful changes. I'm curious about the use case for the array filter. Is there a place in the codebase where this is needed, or is it hypothetical?

I'm looking at the LIKE syntax and I wonder if the % wildcards are too open ended, having a % between the key and the value means that any number of other values could come in the middle, so it may match a value for the wrong key. Perhaps a REGEXP may be more accurate here.

raynichc: As far as I can tell, its only ever used in the Help Desk when gathering logs from a specific technician, which is stored in the serialized array. I'll try experimenting with REGEXP.

SKuipers: Okay, if its only ever used in that one spot, I'll merge it now and leave it to you to refine the filter later if you run into any issues there. Thanks!

SKuipers (approved):

PR #1257, opened 2021-02-25
Number Spinner input

powertashton:
**Description**
This extends the current number input to introduce a fresh look with nice buttons (mobile friendly!) for incrementing and decrementing quantities

**Motivation and Context**
I felt the current number input to not be fitting my needs and decided to give a shot at creating this style of input to use in my own development of other things

**How Has This Been Tested?**
Locally

**Screenshots**

image


yookoala:
Can we use [number input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) instead of text? We can [hide the default browser spinner](https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp) to make things look consistent. Using number input would tell browser (especially mobile browser) how user can interact with it. Mobile devices would know to show the number pad, instead of whole keyboard, for user input.

powertashton:
> Can we use [number input](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number) instead of text? We can [hide the default browser spinner](https://www.w3schools.com/howto/howto_css_hide_arrow_number.asp) to make things look consistent. Using number input would tell browser (especially mobile browser) how user can interact with it. Mobile devices would know to show the number pad, instead of whole keyboard, for user input.

I actually asked sandra the same thing, because I was curious why it was done in that way, I'm certainly not opposed to making such a change.

yookoala:
If I have to guess, that was probably some legacy coding issue. HTML5 form took some time to be accepted. And I think it's quite mature nowadays.

yookoala:
@powertashton: The CI installer timeout issue has been fixed in #1259. Can you please [rebase](https://www.freecodecamp.org/news/an-introduction-to-git-merge-and-rebase-what-they-are-and-how-to-use-them-131b863785f/) your commits here onto the latest v22.0.00 branch then force push to your branch?

SKuipers:
Looks like a useful improvement. I wonder about the look of the spinner, the icons feel a bit clunky. Is is possible to do something closer to the image you shared on Slack?

image

powertashton:
I've now updated to look like this,
image
I could add some additional CSS styling if you think it's necessary, but I think this fits in with the general gibbon theme (and uses the existing inline-button class)

SKuipers (approved):
Nice, thanks for making these tweaks. I'm going to merge it in and check it out 👍

SKuipers:
Hi Ash. In testing the input, I discovered that the validation was not working and that more than one spinner in the same form causes all the spinners to change on click. I've pushed a fix, which you can see here: https://github.com/GibbonEdu/core/commit/0a0e9227653330ec7d873b0eeb3fd6f4644729d0 I've also taken the liberty to tweak the layout slightly, so that it's the same height as other text fields.

Screenshot 2021-03-12 at 11 51 20 AM

PR #1260, opened 2021-03-01
System: refactor the getCustomFields function into CustomFieldHandler class

SKuipers:
Second piece of the custom fields changes (and another one to come). This step refactors and removes the getCustomFields function from User Admin and creates a CustomFieldHandler class. The handler class contains methods for adding custom fields to forms, updating/uploading custom field data in process pages, and displaying custom field data in details tables. This refactors several sections of copy-pasted code into reusable methods.

**How Has This Been Tested?**
Locally

**Screenshots**
There are not too many visual changes in this PR, that comes next.

### DIFF
--- a/modules/User Admin/user_manage_edit.php
+++ b/modules/User Admin/user_manage_edit.php
@@ -19,6 +19,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use Gibbon\Forms\Form;
 use Gibbon\Services\Format;
+use Gibbon\Forms\CustomFieldHandler;
 use Gibbon\Forms\DatabaseFormFactory;
 use Gibbon\Domain\System\DataRetentionGateway;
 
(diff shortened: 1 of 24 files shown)


modules/User Admin/user_manage_edit.php
```diff
@@ -36,7 +37,9 @@
 
     $returns = array();
     $returns['warning1'] = __('Your request was completed successfully, but one or more images were the wrong size and so were not saved.');
-    $page->return->addReturns($returns);
+    if (isset($_GET['return'])) {
```
rossdotparker, line 40: This seems to be undoing some OOification work...?

rossdotparker:
This was a lovely, DRY PR to read: a very elegant solution, with a lot of repeated cruft sloughed off. Going to merge now, and then test. The one comment I left can be addressed post-merge if needed.

modules/User Admin/user_manage_edit.php
SKuipers: Oh! Must have accidentally reverted that when rebasing, will push the fix post-merge. Thanks!

PR #1264, opened 2021-03-03
Students: remove the extendedBriefProfile setting from student profiles

SKuipers:
This PR removes the extendedBriefProfile option, which was added some time in the past to display additional parent information on student profiles. This has since been deemed unnecessary and potentially too much visibility of parent data. The setting and it's related section in the student profile have been removed.

**Motivation and Context**
Part of an overall review of data protection and privacy measures in Gibbon.

**How Has This Been Tested?**
Locally


rossdotparker:
Tests out nicely : ) The only thing it does not do is to clean up the setting from the database.

SKuipers:
Aha right, I'll push that directly.

SKuipers:
Added to #1265 so as to not immediately invalidate my PR 😄

PR #1267, opened 2021-03-04
System Admin: add a composer.lock hash check to the Update page

SKuipers:
In preparation for merging #1209, we're looking at how we can help inform developers of the change in workflow, as well as ensure their systems remain up to date with the cutting edge code. This PR adds a check to the Update process to see if the content-hash in the composer.lock file has changed, and prompts users if they need to run a composer install.

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2021-03-04 at 1 11 36 PM


yookoala:
This is brilliant!

PR #1268, opened 2021-03-08
Automate Lesson plan naming

fvlasie:
A simple change to creating Lesson plans that does not require that the Lesson be named. This makes it a bit quicker for teachers to add lessons. If the Lesson name is blank it is set to "Lesson".

### DIFF
--- a/modules/Planner/planner_add.php
+++ b/modules/Planner/planner_add.php
@@ -187,7 +187,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/planner_add.php') 
 
             $row = $form->addRow();
                 $row->addLabel('name', __('Lesson Name'));
-                $row->addTextField('name')->setValue()->maxLength(50)->required();
+                $row->addTextField('name')->setValue()->maxLength(50);
 
             $row = $form->addRow();
                 $row->addLabel('summary', __('Summary'));
--- a/modules/Planner/planner_addProcess.php
+++ b/modules/Planner/planner_addProcess.php
@@ -54,7 +54,10 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/planner_add.php') 
             $timeStart = $_POST['timeStart'];
             $timeEnd = $_POST['timeEnd'];
             $gibbonUnitID = !empty($_POST['gibbonUnitID']) ? $_POST['gibbonUnitID'] : null;
-            $name = $_POST['name'] ?? 'Lesson';
+            $name = $_POST['name'] ?? '';
+            if (empty($name)) {
+               $name = 'Lesson for ' . $gibbonCourseClassID;
+            }
             $summary = $_POST['summary'] ?? '';
             if (empty($summary)) {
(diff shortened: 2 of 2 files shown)


yookoala:
The tests are failed not caused by this PR but by previous commits. Added #1269 to fix this issue. Please wait to merge upstream when that PR (or equivalent fixes) are made to the v22.0.00 branch. Thanks.

rossdotparker:
Hi FV, this is an interesting change, but I wonder if we run the risk of having endless lessons called "Lesson", much like some students have many "Untitled Folders" on their desktops. I think the lesson name is an important element. Wonder what @SKuipers thinks? Thanks for taking the time all the same...

fvlasie:
I think not _requiring_ the field makes sense since it eliminates typing. Having the lesson all with the same name is not so bad since they are differentiated by course and time slot. (I actually had people asking me what was supposed to be put in the name field!) If people want to name their lessons they are still welcome to.

With this new scheme creating a quick lesson is 3 steps: 1) Press the + symbol 2) Paste in your content 3) Click Submit.

fvlasie:
As a compromise perhaps we could make the automatic naming smarter. "Lesson for shortclassname"? :)

 But I am not smart enough to get the shortclassname in the code. :(

SKuipers:
Hi FV. I can certainly see the aim here. However, like Ross, I wonder if a feature like this could make it ultimately harder for teachers to manage their lessons. In the Lesson Planner, if all the lessons have the same name, then it makes it harder to find and edit a specific lesson. For students and teachers, the Today's Lesson gives them a reminder on what's coming up for that day. It may take teachers a few moments longer to fill in the form and name the lesson (eg: Unit 1 intro, Vocabulary review, Assessment, etc). but it does help give students, parents and teachers a better idea of the content of lessons throughout the year.

fvlasie:
Okidoki!

fvlasie:
One last desperate idea! :)
How about making "Require Lesson Names" a setting (on by default)?

fvlasie:
I understand! Thank you for taking the time to explain it. Woud you be adverse to changing the label to "Lesson Title" rather than "Lesson Name"? This is an old point of confusion on our end and I think that might make the fileds purpose a little clearer. (I don't know if you remember but the filed was originally labeled "Name" and I changed it to "Lesson Name" because techers here were wondering whose name to put.) :)

As it is right now, here, our teachers are just putting _something_ in the field because it is required:
Screenshot 2021-03-10 at 10 24 20 AM

rossdotparker:
Hi FV, rather than a code change, I think this would be best done using String Replacement (under Admin > System Admin). We've not had any other reports of issues with this, and from a consistency point of view, we use "Name" everywhere else in Gibbon.  What do you think?

On reflection "Title" might well be better than "Name", but "Name" is now so well established in the system!

fvlasie:
That makes good sense!

SKuipers:
Although I do hate to close a PR, I think we're all wrapped up here, and it has certainly served its educational purpose 😄 Thanks again FV!

PR #1272, opened 2021-03-10
Display dashboard tab (Enrolment) for high resolution screens fixed

leofreitas:
**Description**
The value of  property xl:overflow-x-unset changed to xl:overflow-x-scroll

**Motivation and Context**
When the button "sidebarToggle" is clicked, the display tab (Enrolment) on high-resolution screens not working correctly.

**How Has This Been Tested?**
Local in high resolution 1920 x 1080

**Screenshots**
v22_dashboard_error

### DIFF
--- a/resources/templates/index.twig.html
+++ b/resources/templates/index.twig.html
@@ -69,7 +69,7 @@ TODO: add template variable details.
                         <br style="clear: both">
                     {% endif %}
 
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Thanks Leo, this is an interesting error. It looks like it happens because the chart.js element scales up when the space gets larger, but doesn't scale back down. I've tweaked your solution so that it still works, but only applies to the dashboard page (identified by a lack of moduleMenu). The reason the somewhat odd `xl:overflow-x-unset` class exists is that sticky headers don't work unless they're in a container that does not have any overflow set. The data tables have sticky headers on scroll, which is handy, and requires the `xl:overflow-x-unset` class (set through the preventOverflow variable). Complex explanation for a small change 😆 Thanks again!

PR #1278, opened 2021-03-19
System Admin: enable custom fields to be listed under specific headings

SKuipers:
This PR contains a number of general fixes and changes that make implementing and managing custom fields easier. The overall aim is for custom fields to more seamlessly be added to areas of the system, rather than tacked on at the end under the heading 'Custom Fields'.

- Improves form rendering so rows are grouped by heading and rendered in order. This enables a form field to be added to a specific heading after the rest of the form is declared. It's backwards compatible and doesn't affect existing forms and headings.
- Enables detail tables to make use of nested columns, which are already supported by the base DataTable objects. This allows a page of details with headings to be rendered as a single table rather than as separate tables. It also enables adding a field under a specific heading.
- Enables selecting a heading for a custom field from a drop-down or entering a new custom heading. Existing headings are listed based on the selected context.
- Splits the list of fields on the Custom Field page to group the fields by context. This makes the drag-drop ordering more intuitive.
- Refactors the Personal page of the Student Profile and enables it to support the display of custom fields.
- Renames the existing 'Person' context to 'User' for better clarity and naming (eg: User Fields vs Person Fields)

**How Has This Been Tested?**
Locally.

**Screenshots**
An example of custom fields added to an existing heading, a new heading, and no heading (defaults to Other Information)
Screenshot 2021-03-19 at 2 02 06 PM


rossdotparker:
This is a really significant set of developments, and definitely adds considerably to the power and flexibility of Gibbon : )

PR #1284, opened 2021-03-25
System Admin: prevent installing or updating a module that requires a newer core version

SKuipers:
This PR adds a `$coreVersion` variable to a module's version.php file. This enables tagging which core version a module requires. If a module does not meet the minimum requirement, it cannot be installed or updated. It currently ignores version.php files that do not have a $coreVersion defined, which will enable us to add them in future module releases.

**Motivation and Context**
Help prevent errors with modules that require a different core version.

**How Has This Been Tested?**
Locally.

**Screenshots**
Screenshot 2021-03-25 at 11 57 10 AM

Screenshot 2021-03-25 at 11 53 09 AM


rossdotparker:
I'm excited to see this one in Gibbon...it's something I've had in mind for ages, so thanks for building it.

PR #1290, opened 2021-03-28
Review of translation support and some improvements

leofreitas:
**Description**
Review of translation support and some improvements

**Motivation and Context**
In user_manage_editProcess.php I applyed the same logic as the user_manage_addProcess.php for the gibbonSchoolYearIDClassOf field, because if It's empty, the record receives 000 and this generates an error.

**How Has This Been Tested?**
Local and Travis

**Screenshots**
user_error


SKuipers (approved):
Thanks Leo, looking good 👍 I've made a couple small tweaks, such as the yesNo formatter, which already translates the content. Otherwise looks ready to merge and test. Thanks!!

PR #1291, opened 2021-03-29
New function genderName and adjusting reports

leofreitas:
**Description**
- Created new function to return the full gender name with translation support
- Using the new function in student/staff view details and adjusting the reports to genders (others and unspecified)

**Motivation and Context**
Code Review

**How Has This Been Tested?**
Local and Travis


SKuipers (approved):
Looks great, thanks! 👍  Yes, this makes sense (in hindsight) that this wouldn't be easily translatable from the database without a function to convert it.

PR #1297, opened 2021-04-09
Stop the blue border!

fvlasie:
Stop a blue border appearing around the tab when clicked. (It was driving me crazy!) :~{

### DIFF
--- a/themes/2021/css/main.css
+++ b/themes/2021/css/main.css
@@ -526,6 +526,7 @@ ul.ui-tabs-nav li.ui-state-default {
 }
 ul.ui-tabs-nav li.ui-state-default a {
     color: rgba(0, 0, 0, 1);
+    outline: none;
 }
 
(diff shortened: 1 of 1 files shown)


yookoala:
Did you try it on browser? From my past experience, instead of applying it to `a`, I think you probably need to apply `outline: none` to `a:active` and `a:focus` to make it work.

(I have not tested your CSS on browser. This is simply from past experience.)

fvlasie:
I did test it in a browser and it worked. I suppose there are several ways to accomplish the same thing and I am not sure which is best. This will work as long as the browser does not decide to apply the outline effect further down the CSS chain.

SKuipers (approved):
Looks like a handy fix. I tend to use Firefox so I hadn't seen this issue, but looks like it must have been happening in other browsers. Thanks! 👍

PR #1298, opened 2021-04-12
Experimental php 8.1 test

yookoala:
**Description**
Modify GitHub action workflow to add the developing PHP 8.1 as an experimental target. It means that CI workflow will test the target, but failure of it will not stop PR from merging.

**Motivation and Context**
PHP 8.1 is [now supported](https://github.com/marketplace/actions/setup-php-action#tada-php-support) in GitHub Actions. Adding it as a target could provide insight for Gibbon's potential forward compatibility issue, if any. Tests are run in parallel. There should be no perceivable time difference after adding the target.

2021-04-12 09-48-40 的螢幕擷圖

**How Has This Been Tested?**
In CI environment.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ jobs:
         php-version: ["7.3", "7.4", "8.0"]
         experimental: [false]
         name: ["CI Test"]
+        include:
+          - php-version: "8.1"
+            experimental: true
+            name: "CI Test, Experimental"
     services:
       mysql:
         image: mysql:5.7


yookoala:
PHPUnit seems to now trigger a PHP 8.1 segfault error. Not really the issue of PHPUnit or Gibbon.

There are some deprecated message triggered by Gibbon. We'd probably want to fix them in the future. But before fixing those, we can still run the tests and pass without any issue.

SKuipers (approved):
Thanks Koala, exciting to be a bit ahead of the curve! 😁

PR #1300, opened 2021-04-16
Themes: set 2021 as the new Default theme, rename old theme to Legacy

SKuipers:
In preparation for the v22 release, this PR sets the new theme as the Default, and renames the old theme to Legacy. The old theme is still available in the core, but not installed by default. This PR looks big, but the majority of the changes are files being moved around and being replaced.

Since these changes also affect the appearance of the installer, I've added a few more tweaks to the installer to make it look nice and clean, including a multi-part progress indicator. This indicator is now built-into the Form class, so any multi-part forms can make use of it.

**How Has This Been Tested?**
Locally

**Screenshots**
Step 2:
Screenshot 2021-04-16 at 10 33 30 AM

Step 4:
Screenshot 2021-04-16 at 10 34 41 AM


themes/Default/credits.txt
```diff
@@ -17,6 +17,3 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 THEME CREDITS
```
rossdotparker, line 18: Can this file now be deleted?

rossdotparker:
Another big leap in the right direction : )

PR #1303, opened 2021-04-16
Data Updater: add a Staff Data Updater page

SKuipers:
This PR enables staff or other users (depending on permissions) to request updates to Staff records. These function much the same way as other data updates. They can be required via Data Updater Settings, and they show up on the My Data Updates page for staff. Also supports staff custom fields.

**Motivation and Context**
There are some fields, such as job title and biography, that are nice to have staff update themselves.

**How Has This Been Tested?**
Locally.

**Screenshots**
Screenshot 2021-04-16 at 4 26 47 PM

Screenshot 2021-04-16 at 4 26 58 PM


rossdotparker:
This one will definitely be handy. I've had a look through, and it is looking good : ) I'll merge and test now.

PR #1308, opened 2021-04-23
Attendance: table refactoring and print view cleanup

SKuipers:
Refactors the Take Future Attendance table of attendance logs, plus removes print pages that are no longer necessary.


rossdotparker:
Nice to see those unDRY print pages disappear!

PR #1312, opened 2021-04-28
Form Group Changes - Part 2

rossdotparker:
**Description**
Moves the Roll Group -> Form Group change from strings to database and code, particularly highlighting table and field names. Will be followed by other PRs in the coming days for 1) Settings/Variables 2) file names and 3) enums. After this I'll look for 4) edge cases and then 5) Additional modules and finally 6) Query Builder. DON'T PUSH TO PRODUCTION UNTIL THE WHOLE THING IS DONE, SETTLED AND TESTED!

**Motivation and Context**
Roll Groups is US English, but Gibbon uses Form Groups by default. For code base consistency, we are moving to Form Group throughout.

**How Has This Been Tested?**
Locally, but not as thoroughly as it needs to be.


SKuipers:
Hmm, looks like the tests may be failing because it's unable to install a functioning set of demo data. I know you generally try not to update gibbon.sql, but in this case for us to be able to test the changes, I think we'll need to apply these changes to gibbon.sql and gibbon_demo.sql.

Screenshot 2021-04-29 at 11 01 43 AM

PR #1314, opened 2021-04-30
Activities - refactoring $_SESSION - WIP

hmerrettICHK:
**Description**
Work in progress - first tries at refactoring $_SESSION code - please take a look and see if I'm on the right lines

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


SKuipers (commented):
Hi Harry, these are looking good, looks like you're off to a good start.

modules/Activities/report_activitySpread_formGroup.php
```diff
@@ -97,7 +97,7 @@
         ->sortable(['surname', 'preferredName'])
         ->format(function ($student) use ($guid) {
             $name = Format::name('', $student['preferredName'], $student['surname'], 'Student', true);
-            return Format::link($_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/Students/student_view_details.php&gibbonPersonID='.$student['gibbonPersonID'].'&subpage=Activities', $name);
+            return Format::link($session->get('absoluteURL').'/index.php?q=/modules/Students/student_view_details.php&gibbonPersonID='.$student['gibbonPersonID'].'&subpage=Activities', $name);
```
SKuipers, line 100: I suspect that this would throw an error. Check the syntax but also consider the lines above and below this line: what do you notice?

modules/Activities/report_participants.php
```diff
@@ -39,15 +39,15 @@
...
 
-        $form = Form::create('filter', $_SESSION[$guid]['absoluteURL'].'/index.php','get');
+        $form = Form::create('filter', $session->get('absoluteURL').'/index.php','get');
 
         $form->setTitle(__('Choose Activity'));
         $form->setFactory(DatabaseFormFactory::create($pdo));
         $form->setClass('noIntBorder fullWidth');
 
-        $form->addHiddenValue('q', "/modules/".$_SESSION[$guid]['module']."/report_participants.php");
+        $form->addHiddenValue('q', "/modules/".$session->get($guid)['module']."/report_participants.php");
```
SKuipers, line 48: Check the syntax here. What is the variable we're looking to get?

modules/Activities/report_participants.php
```diff
@@ -89,7 +89,7 @@
         ->sortable(['surname', 'preferredName'])
         ->format(function ($student) use ($guid) {
             $name = Format::name('', $student['preferredName'], $student['surname'], 'Student', true);
-            return Format::link($_SESSION[$guid]['absoluteURL'].'/index.php?q=/modules/Students/student_view_details.php&gibbonPersonID='.$student['gibbonPersonID'].'&subpage=Activities', $name);
+            return Format::link($session->get('absoluteURL').'/index.php?q=/modules/Students/student_view_details.php&gibbonPersonID='.$student['gibbonPersonID'].'&subpage=Activities', $name);
```
SKuipers, line 92: This looks like it would also throw an error. Read this line and the surrounding lines and consider why this might be the case.

SKuipers (changes requested):
Hi Harry, I've added some comments and areas to check out, but I haven't necessarily provided the solution. Be sure to read the code and think critically about what it means and how it's being used. Keep going, you're off to a good start!

hmerrettICHK:
Hi, Sandra! I've made the tweaks to the first couple of files (still need to commit them though), but I'm not sure what's going on with report_participants. If you have time, could you walk me through those issues a bit more?

SKuipers (approved):
Looking good, going to merge and test 👍

PR #1317, opened 2021-05-05
Form Group Changes - Part 5 - final sweep and enums

rossdotparker:
**Description**
Final sweep for variations occurrences of roll, roll group and rollgroup, as well as enums

**Motivation and Context**
Must get rid of roll groups!

**How Has This Been Tested?**
Some testing, but it is a big set of changes. We'll need to keep a careful eye out in the full system test at the end of this release cycle!


SKuipers (approved):
Fixed one small syntax error, otherwise looking ready to merge 👍

PR #1320, opened 2021-05-10
Attendance - session variable refactoring (WIP)

hmerrettICHK:
**Description**
Attendance, refactoring to fix/update session variables - WIP

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**

### DIFF
--- a/modules/Attendance/attendance.php
+++ b/modules/Attendance/attendance.php
@@ -56,7 +56,7 @@ $form->setClass('noIntBorder fullWidth');
 $form->addHiddenValue('q', '/modules/' . $session->get('module') . '/attendance.php');
 
 $row = $form->addRow();
-$row->addLabel('currentDate', __('Date'))->description($_SESSION[$guid]['i18n']['dateFormat'])->prepend(__('Format:'));
+$row->addLabel('currentDate', __('Date'))->description($session->get('i18n')['dateFormat'])->prepend(__('Format:'));
 $row->addDate('currentDate')->setValue(Format::date($currentDate))->required();
 
 if (isActionAccessible($guid, $connection2, '/modules/Attendance/report_formGroupsNotRegistered_byDate.php')) {
@@ -75,7 +75,7 @@ $page->write($form->getOutput());
 
 
 // define attendance tables, if user is permit to view them
-if (isset($_SESSION[$guid]["username"])) {
+if (isset($session->get("username"))) {
     // generator of basic attendance table
     $getDailyAttendanceTable = function ($guid, $connection2, $currentDate, $rowID, $takeAttendanceURL) use ($session) {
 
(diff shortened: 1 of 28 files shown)


modules/Attendance/attendance.php
```diff
@@ -75,7 +75,7 @@
 
 
 // define attendance tables, if user is permit to view them
-if (isset($_SESSION[$guid]["username"])) {
+if (isset($session->get("username"))) {
```
SKuipers, line 78: This is one case where an isset() on the return value for the get method is not ideal. There is a $session->has() method that can be used for this case, which also makes the line more readable.
```suggestion
if ($session->has('username')) {
```

modules/Attendance/attendance_take_byCourseClass.php
```diff
@@ -106,7 +106,7 @@
 
                 // Check class
                 try {
-                    $data = array("gibbonCourseClassID" => $gibbonCourseClassID, "gibbonSchoolYearID" => $_SESSION[$guid]["gibbonSchoolYearID"]);
+                    $data = array("gibbonCourseClassID" => $gibbonCourseClassID, "gibbonSchoolYearID" => $session->get("gibbonSchoolYearID"));
```
SKuipers, line 109: For consistency, we do prefer single quotes (string literal) over double quotes here. It's not a syntax issue, but it's nice to have this consistent in the codebase, so see if you can change these going forward.
```suggestion
                    $data = array("gibbonCourseClassID" => $gibbonCourseClassID, "gibbonSchoolYearID" => $session->get('gibbonSchoolYearID'));
```

modules/Attendance/attendance_take_byPerson.php
```diff
@@ -132,7 +132,7 @@
 
                         $output = '<b>'.__($log['direction']).'</b> ('.__($log['type']). (!empty($log['reason'])? ', '.__($log['reason']) : '') .')';
                         if (!empty($log['comment'])) {
-                            $output .= '&nbsp;<img title="'.$log['comment'].'" src="./themes/'.$_SESSION[$guid]['gibbonThemeName'].'/img/messageWall.png" width=16 height=16/>';
+                            $output .= '&nbsp;<img title="'.$log['comment'].'" src="./themes/'.$session->get('gibbonThemeName').'/img/messageWall.png" width=16 height=16/>';
```
SKuipers, line 135: This is inside a function and will throw an error. Be sure to test these as it'll help you catch these kinds of errors.

SKuipers (changes requested):
Hi Harry, you're making good progress, I've added some comments to areas to check out in the code.

modules/Attendance/attendance_take_byPerson.php
SKuipers: It doesn't look like this issue has been resolved yet.

SKuipers (changes requested):
Still some more changes to make.

SKuipers (approved):
Looks ready to merge, thanks!

PR #1322, opened 2021-05-11
System Admin/Library - added new library type for telephones and abil…

hmerrettICHK:
…ity to import telephone data

**Description**
Added a new library item type for telephones, and added a data importer in System Admin to import telephone data

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


resources/imports/libraryRecordTelephone.yml
```diff
@@ -0,0 +1,146 @@
...
+        args: { filter: schoolyear, custom: true }
+        relationship: { table: gibbonSchoolYear, key: gibbonSchoolYearID, field: name  }
+    comment:
+        name: "Comments/Notes"
+        desc: ""
+        args: { filter: string, custom: true  }
+
+    # Type-Specific Serializable Fields
+
+    fieldType:
```
rossdotparker, line 95: I don't think the field Type is needed in the import or in the database. This is useful for the more general "Electronic" library type, but for this new Telephone type we know what we are dealing with.

PR #1339, opened 2021-05-25
Timetable & System Admin: session refactoring

rossdotparker:
**Description**
All $_SESSION calls (expect in functions) converted to $session

**Motivation and Context**
Refactoring

**How Has This Been Tested?**
Locally


SKuipers (approved):
Looking great! Be sure to change those "" to '' string literals as you go too, since we love consistency 😁

PR #1341, opened 2021-05-27
Students: session refactoring

rossdotparker:
**Description**
More session refactoring! Students module this time

**Motivation and Context**
Ongoing refactoring focus

**How Has This Been Tested?**
Locally

**Screenshots**
Does anyone read this far down?!


SKuipers (approved):
Looks good! I'm going to make a couple post-merge tweaks, I think for the sidebar extra, it makes sense to append it all to a variable first, then pass that into $session->set() rather than calling the method every time to append data.

PR #1347, opened 2021-05-28
Messenger: session and data table refactoring

SKuipers:
Refactored the Manage Messages data table and all $_SESSION variables.

### DIFF
--- a/modules/Messenger/messenger_manage.php
+++ b/modules/Messenger/messenger_manage.php
@@ -19,461 +19,242 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 use Gibbon\Forms\Form;
 use Gibbon\Services\Format;
+use Gibbon\Tables\DataTable;
 use Gibbon\Domain\Messenger\MessengerGateway;
 
 $page->breadcrumbs->add(__('Manage Messages'));
 
-if (isActionAccessible($guid, $connection2, "/modules/Messenger/messenger_manage.php")==FALSE) {
-	//Acess denied
-	print "<div class='error'>" ;
-		print __("You do not have access to this action.") ;
-	print "</div>" ;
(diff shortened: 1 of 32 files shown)


modules/Messenger/messenger_manage.php
```diff
@@ -19,461 +19,242 @@
...
+    $page->addError(__('You do not have access to this action.'));
+} else {
+    // Get action with highest precedence
+    $highestAction = getHighestGroupedAction($guid, $_GET["q"], $connection2);
+    if ($highestAction==FALSE) {
+        $page->addError(__('The highest grouped action cannot be determined.'));
+        return;
+    }
+
+    $search = isset($_GET['search'])? $_GET['search'] : '';
```
rossdotparker, line 38: Could this not be a null coalesce?

src/Domain/Messenger/MessengerGateway.php
```diff
@@ -35,8 +35,35 @@ class MessengerGateway extends QueryableGateway
 
     private static $tableName = 'gibbonMessenger';
     private static $primaryKey = 'gibbonMessengerID';
-    private static $searchableColumns = [''];
+    private static $searchableColumns = ['gibbonMessenger.subject', 'gibbonMessenger.body'];
     
+    /**
+     * Queries the list of users for the Manage Users page.
```
rossdotparker, line 41: Looks like a reused doc block. Maybe worth noting that this does not determine which messages to display on, say, the message wall, but is for managing messages.

rossdotparker:
Two refactorings in one : ) I'll merge this now, but there are two comments if you could pick them up post merge. Thanks @SKuipers!

SKuipers:
Thanks! Well caught on those two spots 👍

PR #1353, opened 2021-05-31
Planner: session variable refactoring

hmerrettICHK:
**Description**
Session variable refactoring in Planner module

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


SKuipers (approved):
Hi Harry. This was a massive change set, so overall well done. I did catch nine spots where you had used $session->get rather than $session->set. These wouldn't have generated fatal errors, but in this case would have caused the sidebar content to be missing. Keep an eye on your method calls in the future 👍

PR #1358, opened 2021-06-02
System: final pass on refactoring session variables

hmerrettICHK:
**Description**
System: final pass on refactoring session variables

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


SKuipers (approved):
It's looking ready to merge. Note that I pushed a number of fixes, in the interest of time, including about 7 areas where you were using $session->get rather than $session->set

PR #1365, opened 2021-06-06
Activities - Translation support and code cleanup

leofreitas:
**Description**
Date range translation support and code cleanup

**How Has This Been Tested?**
Local and Travis


SKuipers (approved):
Thanks, looking good! Now that we're into the full system test we're aiming to only focus on bug fixes, rather than more refactoring or cleanup (to prevent introducing new bugs). Since this change looks pretty small and self-contained I'll merge it in and test it out 👍 Thanks!

PR #1366, opened 2021-06-15
System: replaced all usage of PHPExcel library with PHPSpreadsheet

SKuipers:
I discovered that a number of scripts were still using the old PHPExcel library, even though we've since moved to using PHPSpreadsheet. PHPExcel is not compatible with PHP 8, so it was time to clear these old remnants out of the codebase. This PR updates those scripts, since luckily PHPSpreadsheet is a nice drop-in replacement. There were a couple tweaks to make to the output process and the styles, but otherwise these changes test out, and fix a number of export-related issues found during testing.


rossdotparker:
Nice to see some added consistency. I've merged, and will leave you to do any further tests you see fit, given that I know you are already doing a lot of testing on this.

PR #1372, opened 2021-06-25
GitHub actions release update

yookoala:
**Description**
Remove i18n folder content from release bundle.

**Motivation and Context**
* That was the way release was done previously.
* Reduce the file size. Allow people to choose the language pack they needed.

**How Has This Been Tested?**
Created a test build here with this in my fork:
https://github.com/yookoala/GibbonEduCore/releases/tag/v23.0.00-testbuild3

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -70,7 +70,7 @@ jobs:
           echo "::endgroup::"
 
           echo "::group::Create Bundle (zip)"
-          zip --quiet -r "./dist/GibbonEduCore-InstallBundle.zip" . -x "dist/**/*" -x ".git/**/*"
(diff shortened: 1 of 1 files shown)


SKuipers (commented):
Thanks for making these changes! I noticed, in hindsight, that the resources/build submodule is also being included in the release bundle. I think for all submodules, we will want to exclude them from install bundles. Perhaps with this in mind we could remove the submodule recursive section from lines 31-33:

```
        with:
          submodules: "recursive"
          fetch-depth: 1
```

yookoala:
Just remove the `submodules` setting. The default would be not to checkout with submodules, so it should work as we wanted.

PR #1373, opened 2021-06-25
System: improved the appearance of the fatal error message page

SKuipers:
In the past, we've moved the error message page into a template, which is a nice way to ensure it's formatted and looks good. However, in the event that a fatal error occurs, the current error message may be plain text and looks jarring, especially if font-end users see it. 

This PR adjusts the error messages so that the soft-fail messages will use the template, but any hard-fail states will default back to an error.php page that contains enough HTML to render and still look okay. The error.php page does not depend on any functions or libraries, and should always be renderable.

**Screenshots**
Soft-fail:
Screenshot 2021-06-25 at 2 14 00 PM

Hard-fail:
Screenshot 2021-06-25 at 2 13 54 PM

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -26,6 +26,7 @@ v23.0.00
 
     Tweaks & Additions
         System: updated table actions to enable additional modules to use their own icon paths
+        System: improved the appearance of the fatal error message page
         Students: updated student enrolment dropdown to include students with Expected status
 
     Bug Fixes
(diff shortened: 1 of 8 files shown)


rossdotparker:
I've tested this by removing /vendor, and it seems to work well : ) However, if I mangle index.php to produce an error, I just get the plain white page of doom.

SKuipers:
> I've tested this by removing /vendor, and it seems to work well : ) However, if I mangle index.php to produce an error, I just get the plain white page of doom.

Ah, I see. Looks like since everything is routed through index.php, a compile error on index will break things, but any other type of runtime error or exception should be caught. Not sure we can catch the compile errors, but I don't think we'd ever commit a bug that egregious either.

PR #1375, opened 2021-06-29
Comment fix

powertashton:
**Description**
This PR fixes a couple hundred comments that were incorrectly stating that the code was checking if school year specified.

**Motivation and Context**
In working on some code in the library module I noticed something was off with a comment, determined that it had likely been copy pasted from another piece of code, and then wondered how many times it had happened.. The answer was at least 260 times in the core.
image
image

**How Has This Been Tested?**
As the only changes have been to comments the functionality of the code should remain unchanged, hopefully Github Actiosn will agree with me on this.

### DIFF
--- a/modules/Activities/activities_manage_delete.php
+++ b/modules/Activities/activities_manage_delete.php
@@ -27,7 +27,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Activities/activities_mana
     $page->addError(__('You do not have access to this action.'));
 } else {
     //Proceed!
-    //Check if school year specified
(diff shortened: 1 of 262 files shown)


powertashton:
Would it have been easier for me to simply change all the lines of code to "Check if required variables specified"? Yes, but that just wouldn't have had the same effect, would it?

SKuipers (approved):
Awesome 😂 I had seen those around and had been changing them as they crossed my path. No idea there was over 250 instances of them, nicely caught. Love your methodical approach to not only update the comment, but the context of the comment as well. Happy to merge this in 😁

PR #1379, opened 2021-07-05
Reporting: fix issue with empty milestone

### LINKED ISSUE #1378: Error updating Reporting Cycle
**Describe the bug**
On an existing reporting cycle, I attempted to update the end date and could not due to an error.

**To Reproduce**

1. Assess / Reports / Manage Reporting Cycles 
2. Save
3. Resulting in a blank screen
4. In error log:
`
PHP Fatal error:  Uncaught TypeError: array_map(): Argument #2 ($array) must be of type array, string given in /home/gibbon/public_html/modules/Reports/reporting_cycles_manage_editProcess.php:58
Stack trace:
#0 /home/gibbon/public_html/modules/Reports/reporting_cycles_manage_editProcess.php(58): array_map(Object(Closure), '')
#1 {main}
  thrown in /home/gibbon/public_html/modules/Reports/reporting_cycles_manage_editProcess.php on line 58
`

**Expected behavior**
It should save normally. I believe the issue is related to milestones, but I'm not sure why it just appeared. We do not have milestones on this reporting cycle.

This bypassed the error successfully (but it doesn't fix the error it just bypasses milestones altogether):
reporting_cycles_manage_editProcess.php, line 54:

`    // Sort and save milestones as a JSON blob`
`/*    $data['milestones'] = array_map(function ($item) {`
`        $item['milestoneDate'] = Format::dateConvert($item['milestoneDate']);`
`        return $item;`
`    }, $data['milestones']);`
`    $data['milestones'] = array_combine(array_keys($_POST['order']), array_values($data['milestones']));`
`    ksort($data['milestones']);`
`    $data['milestones'] = json_encode($data['milestones']);`
`    */`
`    $data['milestones'] = json_encode(array());`

**Your System**

 - Gibbon version 22.0

yookoala:
**Description**
* In Reporting ([reporting_cycles_manage_editProcess.php](https://github.com/GibbonEdu/core/blob/v22.0.00/modules/Reports/reporting_cycles_manage_editProcess.php#L55-L58)), the submitted value `$_POST['milestones']` is supposed to be an array. However, if there is no milestone submitted, the value `$_POST['milestones']` would be not set.
* The null coercion in [line 48](https://github.com/GibbonEdu/core/blob/v22.0.00/modules/Reports/reporting_cycles_manage_editProcess.php#L48) would provide an empty string as the fallback value, which is not an array. This would cause issue in the array_map in [line 55-58](https://github.com/GibbonEdu/core/blob/v22.0.00/modules/Reports/reporting_cycles_manage_editProcess.php#L55-L58).
* Changing the fallback value from an empty string could fix the issue. (Need original reporter @tr-j verification)
* Best if we can have this in our acceptance test. But there is currently no acceptance test for it. And I'm not familiar with the module enough to write one.

**Motivation and Context**
Fix #1378.

**How Has This Been Tested?**
* Not tested yet. Pending on @tr-j hands-on testing.

### DIFF
--- a/modules/Reports/reporting_cycles_manage_editProcess.php
+++ b/modules/Reports/reporting_cycles_manage_editProcess.php
@@ -45,7 +45,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Reports/reporting_cycles_m
         'cycleNumber'           => $_POST['cycleNumber'] ?? '1',
         'cycleTotal'            => $_POST['cycleTotal'] ?? '1',
         'notes'                 => $_POST['notes'] ?? '',
-        'milestones'            => $_POST['milestones'] ?? '',
+        'milestones'            => $_POST['milestones'] ?? [],
     ];
 
     $data['dateStart'] = Format::dateConvert($data['dateStart']);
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Well caught, thanks! 👍

PR #1380, opened 2021-07-05
Fix PDOException catches in module

yookoala:
**Description**
* Need to add backslash to the front of root classname in
  namespace.

**Motivation and Context**
* Found some `PDOException` call without proper escaping. Could result in unhandled error.

**How Has This Been Tested?**
* Locally.


SKuipers (approved):
Looks good 👍

PR #1381, opened 2021-07-05
Remove uses of obsoleted getUserPhoto

yookoala:
**Description**
* Replace with Format::userPhoto() calls.

**Motivation and Context**
* The function `getUserPhoto` was marked obsoleted in v16 but found some trace of it.

**How Has This Been Tested?**
* Locally


SKuipers (approved):
Thanks! 👍

PR #1374, opened 2021-06-29
Library: Cost field

powertashton:
**Description**
This PR introduces a cost field to the library module, separate from the replacement cost field.

**Motivation and Context**
There are cases where the original cost of an item and it's replacement cost may differ, and schools may wish to retain both pieces of information for the purpose of record keeping/accounting

**How Has This Been Tested?**
Locally, using pre-existing data from a separate system.

**Notes**
This is part of a bigger series of changes to the Library module, some future features may include a "Details" page from the library_manage_catalog.php table (separate from the "Edit" page, so those with permission may view extended details of an item without being able to edit it) as well as the ability to go into an auditing/stocktaking mode where it is expected that all records are scanned and updated where relevant.

PR #1382, opened 2021-07-12
Form Inputs: InstaScan QR code reading && Library: QR code reading implementation

powertashton:
**Description**
This PR builds on top of #1374 and introduces further changes to the Library module. This time introducing a new option for the text field input that allows QR code scanning.

**Motivation and Context**
Whilst it is possible to use external devices or extensions to scan and input a QR code into a text field, it is not necessarily convenient or accessible, this builds the functionality directly into gibbon using schmich/instascan which utilises webrtc and should work on most modern browsers on both mobile and desktop, provided a camera is attached to the device.

In a future PR involving a stocktaking/auditing extension to the library module an instascan enabled read-only field would also likely suffice as a means to ensure staff are only able to scan QR codes to fill the field and not bypass procedures by manually inputing values.

**How Has This Been Tested?**
This has been tested on Firefox, Chrome, and Microsoft Edge on both Mac OS Mojave and Windows 10, as well as on Chrome and Edge on Android (Oxygen OS) 10 whilst running on servers with PHP 7.3 and 8.0

**Screenshots**
image
ezgif-2-b57d1a6f7665


powertashton:
Only took a week of troubleshooting and over 20 commits to fix the failing test... Turns out it wasn't my code that was messed up after all but the librarytype migration code somehow slipped past all the other tests for everyone else...

SKuipers (changes requested):
Hi Ash, I've merged in everything so now we can see your changes isolated. It's looking good. Before we merge, since the scanner is a pretty specialized field type, rather than adding the logic to every single text field, it'd be cleaner to create a new class called Scanner which extends the TextField class. This way, the code is organized and compartmentalized. It'll require creating a new class and adding an entry in the FormFactory, but if there's anywhere else in the core we can use the QR scanner, it'll be more robust. Thanks!

powertashton:
Smart idea, I think I've just pushed the changes to do just that. :)

src/Forms/Input/Scanner.php
```diff
@@ -0,0 +1,191 @@
...
+namespace Gibbon\Forms\Input;
+use Gibbon\Forms\Element;
+
+/**
+ * Scanner
+ *
+ * @version v23
+ * @since   v23
+ */
+class Scanner extends Input
```
SKuipers, line 29: If this class were to extend TextField rather than Input, then you don't need to duplicate the existing methods, you can just add the new ones that you need. Keeps the code DRY :)

modules/Library/report_catalogSummary_view.php
```diff
@@ -0,0 +1,44 @@
...
+    // Access denied
+    $page->addError(__('You do not have access to this action.'));
+} else {
+    //Proceed!
+    //Check if ItemID specified
+    $gibbonLibraryItemID = $_GET['gibbonLibraryItemID'];
+    if ($gibbonLibraryItemID == '') {
+        $page->addError(__('You have not specified one or more required parameters.'));
+    } else {
+
```
SKuipers, line 40: Is this page meant to do something? 😉

SKuipers (changes requested):
Nice, thanks for the quick response (lightspeed compared to my response time!). Almost there.

modules/Library/report_catalogSummary_view.php
powertashton: You saw nothing.

src/Forms/Input/Scanner.php
powertashton: Ah yes, I /think/ I get it? Working with classes is not my forte.

SKuipers (approved):
Looking good, I'm going to merge and test 👍 Thanks for the quick changes!

SKuipers:
Tests out perfectly, and I bet there might be areas in the future where this could be useful. I've made [a small tweak](https://github.com/GibbonEdu/core/commit/f6dae5a01300ef1dcb947ca62ab58fc5a4c4cf0c) to the icons so that it's purpose is more apparent: 

Screenshot 2021-09-13 at 1 24 42 PM

PR #1383, opened 2021-07-17
Draft: Url class for formatting all Gibbon internal URLs

yookoala:
**Description**
Provides [PSR-7](https://www.php-fig.org/psr/psr-7/) `Psr\Http\Message\UriInterface` compatible implementation of Url class for Gibbon internal URLs.

Note: This is still in the draft phase. Please let me know what you think. Will still need to fine tune the way things work before this is ready for merge.

**Motivation and Context**
* More descriptive declaration of all Gibbon URLs.
* Improve compatibility with other PSR compliant libraries.
* Future-proof. Make it easier to tweak the URL structure in the future (e.g. routing paradigm change).

**How Has This Been Tested?**
* CI platform

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ themes/*
 # Tests Folder
 tests/log/*
 tests/config/support/_generated/*
+tests/.phpunit.result.cache
 
 # Template Cache
 /resources/templates/cache/*
--- a/functions.php
+++ b/functions.php
@@ -21,6 +21,7 @@ use Gibbon\Forms\Form;
 use Gibbon\Services\Format;
 use Gibbon\Contracts\Comms\Mailer;
 use Gibbon\Domain\System\LogGateway;
+use Gibbon\Url;
 
 require_once dirname(__FILE__).'/gibbon.php';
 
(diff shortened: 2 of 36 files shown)


SKuipers (commented):
I really like this, it would help bring us closer to the aim of implementing proper routing, and cleans up the use of strings everywhere for URLs. Making it PSR-7 compliant is a huge plus. It looks much more readable and helps us move towards using arrays for URL params, which are easier for validate.

I wonder if there is a better namespace to put it in. Initially `src/Gibbon` was a folder of legacy classes, and not very descriptive (everthing in src is Gibbon-related, otherwise it'd be in vendor), ideally I'd like to relocate (or deprecate) all the classes in `src/Gibbon`. I sometimes find myself looking at the [Laravel/Illuminate](https://github.com/laravel/framework/tree/8.x/src/Illuminate) folder structure for ideas, so I wonder if a `Gibbon/Http` namespace/folder would be best for this class.

yookoala:
Since the PSR-7 `UriInterface` is in the `\Psr\Http\Message` namespace, I do think this is appropriate.

I'll do the rename.

yookoala:
@SKuipers This is quite ready for use already, and some of it is conflicting with the session refactor (`Url::fromRoute` can replace the use `$_SESSION['guid']['absoluteURL']` or `$session->get('absoluteURL')` in URL building). So I'll have to redo the refactor when more session refactor works are merged.

Do you think it needs any change (e.g. API rename) before merge? Or can we merge this sooner?

PR #1384, opened 2021-07-17
Remove gibbon.php include from functions.php

yookoala:
**Description**
* With the statement, functions.php can never be included by itself.
  And gibbon.php might not work within the context where functions.php
  could be useful.

* The lib/paypal/expresscheckout.php includes functions.php, which
  in turn includes gibbon.php (which includes functions.php). This is
  purely redundent. Changed to include gibbon.php directly.

**Motivation and Context**
Another piece of my work needs to include the `functions.php` file. But since it includes `gibbon.php`, there are logics that stops things from working in my situation. There is no point blocking the include of `functions.php`. It should be dry and without unnecessary dependencies.

I've only found 1 direct include of `functions.php` in the code base in `lib/paypal/expresscheckout.php`. I figure it might need `gibbon.php` environment to work, so I changed it to directly include `functions.php`.

**How Has This Been Tested?**
CI platform

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -22,8 +22,6 @@ use Gibbon\Services\Format;
 use Gibbon\Contracts\Comms\Mailer;
 use Gibbon\Domain\System\LogGateway;
 
-require_once dirname(__FILE__).'/gibbon.php';
-
 function getIPAddress() {
     $return = false;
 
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
Well caught. This was here for when the gibbon.php file was initially introduced, it's nice to see that we've refactored enough code that this line is no longer needed 👍

PR #1386, opened 2021-07-18
Core: fix phpdoc for database connection select()

yookoala:
**Description**
* Fix phpdocs for better IDE support.

**Motivation and Context**
* The original return was ambiguously "mixed", which was not very helpful describing what the select method would return.
* Especially annoying when I was working on rewriting some obsoleted `executeQuery()` calls to `select()`. VSCode complains that it does not know any method of the select result.

**How Has This Been Tested?**
* Locally on my VSCode.

**Screenshots**
cropped-3
cropped-2
cropped-1

### DIFF
--- a/src/Contracts/Database/Connection.php
+++ b/src/Contracts/Database/Connection.php
@@ -20,7 +20,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 namespace Gibbon\Contracts\Database;
 
 /**
- * Database Connection Interface 
+ * Database Connection Interface
  * Borrowed in part from Illuminate\Database\ConnectionInterface
  *
  * @version	v16
@@ -49,7 +49,8 @@ interface Connection
      *
      * @param  string  $query
      * @param  array   $bindings
-     * @return object
+     *
+     * @return \Gibbon\Database\Result Result of the database query.
      */
     public function select($query, $bindings = []);
 
@@ -121,8 +122,8 @@ interface Connection
 
     /**
      * @deprecated
-     * Backwards compatability for the old Gibbon\sqlConnection class. 
-     * Replaced with more expressive method names. Also because the 
+     * Backwards compatability for the old Gibbon\sqlConnection class.
+     * Replaced with more expressive method names. Also because the
      * parameters are backwards. Hoping to phase this one out in v17.
      *
      * @param  string  $query
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -100,11 +100,7 @@ class Connection implements ConnectionInterface
     }
 
     /**
-     * Run a select statement against the database.
-     *
-     * @param  string  $query
-     * @param  array   $bindings
-     * @return object
+     * {@inheritDoc}
      */
     public function select($query, $bindings = [])
     {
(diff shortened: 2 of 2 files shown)


src/Contracts/Database/Connection.php
```diff
@@ -49,7 +49,8 @@ public function selectOne($query, $bindings = []);
      *
      * @param  string  $query
      * @param  array   $bindings
-     * @return object
+     *
```
SKuipers, line 52: The connection class uses an extended Gibbon\Database\Result class, rather than the base PDOStatement class. It internally handles checking for false, and returns an empty Result object on failure, to avoid areas in the code constantly having to check for false values. So there shouldn't be cases where these methods are returning false 🤔

yookoala: Thanks for pointing this out. Updated the phpdocs. Please check.

src/Contracts/Database/Connection.php
```diff
@@ -49,7 +49,8 @@ public function selectOne($query, $bindings = []);
...
+     * @return \PDOStatement|\Gibbon\Database\Result The PDO statement, or an empty Result if query failed.
```
SKuipers, line 53: Thanks for the quick changes, as you can see I've carved some time out of my day to work on PRs 😄 

This may be more clear, as it will return a Result object on success too.

```suggestion
     * @return \Gibbon\Database\Result The Result object, or an empty Result if query failed.
```

yookoala: If I read the current implementation right, the `\Gibbon\Database\Connection::select` method is using `\Gibbon\Database\Connection::run` underneath.

https://github.com/GibbonEdu/core/blob/2b21872b3fafbc73c253cf254028dfb3e3d06a5c/src/Database/Connection.php#L178-L198

1. The method would return `$this->result`.
2. If there is no database error, `$this->result` would be receiving the return value of `$this->pdo->prepare()`.
3. Object `$this->pdo` here is of the \PDO type. The prepare function will therefore returns `\PDOStatement`. Only when there is a PDO Exception, the method would assign `$this->result` the `new Result()` for return.

SKuipers: Yes, you read correct. The difference is in the wiring in the MySqlConnector class, which uses the line `$connection->setAttribute(\PDO::ATTR_STATEMENT_CLASS, array(Result::class));` [here](https://github.com/GibbonEdu/core/blob/v23.0.00/src/Database/MySqlConnector.php#L96) to set the return class to a Result, which is an extended version of the base PDOStatement. This Result class enables us to add some more readable methods, such as `fetchGroupedUnique()` to replace `->fetchAll(PDO::FETCH_GROUP|PDO::FETCH_UNIQUE)`, and to ensure that a `fetchAll()` call will always return an array, rather than null or false, to make iterating over results less cumbersome.

yookoala: This should probably belongs to another PR, but the current code arrangement seems strange. The `Connection` does not depends on `MySqlConnector`. `MySqlConnector` uses `Connection`. So `Connection` should not assume that `PDO` object is initialized by `MySqlConnector::connect`. So I think its structurally better to have `configureEncoding` and `setModes` in the `Connection` class instead.

yookoala: Documentation updated. Please check. Thanks.

SKuipers: Hmm. The purpose of the MySqlConnector connector class is to produce a fully configured Connection. The Connection contract is then moderately database agnostic, so that the configuration is separate from the function of the Connection, and ideally opening the possibility of future support for database types.

SKuipers: In any case, happy to merge this PR, it's looking good 😄

PR #1387, opened 2021-07-18
Refactor Session constructor

yookoala:
**Description**
* Extract container and environment / side-effect dependent logics into Session::create.
* Be specific about the variables that Session needed in its constructor.
* Changed the `CoreServiceProvider` to adapt it.

**Motivation and Context**
* `Session::__construct` was heavily depending on the environment and specific container setup.
* By separating the container dependency, now the Session is dry and easy to mock. It's much easier to debug, too.

**How Has This Been Tested?**
In CI environment.

### DIFF
--- a/src/Gibbon/Session.php
+++ b/src/Gibbon/Session.php
@@ -26,7 +26,7 @@ use Psr\Container\ContainerInterface;
 /**
  * Session Class
  *
- * @version	v13
+ * @version	v23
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
This is looking good, I like the logic of making the Session easier to mock and test. Thanks for these improvements 👍

PR #1399, opened 2021-08-18
A massive refactor to the installer

yookoala:
**Description**

* Refactor the monolithic installer script into separated component in the `\Gibbon\Install` namespace. Includes:
  * HttpInstallController
  * Context
  * Config
  * Installer
  * NonceService

* Design follows, as much as sensible, the single responsibility principle:
  * **HttpInstallController**
    Knows nothing about anything. It handles the user input. Then, according to context and installer functions return values or exception, formulate some HTML views for users.
  * **Context**
    Does nothing but provide information about the server. Includes PHP version, extension installed, the install os path and config path.
  * **Config**
    Simply an abstraction of config data. It does not know where to save a config file.
  * **Installer**
    Only does installation steps. Knows nothing.
  * **NonceService**
    Only does nonce create and verification.

* After refactoring, the **HttpInstallController** would takes what it learn from **Context** and user input, to do installation steps with **Installer**, then format the result or the next step UI until installation is done.

* Each step now submit to itself instead of submitting to the next one. If there is any Exception thrown, simply show message on the same page. Retains all data entry of the step after failure. User do not need to start from beginning when one install step failed.

* There are too many pieces involved. It's really hard to do small, incremental refactor. Hence a huge pile of code changes in one PR.

**Motivation and Context**
* Got sick of having to reset everything if filling something wrong in any of the installation steps.
* The code was a huge piece of monolithic code that is hard to read or debug.
* Pave way to create unattained installation (e.g. CLI installer) that can reuse the same code. And if that went well, make it possible to dockerize Gibbon. So in general, make it easier to install Gibbon on different platform / setup.

**How Has This Been Tested?**
* Locally,
* CI Environment.


SKuipers:
Wow Koala, this sounds amazing 😃 Thank you for your incredible work putting this together! We're just in the midst of school year startup here, but I look forward to checking out this PR and the other pending ones soon. Thank you!!

SKuipers:
This looks like a huge step forward, I'm just reviewing it and testing it out. When I load the branch locally and trigger the installer, I see a "Oh no" fatal error page, which seems to be caused by the Context class receiving a string from the `apache_get_version` function, and the constructor expecting an array:

Screenshot 2021-09-20 at 11 58 12 AM

`PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Gibbon\Install\Context::__construct() must be of the type array or null, string given, called in ****/gibbon/src/Install/Context.php on line 57 and defined in ****/gibbon/src/Install/Context.php:44`

yookoala:
Hmm...

Will fix this this week (hopefully today).

yookoala:
My bad. Should have used [apache_get_modules](https://www.php.net/manual/en/function.apache-get-modules.php) there.

Fixed and rebased.

SKuipers:
Thanks! My first install test went well. A couple things I noticed:

- The installer should create a database if it does not exist (which was the previous behaviour). I received the following error. Incidentally, it was nice to see that the error didn't reset the form and lose my data 😁 
  Screenshot 2021-09-20 at 12 17 56 PM

- I received an error that said "Installation cannot proceed. The required field "title" is not set.", however title isn't a required field. Probably and easy fix, again, nice that my data didn't disappear. 

- A couple variables were missing from the post-install message.
  Screenshot 2021-09-20 at 12 23 59 PM
 
Overall looking great, and very slick in the install.php page, I'm going to take a closer look at the code to familiarize myself with it, but I'm really liking the direction of this, and it's a massive improvement on the original script 👍

SKuipers:
Hmm, a couple more kinks to work out yet. Looks like the absolutePath and absoluteURL did not populate in the database or in the session, so the login form is broken, and links/images are also broken:
Screenshot 2021-09-20 at 12 27 08 PM

After clicking to login:
Screenshot 2021-09-20 at 12 29 21 PM

yookoala:
Will figure these out hopefully this week.

SKuipers:
Thanks, no rush! 😄

yookoala:
What stack are you using? Apache + PHP7.4?

SKuipers:
Apache 2.4.46 (via MAMP) + PHP 7.3.24 + MySQL 5.7.32

yookoala:
php-fpm + mod_proxy_fcgi or mod_php?

SKuipers:
php-fpm yes, no mod_php (but it's not a requirement for Gibbon).

Screenshot 2021-09-20 at 13-07-57 DEMO - Gibbon - System Admin

resources/templates/index.twig.html
```diff
@@ -18,14 +18,14 @@
...
+
         <div class="px-4 sm:px-6 lg:px-12 pb-24">
-        
+
             {% block header %}
                 <div id="header" class="relative flex justify-between items-center">
 
                     <a id="header-logo" class="block my-4 max-w-xs sm:max-w-full leading-none" href="{{ absoluteURL }}">
-                        <img class="block max-w-full {{ isLoggedIn ? 'h-12' : 'h-20 mt-4 mb-4' }}" alt="{{ organisationName }} Logo" src="{{ absoluteURL }}/{{ organisationLogo|default("/themes/Default/img/logo.png") }}" style="max-height:100px;" />
+                        <img class="block max-w-full {{ isLoggedIn ? 'h-12' : 'h-20 mt-4 mb-4' }}" alt="{{ organisationName }} Logo" src="{{ absoluteURL }}{{ organisationLogo|default("/themes/Default/img/logo.png") }}" style="max-height:100px;" />
```
SKuipers, line 28: The absoluteURL and absolutePath are always saved without trailing slashes, so this would break existing installs. Perhaps double-check that the refactored installer is also `rtrim`ing trailing slashes, for consistency.

yookoala: Although the absoluteURL and absolutePath are stored without trailing slashes, the logo path here has a leading slash. So there is no need for an additional trailing slash to the absoluteURL.

SKuipers: Hmm, shoot, that must be a bug then from when the field was changed from text to a file uploader 😬 Existing installs do not have a leading slash on this setting

Screenshot 2021-09-30 at 2 50 49 PM

yookoala: Nevermind. Modified again to remove any leading slash from the logo path.

yookoala:
Spent some time to rewrite some part of this. Should have addressed all the issue mentioned above.

yookoala:
At this point, things are pretty testable. I can further improve the test coverage.

SKuipers (approved):
Koala, this is looking amazing. The code is clear and easy to read, and testable, which is awesome. I've done two install tests, with and without demo data, and both worked well. I fixed one small missing variable in the final Step 4 success message, which wasn't redirecting back to the root level index page. Aside from that, it tested out perfectly, and all previous issues appear to have been resolved. I'm happy to merge this into v23 👍 🎉

SKuipers:
In testing the GibbonEdu services post-merge, the statistics collection appears to be working, but the support request email is not working. I suspect this is because `registerGibbonSupport` is being treated like a system setting, but it is not actually in the gibbonSetting table, it only exists in the installer form. What are your thoughts on the best way to handle this? I don't want to just start randomly hacking around in this nice clean code 😄

SKuipers:
Thanks! I think perhaps we just add `registerGibbonSupport` as a setting, I could see where it would be useful to have an endpoint to update/unsubscribe people when the system administrator email changes, so it'd be good to know about what this setting was at install time. This would require no additional wiring in the installer, from what I can see.

rossdotparker:
The new code really is very elegant!

yookoala:
Sure. Will take a look. Thanks for reporting :-)

zcraber:
> * Pave way to create unattained installation (e.g. CLI installer)

Hi,
Is CLI installer still on the works?

PR #1401, opened 2021-08-19
Core: remove Gibbon\Exception

yookoala:
**Description**
* Remove `Gibbon\Exception`.
* Replace its only usage in `Gibbon/Core` with `\Exception` instead.

**Motivation and Context**
* The class is not used in any meaningful way. Its only method `errorHandlerCallback` was not used anywhere in the code base (core, modules, contrib modules). Should striped down for simplicity.

**How Has This Been Tested?**
CI Environment.


SKuipers (approved):
Thanks! These were part of some vestigial functionality that are nice to see cleaned up 👍

PR #1404, opened 2021-08-25
Messenger: null coalesce refactoring in messenger_postProcess

hmerrettICHK:
**How Has This Been Tested?**

**Screenshots**

### DIFF
--- a/modules/Messenger/messenger_postProcess.php
+++ b/modules/Messenger/messenger_postProcess.php
@@ -51,7 +51,9 @@ else {
         $_POST = $validator->sanitize($_POST, ['body' => 'HTML']);
 
 		//Validate Inputs
-		$email=$_POST["email"] ?? 'N';
+
+    $email=$_POST["email"] ?? 'N';
+    $email = ($email != 'Y') ? 'N' : $email;
 		if ($email=="Y") {
 			$from=$_POST["from"] ?? '';
 		}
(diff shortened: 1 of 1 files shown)


modules/Messenger/messenger_postProcess.php
```diff
@@ -51,47 +51,22 @@
         $_POST = $validator->sanitize($_POST, ['body' => 'HTML']);
 
 		//Validate Inputs
-		$email=$_POST["email"] ;
-		if ($email!="Y") {
-			$email="N" ;
-		}
+		$email=$_POST["email"] ?? 'N';
```
yookoala, line 54: The original code would set `$email` to `'N'` if it is anything other than `'Y'`. The new code would only set `$email` to `'N'` if it is not set. This could introduce new logical errors.

yookoala (changes requested):

modules/Messenger/messenger_postProcess.php
```diff
@@ -232,7 +204,7 @@
 			//Role Categories
 			if (isActionAccessible($guid, $connection2, "/modules/Messenger/messenger_post.php", "New Message_role")) {
 				if ($_POST["roleCategory"]=="Y") {
```
SKuipers, line 206: It looks like you're adding null coalesce to areas that are assigning values based on POST variables, which is certainly the most common ways POST variables are used. However, we can see that $_POST["roleCategory"] is being accessed directly, which will also cause an error if this value is not assigned. These are best not overlooked. 

One way to handle this is to create a variable with the null coalesce first, if it does not already exist, then check the value of the variable.

```suggestion
				$roleCategory = $_POST["roleCategory"] ?? 'N';
				if ($roleCategory == "Y") {
```

Can you look through the code for areas like this that you could improve? Thanks!

SKuipers (changes requested):
I'm noticing a lot of $_POST variables still missing null coalesce. I know this is a big file, but be sure to use a find tool to look through it and check each $_POST variable, so we can catch them all and not have to come back and re-visit this file.

PR #1405, opened 2021-08-25
System Admin: enable creating multiple different email templates per type

SKuipers:
The email template feature isn't widely utilized (yet!), however to improve it's usability we discovered it'd be nice to be able to create and select specific templates for different tasks. In this case, we needed one email template for sending report cards home, one for sending transcripts. In hindsight, this feature does work best if the user can select which template to use, rather than having a single template.

This PR:
- Adds the ability to duplicate existing email templates, which tags them as "Custom"
- Adds the ability to delete custom email templates (but not the built in ones, so there's always at least one of each type)
- Adds a drop down to the Reports > Send Reports bulk action, which is the only area that uses Email Templates
- Adds table fields as needed to gibbonEmailTemplate

**Motivation and Context**
Improving the feature before using it more widely.

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2021-08-25 at 2 19 35 PM

### DIFF
--- a/modules/System Admin/emailTemplates_manage_duplicate.php
+++ b/modules/System Admin/emailTemplates_manage_duplicate.php
@@ -26,14 +26,18 @@ use Gibbon\Forms\DatabaseFormFactory;
 use Gibbon\Domain\System\SettingGateway;
 use Gibbon\Domain\System\EmailTemplateGateway;
 
-if (isActionAccessible($guid, $connection2, '/modules/System Admin/emailTemplates_manage_edit.php') == false) {
+if (isActionAccessible($guid, $connection2, '/modules/System Admin/emailTemplates_manage_duplicate.php') == false) {
     // Access denied
     $page->addError(__('You do not have access to this action.'));
 } else {
     // Proceed!
     $page->breadcrumbs
         ->add(__('Email Templates'), 'emailTemplates_manage.php')
-        ->add(__('Edit Email Template'));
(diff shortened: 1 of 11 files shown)


modules/System Admin/emailTemplates_manage_duplicate.php
```diff
@@ -0,0 +1,80 @@
...
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+use Gibbon\Forms\Form;
+use Gibbon\Tables\Action;
+use Gibbon\Services\Format;
+use Gibbon\Comms\EmailTemplate;
+use Gibbon\Domain\User\UserGateway;
```
rossdotparker, line 24: This gateway does not seem to be used in this file...unless I'm missing something.

modules/System Admin/emailTemplates_manage_duplicate.php
```diff
@@ -0,0 +1,80 @@
...
+use Gibbon\Forms\DatabaseFormFactory;
+use Gibbon\Domain\System\SettingGateway;
```
rossdotparker, line 26: And neither does this one.

rossdotparker:
Looks solid to me. The only comments are about some superfluous gateways being used, which can be fixed post-merge if you agree with my assessment. I think the same might be true of the Edit Template interface too.

PR #1408, opened 2021-08-29
System Admin: implement Omnipay library, enable Stripe payment gateway

SKuipers:
This PR adds the [PHP League's Omnipay](https://omnipay.thephpleague.com/) gateway, which has the capability to support many different types of payment gateways. All functionality that involves payments has now been decoupled from the previous PayPal scripts, and use a new Payment class, which interfaces with Omnipay. 

Because payment gateways still require some configuration and adaptation of the values they return, even with Omnipay, this PR doesn't automatically add all the gateways that Omnipay supports (which are many and exist as separate libraries), but it does nicely decouple things and enable the addition of new gateways in the future.

The new Payment class is the single point of contact for all process scripts that handle payments. It bridges the Omnipay functionality with Gibbon by handling the gibbonPayment table and returning Gibbon-specific error messages. Adding new gateway support is a matter of updating the Payment class to handle those gateways.

**Description**
- Adds the Omnipay common library via Composer
- Adds initial Omnipay PayPal and Omnipay Stripe gateways
- Adds a Payment class to interface with the database & provide error messages
- Adds a PaymentInterface class to define the public contract for payment methods
- Changes all paypalAPI settings to paymentAPI settings
- Adds an option to select which payment gateway to use
- Updates strings that previously refer to PayPal
- Adds a test payment feature (as well as a handy test email feature)
- Removes now-unused paypal scripts

**Motivation and Context**
Enables the future support of more payment gateways, making the invoice and finances features of Gibbon more widely usable.

**How Has This Been Tested?**
Locally with developer accounts for PayPal and Stripe.

**Screenshots**
Screenshot 2021-08-29 at 12 38 47 PM

Screenshot 2021-08-29 at 12 38 59 PM

Adds a test feature for SMTP email as well:
Screenshot 2021-08-29 at 12 39 23 PM


SKuipers:
Switched to draft, I noticed there's some more general refactoring that'd be nice to catch on the payment pages.

CHANGEDB.php
```diff
@@ -1011,4 +1011,12 @@
 UPDATE `gibbonAction` SET `URLList` = 'emailTemplates_manage.php,emailTemplates_manage_duplicate.php,emailTemplates_manage_edit.php,emailTemplates_manage_delete.php' WHERE `name`='Email Templates' AND `gibbonModuleID`=(SELECT gibbonModuleID FROM gibbonModule WHERE name='System Admin');end
 ALTER TABLE `gibbonEmailTemplate` DROP INDEX `templateName`, ADD UNIQUE `moduleTemplate` (`templateName`, `moduleName`) USING BTREE;end
 ALTER TABLE `gibbonEmailTemplate` ADD `type` ENUM('Core','Additional','Custom') NOT NULL DEFAULT 'Core' AFTER `gibbonEmailTemplateID`;end
+UPDATE `gibbonSetting` SET name='paymentAPIUsername', nameDisplay='API Username', description='API details are provided by the payment gateway provider' WHERE scope='System' AND name='paypalAPIUsername';end
```
rossdotparker, line 1014: I wonder if the human-reading name should be Payment API Username, of if this is overkill?

modules/Finance/invoices_payOnline.php
```diff
@@ -76,22 +73,17 @@
...
                     if ($financeOnlinePaymentEnabled == 'Y') {
                         echo "<h3 style='margin-top: 40px'>";
                         echo __('Online Payment');
                         echo '</h3>';
                         echo '<p>';
                         if ($financeOnlinePaymentThreshold == '' or $financeOnlinePaymentThreshold >= $feeTotal) {
-                            echo sprintf(__('Payment can be made by credit card, using our secure PayPal payment gateway. When you press Pay Online Now, you will be directed to PayPal in order to make payment. During this process we do not see or store your credit card details. Once the transaction is complete you will be returned to %1$s.'), $session->get('systemName')).' ';
+                            echo sprintf(__('Payment can be made by credit card, using our secure %2$s payment gateway. When you press Pay Online Now, you will be directed to %2$s in order to make payment. During this process we do not see or store your credit card details. Once the transaction is complete you will be returned to %1$s.'), $session->get('systemName'), $paymentGateway).' ';
```
rossdotparker, line 86: Have such strings been updated in POEditor?

modules/Finance/moduleFunctions.php
```diff
@@ -1196,15 +1194,15 @@ function receiptContents($guid, $connection2, $gibbonFinanceInvoiceID, $gibbonSc
...
+                    $sqlPayment = "SELECT gibbonPayment.*, surname, preferredName FROM gibbonPayment LEFT JOIN gibbonPerson ON (gibbonPayment.gibbonPersonID=gibbonPerson.gibbonPersonID) WHERE foreignTable=:foreignTable AND foreignTableID=:foreignTableID LIMIT $receiptNumber, 1";
                     $resultPayment = $connection2->prepare($sqlPayment);
                     $resultPayment->execute($dataPayment);
                 } catch (PDOException $e) {
                     $paymentFail = true;
                     $return .= "<div class='error'>".$e->getMessage().'</div>';
                 }
 
-                if ($resultPayment->rowCount() != 1) {
+                if ($resultPayment->rowCount() <= 0) {
```
rossdotparker, line 1205: I've traditionally used !=1 because multiple records should not be returned, but if they are for some reason, an error seems better than grabbing the first record. Any thoughts?

rossdotparker:
This is looking good. Other than the three comments, I'm happy to merge when you feel that it is ready to go. Thanks!

CHANGEDB.php
SKuipers: Maybe a bit overkill, since it's in a section called Payment API already 🤔

modules/Finance/invoices_payOnline.php
SKuipers: They have not. I looked at doing this manually, but since the strings are old enough, there's a decent number of translations already in there, and for any translations using different character sets or rtl, it's not just a matter of removing/replacing the word PayPay. We may just have to re-translate these strings 😬

modules/Finance/moduleFunctions.php
SKuipers: Okay, there's already a LIMIT 1 on the query, so in this case there shouldn't be a possibility of multiple records. I'll restore the original code.

SKuipers:
Okay, this one is looking ready, thanks for the review Ross. I'm going to merge it in and do a fair amount of post-merge testing. Possibly even update the CHANGELOG to headline this one.

Pl1997:
Hello @SKuipers,

I'm currently looking into Gibbon for a music school I'm working with. We are heavily dependent on our payment provider HelloAsso, which is a quite unique french system specifically designed for associations (no fees for the association nor the payer, their business model relies on tips). They just released a new api implementing a system which is very similar to Stripe Checkout ( https://api.helloasso.com/v5/swagger/ui/index#/ , https://github.com/HelloAsso/checkout-sample ). 

I'm currently trying to understand if your work on Omnipay could make it easy enough for us to implement it in Gibbon. My main concern is this : 

In their [documentation](https://drive.google.com/file/d/1oFciovXkKiMJ7EIDcNYh5nVTq5NLPIEI/view), they state that after each completed purchase, a notification is sent to an endpoint which must be provided by each association. As an alternate way, they do provide a way to retrieve the payment details with a GET request, but they also warn that this method might not work immediately, and that it should only be used at least a few minutes after processing the payment.

My understanding of how Stripe Checkout currently works in Omnipay (and thus, in Gibbon) is that it only relies on such requests (or on the post-payment return url, but this doesn't seem robust enough by itself, since any internet or browser failure on the client side might interrupt the process). I can't find any clue of some endpoint being implemented to handle incoming notifications from payment providers.

Would you be kind enough to confirm that I'm not misunderstanding anything ? And before I eventually try to give it a shot, could you share your feelings about the feasibility of implementing this payment provider, given the constraints that I described ?

Thanks a lot for your help !

PR #1409, opened 2021-08-30
Fix Composer PSR-4 library warnings

### LINKED ISSUE #1392: Composer warning
I am sure this is already known to our illustrious developers, but just in case it is not, I will refer to it here:

Package container-interop/container-interop is abandoned, you should avoid using it. Use psr/container instead.
Package phpoffice/phpexcel is abandoned, you should avoid using it. Use phpoffice/phpspreadsheet instead.

yookoala:
**Description**
* Remove "phpoffice/phpexcel" dependency.
* Upgrade "slim/slim" to "^4.0"

**Motivation and Context**
* Fix #1392.

**How Has This Been Tested?**
* Locally

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -45,7 +45,7 @@
         "aura/sqlquery": "3.*-dev",
         "tecnickcom/tcpdf": "6.0.038",
         "twig/twig": "^2.0",
-        "slim/slim": "^3.0",
+        "slim/slim": "^4.0",
         "phpmailer/phpmailer": "^6.5.0",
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
Thanks for the fix. We're not using Slim yet, but maybe we keep it for now. It was part of my initial grand plans to introduce proper controllers/middleware/routing, but I had to go back and start with the basics to get a lot more of the system refactored first. We're getting closer though! 😄

PR #1410, opened 2021-08-30
System Admin: fix missing Custom Imports Folder setting

### LINKED ISSUE #1391: Import from File not detecting custom YMLs
**Describe the bug**
I wanted to set up a custom YML for importing some student application data. When I uploaded the custom yml file to both the resources/imports directory and also the uploads/imports directory, Gibbon did not detect it nor list it on the Import from File page.

**To Reproduce**

Using Gibbon 22.0, place a custom YML import file in the resources/imports directory or the uploads/imports directory
Go to System Admin / Import from File
Custom Import is not listed

**Expected behavior**
Custom YML import file should be listed at the top of the page

**Your System**

Gibbon version [e.g. v22.0.00]
PHP 7.3 

**Additional context**
After digging through the code, it seems that Gibbon is still looking for the old Data Admin/importcustomfolderlocation setting from the database to know where to look for custom import yml files. You can see this by looking at the function in the ImportType.php file. This setting no longer exists in the database because the Data Admin was moved to the core with v22.

After adding the Data Admin/importcustomfolderlocation setting back to the gibbonSetting table in the database, my custom YML file was recognized just fine and showed up properly on the Import from File page.

Perhaps this needs to be a setting that we can add through the GUI again or perhaps it needs to be hard coded to look in uploads/imports for custom YML files.

SKuipers:
The importCustomFolderLocation setting was previously attached to Data Admin, but after moving the functionality in to the core, the core ImportType.php file was still referencesing the Data Admin setting. This prevented anyone without Data Admin installed from changing the setting. I've fixed ImportType.php and added the setting as a small form at the top of the Import from File page.

**Motivation and Context**
Fixes #1391 

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2021-08-30 at 1 56 04 PM

### DIFF
--- a/CHANGEDB.php
+++ b/CHANGEDB.php
@@ -1011,4 +1011,5 @@ UPDATE `gibbonEmailTemplate` SET `templateName`=`templateType` WHERE `templateNa
 UPDATE `gibbonAction` SET `URLList` = 'emailTemplates_manage.php,emailTemplates_manage_duplicate.php,emailTemplates_manage_edit.php,emailTemplates_manage_delete.php' WHERE `name`='Email Templates' AND `gibbonModuleID`=(SELECT gibbonModuleID FROM gibbonModule WHERE name='System Admin');end
 ALTER TABLE `gibbonEmailTemplate` DROP INDEX `templateName`, ADD UNIQUE `moduleTemplate` (`templateName`, `moduleName`) USING BTREE;end
 ALTER TABLE `gibbonEmailTemplate` ADD `type` ENUM('Core','Additional','Custom') NOT NULL DEFAULT 'Core' AFTER `gibbonEmailTemplateID`;end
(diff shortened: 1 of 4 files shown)


rossdotparker:
So this setting supplements /resources/uploads so that you can keep your own imports in a separate location. Technically it looks good, just keen to get a little more background on it.

SKuipers:
Yes, this setting has always existed and custom uploads have always been loaded through /uploads/imports by default, rather than adding them to the resources folder, which is more likely to be overwritten. The only change is that the setting was still attached to the Data Admin module, and if someone hadn't ever used Data Admin, the setting would be missing and therefore they couldn't use custom imports.

rossdotparker:
Brilliant, thanks for the explanation. I'll now merge.

PR #1419, opened 2021-09-10
Refactor database Connection and MySqlConnector

yookoala:
**Description**
* Migrate `MySqlConnector::configureEncoding` to `Connection::configure`.
* Migrate `MySqlConnector::setModes` to `Connection::setModes`.
* Migrate the usage of the above methods into `Connnection::__construct`.

**Motivation and Context**
* Previously, the class `Connection` depends on the convention to use to  `MySqlConnector` the `PDO` object.
* `Connection` do not directly use `MySqlConnector` in anyway. The dependency is not obvious and can cause unexpected behaviour in the future development.
* Migrating the PDO configuration into `Connection` improves that while still preserve the current behaviour.

**How Has This Been Tested?**
* Locally.
* CI Environment.

### DIFF
--- a/src/Database/Connection.php
+++ b/src/Database/Connection.php
@@ -66,11 +66,22 @@ class Connection implements ConnectionInterface
 
     /**
      * Create the connection wrapper around a \PDO instance.
-     * @param \PDO $pdo
+     *
+     * @param \PDO  $pdo     Should be configured to fit Connection use internally.
      * @param array $config
+     *
+     * @throws \InvalidArgumentException Throws if the \PDO is misconfigured.
      */
     public function __construct(PDO $pdo, array $config = [])
     {
+        // Check the statement class. Expect $pdo to be correctly setup.
+        // See MySqlConnector::configureEncoding().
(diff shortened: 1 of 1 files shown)


SKuipers:
Hmmm. I see where you're going with this one, but I'm not sure about the utility value of the change. The purpose of the MySqlConnector class was for separation of concerns, this class creates and returns a fully configures PDO connection, and then the Connection class uses that connection. I'm not sure about the benefit of moving the configuration of PDO over into Connection, as that mixes the two concerns.

yookoala:
@SKuipers: Perhaps just migrate the PDO attribute setup to Connection and leaving the setModes in MySqlConnector?

PDO setup has nothing to do specifically with MySQL. But they are essential to fulfilling the contract documented in Connection methods (e.g. select).

yookoala:
@SKuipers: Changed the approach to check $pdo in `Connection::__construct` to make sure things work as expected.

SKuipers (approved):
Thanks, this seems like a decent compromise 👍

PR #1421, opened 2021-09-10
System: fix InvalidArgumentException use in class

yookoala:
**Description**
* fix \InvalidArgumentException reference in Gibbon\Data\UsernameGenerator. The root class reference was without the "\" prefix.

**Motivation and Context**
* Found this bug in the middle of other project.

**How Has This Been Tested?**
* Locally in VSCode.


SKuipers (approved):
👍

PR #1422, opened 2021-09-10
Composer: add phpstan test

yookoala:
**Description**
* Add [phpstan](https://phpstan.org/) as development dependency.
* Add `composer test:phpstan` script target.
* Did **NOT** add this to default test target. So will **NOT** be run in CI for now.

**Motivation and Context**
* PHPStan scan and catches PHP issues for the corner cases that user would rarely run into. Have a test like this can improve our overall robustness.
* Will gradually fix all the potential errors found by PHPStan in the code base. When all errors are fixed, can add this into regular test target.

**How Has This Been Tested?**
* Locally with the `composer test:phpstan` command.

**Screenshots**
2021-09-10 17-52-29 的螢幕擷圖

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,9 @@
         "test:phpunit": [
             "cd tests && ../vendor/bin/phpunit --verbose --colors=always --configuration phpunit.xml"
         ],
+        "test:phpstan": [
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
Static analysis, nice! I would imagine there's lots of hits on that, since there's still oodles of unrefactored code, but nice to start getting these things into the pipeline. Thanks Koala!

PR #1423, opened 2021-09-10
System: fix Exception reference in MailDriver

yookoala:
**Description**
* DriverNotConfiguredException is from SMS library namespace. Added
  proper use statement to fix it.

**Motivation and Context**
* Fix issue found by PHPStan

**How Has This Been Tested?**
* Locally with PHPStan


SKuipers (approved):
👍

PR #1424, opened 2021-09-10
System: fix typo in ImportType class

yookoala:
**Description**
* Class name is case sensitive but mistyped with small letter.

**Motivation and Context**
* Fix issue found by PHPStan

**How Has This Been Tested?**
* Locally with PHPStan


SKuipers (approved):
👍

PR #1425, opened 2021-09-10
System: rewrite usage of obsoleted Connection::executeQuery

yookoala:
**Description**
* Replace all usage of obsoleted Connection::executeQuery to
  Connection::select in the core.

**Motivation and Context**
* Connection::executeQuery is obsoleted since v16. It's usages should be rewritten with Connection::select.

**How Has This Been Tested?**
* Locally.
* CI Environment.


yookoala:
Not ready yet. Close for now.

PR #1426, opened 2021-09-10
System: rewrite usage of obsoleted Connection::executeQuery in core

yookoala:
**Description**
* Replace all usage of obsoleted Connection::executeQuery to
  Connection::select in the core.
* Update FileUploaderTest to mock select instead of
  executeQuery.

**Motivation and Context**
* Connection::executeQuery is obsoleted since v16. It's usages should be rewritten with Connection::select.

**How Has This Been Tested?**
* Locally.
* CI Environment.

### DIFF
--- a/src/Data/ImportType.php
+++ b/src/Data/ImportType.php
@@ -333,7 +333,7 @@ class ImportType
     {
         try {
             $sql="SHOW COLUMNS FROM " . $this->getDetail('table');
-            $result = $pdo->executeQuery([], $sql);
+            $result = $pdo->select($sql);
         } catch (\PDOException $e) {
             return false;
         }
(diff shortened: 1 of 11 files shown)


SKuipers (commented):
Looking good. The try/catch blocks can also be removed, as they are handled internally by the connection class, and were part of the old style of querying.

yookoala:
> Looking good. The try/catch blocks can also be removed, as they are handled internally by the connection class, and were part of the old style of querying.

The try-catch blocks seems only exists in the ImportType class. Removed them all in the new commit.

SKuipers (approved):
Looking good, so lovely to have those old deprecated methods cleaned up 😄 thanks!

PR #1431, opened 2021-09-14
Refactor Locale test

yookoala:
**Description**
* Mock global $gibbon and $guid locally in the test class and restore original after the test.
* Simplify Locale shortcut test to test only how it uses `$gibbon->locale->translate` and
  `$gibbon->locale->translateN`. Not the actual result.

**Motivation and Context**
* Remove the dependency on bootstraped global $gibbon for Locale related tests.
* Simplify tests.

**How Has This Been Tested?**
* Locally.
* CI environment.


SKuipers (approved):
Looks like some useful improvements to clean up the dependencies and simplify things, thanks! 👍

PR #1433, opened 2021-09-15
composer: remove stripe/omnipay in vendor folder

yookoala:
**Description**
* ~~Remove accidentally commited vendor folder files.~~
* Remove commited vendor folder files.
* Update composer.json to use dev-master of the "omnipay/stripe" library.

**Motivation and Context**
* Vendor folder files should not be committed, or they will be overriden / removed by composer operations.

**How Has This Been Tested?**
* CI environment


SKuipers:
Hi Koala, these haven't been accidentally committed. There was a change necessary to support the checkout API in Stripe, which [was PRd and merged](https://github.com/thephpleague/omnipay-stripe/pull/202) into the omnipay/stripe repository, but hasn't been released yet so I needed to add it into our system manually, at least until they properly release the next version.

yookoala:
@SKuipers: Understood. But the current approach to include dev-master stuff would be broken after a simple `composer install` or `composer update`. I think a more robust approach would be to use "dev-master" of "omnipay/stripe". This way another composer operation won't break the vendor folder.

Alternatively, we can add fork repository to composer.json (like what we did in #688). But this is only needed if the commit was not committed to the official repo.

SKuipers (approved):
Thanks Koala, this solution is much better than my hack-ish fix 👍

PR #1434, opened 2021-09-15
composer: use latest tcpdf in packagist

yookoala:
**Description**
Remove the unofficial fork from the composer setup.

**Motivation and Context**
* Originally, the old fork was used to keep using older tcpdf version without having to fetch so many tags with GitHub API (has API limit issue). After #1209, updating vendor dependencies became much easier. Seems no longer have issue on GitHub API front after upgrade.
* The tcpdf used was so old. Should be upgraded for various reason.

**How Has This Been Tested?**
* CI environment.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,7 +46,7 @@
         "google/apiclient": "^2.2.0",
         "league/container": "^3.3.3",
         "aura/sqlquery": "3.*-dev",
-        "tecnickcom/tcpdf": "6.0.038",
+        "tecnickcom/tcpdf": "^6.4",
         "twig/twig": "^2.0",
         "slim/slim": "^4.0",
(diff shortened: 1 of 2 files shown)


SKuipers:
Looks good, happy to try this solution out, just need to update the composer.lock to resolve conflicts. What is the best command to do this? I've been using `composer update --no-dev --lock` and it seems to be working, but wanted to check.

yookoala:
@SKuipers: updated the lock file. Should be ready for merge.

SKuipers (approved):
Looks good, I'm going to merge this one in and test it out 👍

SKuipers:
Looking good, I've done some tests and the only breaking change was the addTTFfont was changed into a static method on the TCPDF_FONTS class, but this only affected an additional module, so I'll make that update.

PR #1436, opened 2021-09-15
System: correct phpdocs in Gibbon\Excel::exportWithQuery

yookoala:
**Description**
* Change the `@return` of Excel::exportWithQuery from `string` to `void`.

**Motivation and Context**
* Gibbon\Excel::exportWithQuery does not return anything. The
  string result is directly output to STDOUT stream of
  cgi / fastcgi server. The phpdocs should reflect that.

**How Has This Been Tested?**
* Locally on VSCode.
* Locally with PHPStan.


SKuipers (approved):
👍

PR #1440, opened 2021-09-16
composer: use php-parsecsv in packagist

yookoala:
**Description**
* Remove the local copy of old `ParseCSV` class.
* Replace all usages with the official release of `ParseCsv\Csv`.
* Depends on #1434. (Simply wanted a smoother composer.lock merge. Can be changed if needed.)

**Motivation and Context**
* Better support from upstream updates (e.g. bugfixes, feature update).
* The original library should have fixed the issue when the copied was made and changed.
* The old class file has some phpdocs mistake (found by PHPStan). They were fixed in the latest official version. Kind of pointless to fix the local copy while we can simply use the latest official one.

**How Has This Been Tested?**
* CI platform.
* Locally with PHPStan.
* **Note**: Have **NOT** test import CSV with the new code. Need some hands-on test to make sure it is functional.


SKuipers (approved):
Nice to tidy this up, thanks! 👍

PR #1441, opened 2021-09-20
System: fix typo in NotificationEvent

yookoala:
**Description**
* Class name session should be Session.

**Motivation and Context**
* Fix error found by PHPStan

**How Has This Been Tested?**
* Locally with PHPStan.


SKuipers (approved):
Aha well caught, thanks PHPStan 😄

PR #1442, opened 2021-09-20
System: remove DataRetentionGateway::getAllTables

yookoala:
**Description**
* Remove the method \Gibbon\Domain\System\DataRetentionGateway::getAllTables.

**Motivation and Context**
* The method seems to be obsoleted artifact in the development
  process. It is not used anywhere in the code base. Also the
  property it is accessing was not set at all. Should be removed.

**How Has This Been Tested?**
* Locally.
* CI environment.


SKuipers (approved):
👍

PR #1443, opened 2021-09-20
System: remove LoggerFactory::settingGateway

yookoala:
**Description**
* Remove private property $settingGateway from LoggerFactory.

**Motivation and Context**
* It is not used anywhere in the class.

**How Has This Been Tested?**
* Locally with VSCode.
* CI environment.


SKuipers (approved):
👍

PR #1444, opened 2021-09-20
System: improve phpdocs for SettingGateway

yookoala:
**Description**
* improve phpdocs of the methods and properties in SettingGateway.

**Motivation and Context**
* improve developer experience in IDE.

**How Has This Been Tested?**
* Locally with VSCode.

**Screenshots**
screenshot

### DIFF
--- a/src/Domain/System/SettingGateway.php
+++ b/src/Domain/System/SettingGateway.php
@@ -20,7 +20,6 @@ along with this program. If not, see <http://www.gnu.org/licenses/>.
 namespace Gibbon\Domain\System;
 
 use Gibbon\Domain\Traits\TableAware;
-use Gibbon\Domain\QueryCriteria;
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Thanks! The tableName, primaryKey and searchableColumns are repeated in every single gateway that uses the TableAware trait, so maybe overkill to add comments for each one? Especially since they're private variables. Or is this just for IDE completion?

yookoala:
@SKuipers: Still exploring the best way to document the static variables in TableAware. Will definitely revisit this.

yookoala:
Unfortunately, static variables cannot be declared as abstract. And overriding protected variable would result in error like this:
```
SettingGateway and TableAware define the same property ($tableName) in the composition of SettingGateway. However, the definition differs and is considered incompatible.
```

So I don't yet have a better way to document static variables referenced by TableAware trait.

PR #1448, opened 2021-09-23
Test: phpstan to use config file

yookoala:
**Description**
* Use phpstan.neon to config the behaviour instead of using the
  CLI arguments. Much cleaner and more readable.
* Define how to format a .neon file in .editorconfig to ensure
  consistancy.

**Motivation and Context**
* The old way misses all moduleFunctions.php when scanning for functions. Adding all those files to command argument would be a huge pain in the ass. Using a config file is more sensible.

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -13,3 +13,7 @@ indent_size = 4
 
 [*.yml]
 indent_size = 2
+
+[*.neon]
+indent_style = tab
+indent_size = 4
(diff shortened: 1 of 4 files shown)


SKuipers (commented):
We have a lot of configuration-style files that are starting to clutter the root folder, I wonder if we could put the phpstan.neon file in a folder, such a /tests, to keep it out of the root, since it's part of the CI/test suite.

yookoala:
This should be do-able, but most projects leave these in the root.

SKuipers:
Hmm, I wonder if we try and get the release build process to clean these types of configuration files up, because I've been noticing there's more and more of them in the root. It's likely not good practice to have them in production servers, because many robots out there are often scraping for standard configuration files.

yookoala:
Sure. I think we can do 2 things in parallel:
1. Add .htaccess to forbid access to those files.
2. Update the release package building workflow to exclude those files.

SKuipers (approved):
Looking good 👍

PR #1449, opened 2021-09-23
Improve phpdocs and declare missing property to classes

yookoala:
**Description**
* Property document the properties in the classes.
* Add some undeclared (but used) properties as protected properties.
* Remove unused properties.
* Add abstract method to TableAware trait.

**Motivation and Context**
* Improve developer experience with IDE.

**How Has This Been Tested?**
* Locally with VSCode.
* CI environment.


SKuipers (approved):
How did we live without static analysis 😂  The addition of abstract methods in the TableAware trait sounds like a good addition.

PR #1450, opened 2021-09-23
System: fix BackgroundProcessor sprintf issue

yookoala:
**Description**
* The background command sprintf has more supplied arguments than
  placeholder. Should remove unnecessary ones.
* Use "> /dev/null" for Windows background command.

**Motivation and Context**
* The original sprintf statement has error.
* Windows CMD equivalent of "> /dev/null" would be "> NUL".

**How Has This Been Tested?**
* Locally with PHPStan.


SKuipers (approved):
I don't have a windows setup to test with, but sounds logical 👍

PR #1451, opened 2021-09-23
System: fix Select::getLabelContext

yookoala:
**Description**
* getLabelContext is documented to return string. It should return
  at least an empty string even if the attribute 'multiple' is not
  true.

**Motivation and Context**
* Fix issue found by PHPStan.

**How Has This Been Tested?**
* Locally with PHPStan.
* CI Environment.


SKuipers (approved):
👍

PR #1452, opened 2021-09-23
System: fix Payment::requestPayment

yookoala:
**Description**
* requestPayment is documented to return string. It should return
  a string in any and all cases.
* improve phpdocs for IDE support.
* fixed a potential bug with the IDE type support.

**Motivation and Context**
* Fix issue found by PHPStan.
* Fix bug.
* Improve developer experience on IDE.

**How Has This Been Tested?**
* Locally with PHPStan.
* CI environment.

### DIFF
--- a/src/Services/Payment/Payment.php
+++ b/src/Services/Payment/Payment.php
@@ -164,7 +164,13 @@ class Payment implements PaymentInterface
 
         // Send purchase request to the payment gateway
         $options = $this->getPaymentRequestOptions($amount, $reason);
-        $response = $this->omnipay->purchase($options)->setCurrency($this->currency)->send();
+
+        /**
+         * @var OmnipayGateway $gateway
+         */
+        $gateway = $this->omnipay->purchase($options);
+        $transaction = $gateway->setCurrency($this->currency)->purchase();
+        $response = $transaction->send();
 
         if ($response->isSuccessful()) {
             // Payment request was successful, continue redirect
--- a/src/Services/Payment/Payment.php
+++ b/src/Services/Payment/Payment.php
@@ -139,18 +139,19 @@ class Payment implements PaymentInterface
             // Payment request was successful, continue redirect
             $responseData = $response->getData();
             header("Location: " . $responseData['url']);
+            return self::RETURN_SUCCESS;
 
         } elseif ($response->isRedirect()) {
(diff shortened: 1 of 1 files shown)


src/Services/Payment/Payment.php
```diff
@@ -119,18 +119,19 @@ public function requestPayment($amount, $reason = 'Purchase') : string
             // Payment request was successful, continue redirect
             $responseData = $response->getData();
             header("Location: " . $responseData['url']);
+            return self::RETURN_SUCCESS;
 
         } elseif ($response->isRedirect()) {
             // Redirect to offsite payment gateway
             $response->redirect();
+            return ''; // TODO: should there be a return value for "redirecting"?
```
SKuipers, line 127: Perhaps a warning, since the only way to see this would be if the redirect failed, which could indicate an interruption in the process. Maybe `const RETURN_REDIRECT = 'warning4';` with a message of `__('Your online payment redirect was interrupted before it could be completed. Please try again and if the error persists, contact the school.')`

```suggestion
            return 'self::RETURN_REDIRECT';
```

yookoala:
I added the message accordingly.

But I have doubt. From my brief read to Omnipay documentation, the redirect might not be an error, but could be an intermediate state. Some payment gateways (e.g. PayPal) requires merchant site to redirect payers to them when doing payment.

src/Services/Payment/Payment.php
yookoala: I added the message accordingly.

But I have doubt. From my brief read to Omnipay documentation, the redirect might not be an error, but could be an intermediate state. Some payment gateways (e.g. PayPal) requires merchant site to redirect payers to them when doing payment. So the message here might not match what is actually happening.

SKuipers (approved):
Looking good, thanks for making these updates and adding the new warning4 message 👍

SKuipers:
Hi Koala, we discovered that the payment process had stopped working, and it appears to be related to returning a value from the requestPayment method when performing a redirect. After removing the redirect the payments are working again, so I've reverted a few of the changes from this PR in the commit here: https://github.com/GibbonEdu/core/commit/3c28454edcf58786eaab87dd92a1b7696d00eaff

yookoala:
This is strange...

PR #1454, opened 2021-09-29
System: implement SessionHandler classes and update Session namespace

SKuipers:
Session handling is currently only available through naive PHP session storage. For servers running through clustered web hosting with shared storage, this file-based session handling causes issues. Many modern PHP applications support database sessions, so this PR makes the initial changes to support this functionality. Some of the code is transitional, and will be further built upon in two upcoming PRs, one which adds an interface for the gibbonSession table, and one which implements OAuth2 and updates the login/authentication process.

New SessionFactory class:
- Moves the Session class into it's own namespace, fixes any areas of the codebase that were inadvertently referencing the concrete Session class rather than the contract interface.
- Moves session creation into a SessionFactory class and moves methods which populate session data into the same class. This leaves the Session class with a single responsibility of reading and writing session data.

New SessionHandler classes:
- Implements the native SessionHandlerInterface through two new classes, NativeSessionHandler and DatabaseSessionHandler.
- Adds a gibbonSession table, which is used by the DatabaseSessionHandler. In a follow-up PR, this table will have it's own Gateway class, and will also be used by the NativeSessionHandler for basic who's-logged-in information.
- Enables encrypting session data using openssl AES 256, if available and a key is provided. This adds an extra later of protection especially for database sessions.
- The SessionFactory currently determines the type of session and encryption based on config.php settings. These will later be added as an option to the newly refactored installer.

Updated Core and service providers:
- The gibbon.php script needed to be re-wired so that the database is created before the session, so that the session could access it.
- Session dependencies in the Locale class were removed, so that it could be created before the session.
- Dependencies in the MarkbookView and AttendanceView classes were simplified.

This PR accidentally got a bit larger than expected, but many of the individual file changes are small, and a certain amount of re-wiring was necessary.

**Motivation and Context**
Enable a wider range of session handling options, and tidy up the Session class.

**How Has This Been Tested?**
Locally with native and database sessions.


yookoala:
I think there is one more potential refactor here. User data can be an object in session (instead of having its property values set to the session), for example:

```php
// instead of
doSomething($session->get('username'));

// make it
doSomething($session->get('user')->username);
```

If we have a user class in the future, we can potentially do this:

```php
if ($session->get('user')->isAllowed('access all teaching plans')) {
  // ...
}
```

But its best to leave this for another time.

SKuipers:
> I think there is one more potential refactor here. User data can be an object in session (instead of having its property values set to the session), for example:

I agree, ideally we'll move away from isActionAccessible and similar functions to using a User object to check these things. Once we have some routing in place, the isActionAccessible check could even be a middleware step tied to the authentication/authorization of a user's access to that route, so that it doesn't even need repeated in every script.

rossdotparker:
This looks good to me. I've added one item for review, but it is refactoring related, and so I will merge this now.

rossdotparker:
Merged and tested well. Looking forward to the next PRs, which I guess include at some point the changes to the installer and config.php (for new instances) to allow database session selection. Thanks for pushing this one forward : )

PR #1455, opened 2021-09-30
System: replace use of $gibbon->session with $session in non-module files

SKuipers:
A quick follow-up PR to #1454 that replaces $gibbon->session with $session in non-module files.


yookoala (approved):

yookoala:
@SKuipers: A little side-track suggestion. I think we should have a "refactor" tag in GitHub. There are many refactor works in this project. And they are usually neither "feature" nor "bug".

SKuipers:
> @SKuipers: A little side-track suggestion. I think we should have a "refactor" tag in GitHub. There are many refactor works in this project. And they are usually neither "new feature" nor "bug".

Sounds good, I've added a new label 👍

PR #1456, opened 2021-09-30
Remote CLI

rossdotparker:
**Description**
Allows administrators to set a key to enable CLI scripts to be run via HTTP/S. Off by default.

**Motivation and Context**
Sometimes it is not possible to run CLI scripts locally due to hosting constraints. This allows them to be run remotely with a reason level of security.

**How Has This Been Tested?**
Locally


rossdotparker:
Hi @SKuipers, I considered using a gateway to do this, but it did not really save much, and seemed to add more complexity. An advantage would be to have the logic in a single file, but I was just not sure it was worth it. Feel free to tinker if you have a better way to architect it.

SKuipers (approved):
Looking good. Gateway wise, we could use the SettingGateway here, but it's something we can start changing system-wide, and it's an easy change, so I've added it to the potential Hacktoberfest list 👍

PR #1469, opened 2021-10-13
System: Fix form.twig.html compatibility with elements that are not a layout Element

yookoala:
**Description**
* ~~Fix DataTable's use as form element but without proper getClass method~~.
* Modify the form template to be compatible with form element that are not `Gibbon\Forms\Layout\Element` (i.e. do not provide a `getClass` method.)

**Motivation and Context**
* DataTable is used as Form element in Gibbon (e.g. [here](https://github.com/GibbonEdu/core/blob/d2eae14a420930e4b25fa0c15dc766eea346e626/modules/Activities/activities_manage.php#L116)).
* There are places in the form template (e.g. [here](https://github.com/GibbonEdu/core/blob/d2eae14a420930e4b25fa0c15dc766eea346e626/resources/templates/components/formTable.twig.html#L32) and [here](https://github.com/GibbonEdu/core/blob/d2eae14a420930e4b25fa0c15dc766eea346e626/resources/templates/components/formTable.twig.html#L43)) that expects elements to have a `getClass` method and it returns `string`. Failing to comply would cause cryptic deprecated message in PHP 8.1:
   ```
   Deprecated: strtr(): Passing null to parameter #1 ($string) of type string is deprecated in /home/runner/work/GibbonEduCore/GibbonEduCore/vendor/twig/twig/src/Extension/CoreExtension.php on line 519
   ```
* The deprecated message will become official warning or error in future version of PHP.
* ~~If DataTable is expected to be form element, having the trait actually enhances its feature on that front.~~

**How Has This Been Tested?**
* Locally.
* CI environment with PHP 8.1

### DIFF
--- a/resources/templates/components/form.twig.html
+++ b/resources/templates/components/form.twig.html
@@ -19,12 +19,12 @@ For info about editing, see: https://twig.symfony.com/doc/2.x/
     {% if form.getTitle %}
         <h2>{{ form.getTitle }}</h2>
     {% endif %}
-    
+
     {% if form.getDescription %}
         <p>{{ form.getDescription|raw }}</p>
     {% endif %}
 
-    
+
     {% block header %}
         <header class="relative">
             {% if form.getHeader %}
@@ -50,7 +50,7 @@ For info about editing, see: https://twig.symfony.com/doc/2.x/
             {% set sectionLoop = loop %}
 
             {% for num, row in rows %}
-            
+
                 {% set rowLoop = loop %}
                 {% if true or renderStyle == 'flex' %}
                     {% set rowClass = 'flex flex-col sm:flex-row justify-between sm:items-center content-center p-0' %}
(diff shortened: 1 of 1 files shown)


SKuipers:
Ah, hmm, that must of stuck in without realizing it 🧐  Ideally, the DataTable should not have a BasicAttributesTrait, as it is not an HTML-renderable element in and of itself. It has view methods that are separated into Renderer classes, which have the potential for having HTML attributes, but not guaranteed (eg: spreadsheet renderable). It'd be nice to not add this trait to DataTable, and to keep the data and view methods separate. 

In this case, it looks like the template may need to be improved to handle this in a better way, since ideally the only method it relies on directly is the getOutput from the OutputtableInterface. Since almost every Form element extends from the base Element class, we may be able to put an `element.instanceOf('Gibbon\\Forms\\Layout\\Element')` check before calling the getClass method.

PR #1473, opened 2021-10-15
System: improve session timeout, add Active Sessions page and Maintenance Mode

SKuipers:
This PR builds upon the session changes in #1454

General improvements to classes in the Session namespace:
- Creates a SessionGateway class and migrates session-based SQL into that class.
- Updates the DatabaseSessionHandler to use the SessionGateway.
- Updates the NativeSessionHandler to garbage collect the database sessions.
- Updates all sessions to record basic information in the gibbonSession table, including native sessions.

Additional interfaces:
- Added an Active Sessions page to System Admin, allowing admin to see who is online and for how long.
- Added a Maintenance Mode, which when turned on will log out all non-admin users and display a maintenance message.
- Maintenance mode is only available to admin users, and prevents non-admin users from logging in.
- The maintenance mode message is displayed on all pages when active.
- Adds a separate logout message when users have been manually logged out using maintenance mode.

Session timeout:
- Since sessions are now tracked via the database, the session timeout no longer needs to be client-side. This solves the issue of multiple tabs timing out when the session is still in use.
- The index_notification_ajax.php now checks for timed out sessions, and can both display session warnings as well as remove them across multiple tabs.
- Removed the old javascript-based session timeout and it's associated libraries.

**How Has This Been Tested?**
Locally

**Screenshots**

Default maintenance mode message:
Screenshot 2021-10-15 at 2 41 13 PM

Maintenance mode, currently in the Active Sessions page. Should it be moved elsewhere?
Screenshot 2021-10-15 at 2 41 04 PM
 
Active sessions:
Screenshot 2021-10-15 at 3 17 20 PM

### DIFF
--- a/src/Session/NativeSessionHandler.php
+++ b/src/Session/NativeSessionHandler.php
@@ -104,6 +104,7 @@ class NativeSessionHandler extends SessionHandler implements SessionHandlerInter
      * @param int $max_lifetime
      * @return bool true for success or false for failure
      */
+    #[\ReturnTypeWillChange]
     public function gc($max_lifetime)
     {
         $this->sessionGateway->deleteExpiredSessions($max_lifetime);


rossdotparker:
This looks great, and I know that Tieku will be happy as he mentioned a maintenance mode some time ago. I am going to merge now and then test. Amazed at how you've brought this together so quickly : D

rossdotparker:
Tests out perfectly as far as I can see : ) Oddly, in logging in as a parent I found a bug with some of my recent $_SESSION refactoring. I don't know how I missed that in testing. I've just pushed a fix for it.

PR #1484, opened 2021-10-22
System: change gibbonSession table creation

yookoala:
**Description**
* Add gibbonSession table schema to gibbon.sql

**Motivation and Context**
* Fix "Base table or view not found" warning message:
  2021-10-22 22-35-21 的螢幕擷圖

**How Has This Been Tested?**
* CI Environment


SKuipers:
This is bizarre, as the changes seem unrelated to the errors it's throwing 🤔

SKuipers (approved):
I'm going to merge this one and see what happens 🤞

SKuipers:
It looks like the CHANGEDB is not being run at all for the installer test, as now that we have a database dump in the artifacts, I can spot check some of the changes from v23 and I'm not seeing them show up in the dumped database 🤔

yookoala:
Probably something went wrong after the edge version test. Will need to check it out.

yookoala:
The updater has been run. But the database was not updated. Need more time to check why:
2021-11-01 13-27-36 的螢幕擷圖

yookoala:
The alter table add column failed because gibbon.sql already has the column correctly setup.
2021-11-01 14-07-52 的螢幕擷圖

yookoala:
Unfortunately, not all MySQL / MariaDB version support `ALTER TABLE ADD COLUMN IF NOT EXISTS`. So it's not safe to do it that way. And there is currently no reliable way to detect column existence before adding without some stored procedure.

So the fix I can think of is either:
1. Change the `gibbon.sql` to remove the columns that needed to be added in `CHANGEDB.php`; or
2. Change the updater to know which line it can safely skip (?).

yookoala:
See #1492.

I removed some of the older `CHANGEDB.php` lines and added 2 new lines to it. The new code will simply remove the existing session table and re-create one with the latest definition. This should fix the add column issue.

PR #1490, opened 2021-10-29
System: fix Url class not returning an absolute URL

SKuipers:
We were running into bugs related to the fact that Gibbon relies on using absoluteURL in many areas, but the new Url class was returning a relative URL. These affected the breadcrumbs, fast finder, timetable, password reset, etc. 

For example:

- Breadcrumbs with an absoluteURL of `http://localhost:8888/gibbon/` were ending up as `http://localhost:8888/gibbon/gibbon/index.php?q=/modules/System Admin/systemOverview.php` because the url was relative, so browsers were prepending the current url, but the path has also been parsed and added to the url route path.
- The timetable was not working because the ajax load script was expecting an absolute path and was receiving a relative one.
- Preferences links in notification emails were not working because the NotificationSender was putting a relative link in the href attribute.

This PR updates the Url class to return a fully resolved absolute URL. It appears the base Uri class is designed to parse a url into it's component parts when passed into the constructor, so the Url class no longer needs to parse the url when the setBaseUrl method is called. Instead, it stores the baseUrl and passes it into the constructor when using the withRoute, withModuleRoute, etc. methods.

**Motivation and Context**
Fixes bugs that have occurred after #1383

**How Has This Been Tested?**
Locally

### DIFF
--- a/index.php
+++ b/index.php
@@ -52,8 +52,8 @@ $settingGateway = $container->get(SettingGateway::class);
  * MODULE BREADCRUMBS
  */
 if ($isLoggedIn && $module = $page->getModule()) {
-    $page->breadcrumbs->setBaseURL('index.php?q=/modules/'.$module->name.'/');
-    $page->breadcrumbs->add($module->type == 'Core' ? __($module->name) : __m($module->name), trim($module->entryURL, '/'));
+    $page->breadcrumbs->setBaseURL(Url::fromModuleRoute($module->name));
+    $page->breadcrumbs->add($module->type == 'Core' ? __($module->name) : __m($module->name), $module->entryURL);
 }
 
 /**
--- a/indexFindRedirect.php
+++ b/indexFindRedirect.php
@@ -12,7 +12,7 @@ if ($session->has('absoluteURL')) {
     if ($type == 'Stu') {
         $URL = Url::fromModuleRoute('Students', 'student_view_details')->withQueryParam('gibbonPersonID', $id);
     } elseif ($type == 'Act') {
-        $URL = Url::fromModuleRoute($id);
+        $URL = Url::fromModuleRoute(strstr($id, '/', true), trim(strstr($id, '/'), '/ '));
     } elseif ($type == 'Sta') {
         $URL = Url::fromModuleRoute('Staff', 'staff_view_details')->withQueryParam('gibbonPersonID', $id);
     } elseif ($type == 'Cla') {
--- a/passwordResetProcess.php
+++ b/passwordResetProcess.php
@@ -122,7 +122,7 @@ else {
                 'title'  => __('Password Reset'),
                 'body'   => nl2br(trim($body, "\n")),
                 'button' => [
-                    'url'  => Url::fromRoute('passwordReset')->withQueryParams([
+                    'url'  => Url::fromAbsoluteRoute('passwordReset')->withQueryParams([
                         'input' => $input,
                         'step' => 2,
                         'gibbonPersonResetID' => $gibbonPersonResetID,
--- a/passwordResetProcess.php
+++ b/passwordResetProcess.php
@@ -128,6 +128,7 @@ else {
                         'gibbonPersonResetID' => $gibbonPersonResetID,
                         'key' => $key,
                     ]),
+                    'external' => true,
                     'text' => __('Click Here'),
                 ],
             ]);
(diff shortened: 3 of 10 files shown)


yookoala:
Thanks! Most of the changes looks reasonable to me. But I'd prefer to keep the URLs as [root-relative form](https://mor10.com/html-basics-hyperlink-syntax-absolute-relative-and-root-relative/) instead of absolute. HTMLs with root-relative hyperlink are more flexible to different server setups. For instance, if a school admin changed the domain name to a Gibbon instance, he can still navigate it normally before changing the "absoluteURL" setting. This leave more room in deploying the software.

Wordpress forces user to use the domain name and path it was registered. This, in the past, caused some pain in my ass when doing HTTPS migrations and domain changes.

yookoala (commented):
Thanks! Most of the changes looks reasonable to me. But I'd prefer to keep the URLs as [root-relative form](https://mor10.com/html-basics-hyperlink-syntax-absolute-relative-and-root-relative/) instead of absolute. HTMLs with root-relative hyperlink are more flexible to different server setups. For instance, if a school admin changed the domain name to a Gibbon instance, he can still navigate it normally before changing the "absoluteURL" setting. This leave more room in deploying the software.

Wordpress forces user to use the domain name and path it was registered. This, in the past, caused some pain in my ass when doing HTTPS migrations and domain changes.

SKuipers:
Okay, we can give this a shot, I'll look at updating my PR. I think we will need to add a method to use the absolute path for cases where the script is not running from the root, such as any of the process scripts in any of the modules.

SKuipers:
Okay, I've updated the PR to retain the root-relative URLs. It looks like there were cases were the absoluteURL was being prepended before the root-relative url, which resulted in duplicate paths for any installation not at the root (eg: `http://localhost:8888/gibbon` + `/gibbon/index.php?q=/modules/System Admin/systemOverview.php`). Going forward, we will have to be sure to keep an eye on  not using absoluteURL in relation to the Url class. 

I've added a fromAbsoluteRoute() method, since there are cases where we cannot rely on root-relative links, such as for emails sent out from the system.

I did notice that root-relative links don't appear to work in the src attribute for images, so we cannot use the same technique to remove the absoluteURL from these (eg: images in templates).

yookoala:
Can we list those cases (e.g. pages to check) somehow? It's hard to really study the issue without reproducible cases.

SKuipers:
I've done some more tests and it looks like there are less cases than I had thought (that I've noticed so-far). It looks like Url::fromModuleRoute() will work fine inside module folders, which is great. I think the only two cases I have come across so-far are:
- External urls included in emails and notifications need to be absolute
- Urls used inside templates, such as href and src attributes, cant use absoluteURL combined with a Url class string, as that's where the duplication of paths comes from. One option is for us to add a reference to the url class in the template globals, that way templates can use the same methods to form urls.

yookoala:
For emails and notifications, I think we can have additional method that force the URL to be absolute.

For templates, I think it's best to rewrite all `{{ absoluteURL ~ url }}` to `{{ url }}` or other combination of url with absoltueURL.

SKuipers:
Okay, sounds good. I've added the fromAbsoluteRoute() method & test, and applied it to the email case. I think we can look at updating the templates in a future PR, this PR includes some fixes that would be nice to get in sooner.

yookoala:
I see how `fromAbsoluteRoute()` would work to solve the issue here, but I think another approach would be more flexible and versatle in long run. I think we can have a public method `setAbsolute(bool $flag = true): self` can be added. Then the URLs needed to be absolute can be created by chain-call like this:

```php
Url::fromHandlerRoute('passwordReset')
  ->setAbsolute()
  ->withQueryParams([
      'input' => $input,
      'step' => 2,
      'gibbonPersonResetID' => $gibbonPersonResetID,
      'key' => $key,
  ]);
```

This way, all other internal URL can potentially be converted to absolute url, which would be very useful for email notifications or off-site messages.

The Url class has access to the internal baseUrl variable. The absolute url logic can be implemented in `__toString()` so things can still be changed the last minute.

SKuipers:
I've updated the method and applied it to the password reset. I went with 'withAbsoluteUrl()' to match the naming pattern of the other uri methods that modify the rendered url.

yookoala (approved):

yookoala:
Thanks for taking my opinions :-)

PR #1491, opened 2021-11-01
System: removed deprecated functions no longer used in core and additional modules

rossdotparker:
**Description**
Since v17 quite a few functions.php functions have been deprecated, and often entirely removed from the code base. This PR removes roughly 50% of these functions, including refactoring those final usages that do exist.

**Motivation and Context**
Ongoing refactoring, and a desire to reduce the size of functions.php

**How Has This Been Tested?**
Locally

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -26,7 +26,10 @@ v23.0.00
         System: added ability to run CLI scripts remotely, with secure key
         System: removed deprecated functions no longer used in core and additional modules
             - addressFormat
+            - dateConvert
+            - dateConvertBack
             - dateConvertToTimestamp
+            - formatName
             - formatPhone
             - getAge
             - getGibbonMailer
@@ -34,6 +37,7 @@ v23.0.00
             - getSettingByScope
             - getUserPhoto
             - printUserPhoto
+            - returnProcess
             - setLog
             - tinymceStyleStripTags
             - ynExpander
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -31,6 +31,8 @@ v23.0.00
             - getAge
             - getGibbonMailer
             - getModuleEntry
+            - getSettingByScope
+            - getUserPhoto
             - printUserPhoto
             - setLog
             - tinymceStyleStripTags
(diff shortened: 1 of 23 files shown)


yookoala:
The CI issue will probably be fixed by #1492

functions.php
```diff
@@ -512,47 +477,6 @@ function getWeekNumber($date, $connection2, $guid)
     }
 }
 
-/**
- * @deprecated in v23.
- */
-function getModuleEntry($address, $connection2, $guid)
```
yookoala, line 518: It seems that `getModuleEntry()` is still used in the latest version of [Reporting](https://gibbonedu.org/extend/#reporting) (v2.25). Do we wait for the module to update before removing the function?

SKuipers: Reporting is a somewhat out of date module. The module developer will still have until v23 is released to update, should they choose, otherwise the current module is still (likely) compatible with v22 and previous versions.

rossdotparker: That's correct! I'll alert Andy Statham, the author, to this need.

CHANGELOG.txt
```diff
@@ -24,6 +24,23 @@ v23.0.00
 
     Significant Changes
         System: added ability to run CLI scripts remotely, with secure key
+        System: removed deprecated functions no longer used in core and additional modules
```
SKuipers, line 27: Hmmm. I wonder if this list can be collapsed down, or perhaps the change is mentioned here and  list itself is at the bottom in deprecations? It takes up a lot of space 🤔

SKuipers (approved):
Looking good! Going to merge and test.

CHANGELOG.txt
rossdotparker: My thinking was that having it take up space would be useful for anyone with custom code, as it'll drawn their attention to it.

SKuipers: Fair enough.

functions.php
yookoala: How do they manage their source code? It's not hard to patch it, but it's hard to submit any patch without a proper source hosting platform (e.g. GitHub, GitLab, BitBucket).

rossdotparker: They don't host online as far as I know. You could email a patch to andy@rapid36.com.

yookoala: Thanks :-)

PR #1492, opened 2021-11-01
Debug installer issue with database update

yookoala:
**Description**
* ~~Update `CHANGEDB.php` for session table changes. Replaced several older create table / add column lines to replace with a simple drop-table / create-table operation.~~
* Add several `error_log` to Updater class for improved debug.
* Revert previously `gibbonSession` table declaration in `gibbon.sql`.

**Motivation and Context**
* Fix current issue with session table updates in installation process.

**How Has This Been Tested?**
CI Environment.

### DIFF
--- a/gibbon.sql
+++ b/gibbon.sql
@@ -8067,20 +8067,6 @@ ALTER TABLE `gibbonYearGroup`
   MODIFY `gibbonYearGroupID` int(3) UNSIGNED ZEROFILL NOT NULL AUTO_INCREMENT, AUTO_INCREMENT=8;
 
 --
--- Add new table for session
---
-CREATE TABLE IF NOT EXISTS `gibbonSession` (
-    `gibbonSessionID` VARCHAR(40) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL,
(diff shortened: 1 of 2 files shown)


SKuipers:
I'm concerned that we're going backwards, introducing more complexity to handle a use case that perhaps shouldn't have existed, and might have had a different solution. Our practice is generally to never edit the gibbon.sql until release time, and never delete CHANGEDB.php lines. The CHANGEDB is always run ontop of gibbon.sql at release time, which is where the database changes are applied to the sql file.

If we drop the gibbonSession table during the updater, it's going to force-logout the administrator who is running the update, which is really not ideal.

Rather than making more changes to CHANGEDB.php and redacting lines, I'd rather reverse out the change where we added the session table manually to the gibbon.sql. Then, handle

yookoala:
Let's simply remove the gibbon.sql changes all together for now, then.

SKuipers (approved):
Thanks Koala, the additional error logs look useful for diagnosing update issues 👍

SKuipers:
I've made a post-merge fix because the installer was reporting an error for lines that were run successfully: https://github.com/GibbonEdu/core/commit/4fddcb01ee9ae1ffdef7652ff6c7ba97fbf85aca

PR #1493, opened 2021-11-02
System: removed deprecated core functions from functions.php

rossdotparker:
**Description**
Final core deprecation removal. All functions.php deprecated functions are now removed, other than printPagination, as this requires final tables to be refactored.

This requires releases of 17 additional modules: the changes to these have been made, and will shortly be tested, committed, pushed and released.

**Motivation and Context**
I love deleting my own rubbish old code and replacing it with something better.

**How Has This Been Tested?**
Locally

### DIFF
--- a/modules/Crowd Assessment/crowdAssess_view_discuss_postProcess.php
+++ b/modules/Crowd Assessment/crowdAssess_view_discuss_postProcess.php
@@ -123,16 +123,22 @@ if (isActionAccessible($guid, $connection2, '/modules/Crowd Assessment/crowdAsse
 
                         $homeworkNameSingular = $container->get(SettingGateway::class)->getSettingByScope('Planner', 'homeworkNameSingular');
 
+                        $notificationGateway = new \Gibbon\Domain\System\NotificationGateway($pdo);
+
                         //Create notification for homework owner, as long as it is not me.
                         if ($gibbonPersonID != $session->get('gibbonPersonID') and $gibbonPersonID != $replyToID) {
                             $notificationText = __('Someone has commented on your {homeworkName} for lesson plan "{lessonName}".', ['lessonName' => $name, 'homeworkName' => mb_strtolower(__($homeworkNameSingular))]);
-                            setNotification($connection2, $guid, $gibbonPersonID, $notificationText, 'Crowd Assessment', "/index.php?q=/modules/Crowd Assessment/crowdAssess_view_discuss.php&gibbonPlannerEntryID=$gibbonPlannerEntryID&gibbonPlannerEntryHomeworkID=$gibbonPlannerEntryHomeworkID&gibbonPersonID=$gibbonPersonID");
+                            $notificationSender = new \Gibbon\Comms\NotificationSender($notificationGateway, $session);
+                            $notificationSender->addNotification($gibbonPersonID, $notificationText, 'Crowd Assessment', "/index.php?q=/modules/Crowd Assessment/crowdAssess_view_discuss.php&gibbonPlannerEntryID=$gibbonPlannerEntryID&gibbonPlannerEntryHomeworkID=$gibbonPlannerEntryHomeworkID&gibbonPersonID=$gibbonPersonID");
+                            $notificationSender->sendNotifications();
                         }
 
                         //Create notification to person I am replying to
                         if (is_null($replyToID) == false) {
                             $notificationText = sprintf(__('Someone has replied to a comment on the {homeworkName} for lesson plan "{lessonName}".', ['lessonName' => $name, 'homeworkName' => mb_strtolower(__($homeworkNameSingular))]), $name);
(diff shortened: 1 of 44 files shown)


modules/Crowd Assessment/crowdAssess_view_discuss_postProcess.php
```diff
@@ -123,16 +123,22 @@
 
                         $homeworkNameSingular = $container->get(SettingGateway::class)->getSettingByScope('Planner', 'homeworkNameSingular');
 
+                        $notificationGateway = new \Gibbon\Domain\System\NotificationGateway($pdo);
```
SKuipers, line 126: Could we add Gibbon\Domain\System\NotificationGateway and Gibbon\Comms\NotificationSender as `use` declarations at the top, rather than putting the full namespace in the code.

modules/Crowd Assessment/crowdAssess_view_discuss_postProcess.php
```diff
@@ -123,16 +123,22 @@
...
+
                         //Create notification for homework owner, as long as it is not me.
                         if ($gibbonPersonID != $session->get('gibbonPersonID') and $gibbonPersonID != $replyToID) {
                             $notificationText = __('Someone has commented on your {homeworkName} for lesson plan "{lessonName}".', ['lessonName' => $name, 'homeworkName' => mb_strtolower(__($homeworkNameSingular))]);
-                            setNotification($connection2, $guid, $gibbonPersonID, $notificationText, 'Crowd Assessment', "/index.php?q=/modules/Crowd Assessment/crowdAssess_view_discuss.php&gibbonPlannerEntryID=$gibbonPlannerEntryID&gibbonPlannerEntryHomeworkID=$gibbonPlannerEntryHomeworkID&gibbonPersonID=$gibbonPersonID");
+                            $notificationSender = new \Gibbon\Comms\NotificationSender($notificationGateway, $session);
```
SKuipers, line 131: A more DRY way to write this could be to initialize a single sender below the gateway, then add notifications as needed in your conditionals, and run the `$notificationSender->sendNotifications();` method once at the bottom of this scope. The sender will know how to handle cases where there are no notifications to send.

modules/Finance/expenseRequest_manage_viewProcess.php
```diff
@@ -142,22 +144,25 @@
...
                                 foreach ($approvers as $approver) {
-                                    $notificationText = sprintf(__('Someone has commented on the expense request for "%1$s" in budget "%2$s".'), $row['title'], $row['budget']);
-                                    setNotification($connection2, $guid, $approver['gibbonPersonID'], $notificationText, 'Finance', "/index.php?q=/modules/Finance/expenses_manage_view.php&gibbonFinanceExpenseID=$gibbonFinanceExpenseID&gibbonFinanceBudgetCycleID=$gibbonFinanceBudgetCycleID&status2=&gibbonFinanceBudgetID2=".$row['gibbonFinanceBudgetID']);
+                                    $notificationSender->addNotification($approver['gibbonPersonID'], $notificationText, 'Finance', "/index.php?q=/modules/Finance/expenses_manage_view.php&gibbonFinanceExpenseID=$gibbonFinanceExpenseID&gibbonFinanceBudgetCycleID=$gibbonFinanceBudgetCycleID&status2=&gibbonFinanceBudgetID2=".$row['gibbonFinanceBudgetID']);
                                 }
+                                $notificationSender->sendNotifications();
```
SKuipers, line 165: Similar suggestion as above, with a single NotificationSender instance.

SKuipers (changes requested):
Looking good overall, I've added some comments about namespaces (which apply to the majority of the files, so I didn't repeat it), as well as using an single instance of NotificationSender. Want to check that out and give a shout with any questions?

modules/Crowd Assessment/crowdAssess_view_discuss_postProcess.php
rossdotparker: This makes a lot of sense in terms of ease of changing namespaces. I've just made the changes and will commit after I sort out the DRY-related comments below.

modules/Crowd Assessment/crowdAssess_view_discuss_postProcess.php
rossdotparker: Thanks for this. It is good to know how it works, even keeping track of the separate text notifications. I've committed all the suggested changes.

SKuipers (approved):
Looking good! In hindsight, we can just get NotificationSender from the container, so no need to create a gateway manually and instantiate the class. Also, it looks like NotificationSender does always clear the notifications list when sendNotifications is called, so a single sender instance can be used multiple times. I've made some tweaks to your PR to reflect this, if you want to check them out, and then merge.

rossdotparker:
Those final changes definitely tidy things up! The additional modules are still using the longer declaration, including the namespace, so that is something we can tidy up down the line. Cheers!

PR #1494, opened 2021-11-02
System: update Chart.js library to v3.6.0 and migrate existing charts

SKuipers:
Several of the charts used in Gibbon were not displaying well on mobile devices. This PR updates the chart.js library to the latest version and makes any required updates to chart options based on the recent version. It also removes old versions of the chart.js library.

**Motivation and Context**

**How Has This Been Tested?**
Locally.


modules/Rubrics/src/Visualise.php
```diff
@@ -155,6 +158,13 @@ public function renderVisualise($legend = true, $image = false, $path = '', $id
                         this.options.animation.onComplete = null;
                     }'),
                 ];
+            } else {
```
rossdotparker, line 161: Interested to know why the commented section has been left in here. I know there will be a good reason!

modules/Staff/report_absences_weekly.php
```diff
@@ -84,21 +84,22 @@
 
     // BAR GRAPH
     $chartConfig = [
-        'height' => '70',
+        'height' => '30vh',
         'tooltips' => [
```
rossdotparker, line 88: Should this one be singular?

SKuipers: Yes it should! Well caught.

modules/Rubrics/src/Visualise.php
SKuipers: Oops! Nope, no good reason this time 😆

PR #1560, opened 2021-11-02
Password Field & Demo Data

rossdotparker:
**Description**
Updates demo data so that all users can, after login refactoring, be logged in with password:

Gibbon123.

Also removes the password field from gibbonPerson.

**Motivation and Context**
The password field has been an appendix for a long time, and in a week of cleaning up it seemed like a good target.

**How Has This Been Tested?**
Locally


SKuipers (approved):
This is looking good. I've updated the dump.sql file so that the test data is also updated, and all CI tests are now passing.

PR #1561, opened 2021-11-03
Improve warning detection in test

yookoala:
**Description**
* Modify CI test to treat "Warning: " or "PHP Warning: " in server_log as test failure.

**Motivation and Context**
* Detects php issues when they are committed for submit. Much easier to identify the code relevant to the warning message.
* Just cleared all the issues that creates those kind of message. It is a good point in time to do so.

**How Has This Been Tested?**
* CI environment.


SKuipers (approved):
Looks great, thanks for the great work continuing to refine the CI pipeline 👍

PR #1563, opened 2021-11-09
Users photos automatic size and dimensions

Felipub:
**When uploading a user photo, it fixes automatically the size and dimensions of it.**

Our staff members enjoy uploading and browsing the photos of their students. It is also important to track them because they are many slum children with uncertain names, contact data and difficult to track otherwise. Our teachers need a fast way of uploading a photo of the child immediately with their phones.

The feature has been tested extensively by our teachers users. Before that it did not throw any error or anything.

Teachers upload photos with their phones directly every day since a month ago or so.

Here is my example:

Screenshot from 2021-11-09 20-35-53
There I am, ready to add a photo

Screenshot from 2021-11-09 20-36-07
I select this one that is almost 2MB and with the wrong dimensions 🙀

Screenshot from 2021-11-09 20-36-27
After submitting the form, success!

Screenshot from 2021-11-09 20-37-09
The thumbnail is showing in the user that I was editing 👍🏽

Screenshot from 2021-11-09 20-38-36
Finally I check the uploaded image and its properties and everything is within limits


SKuipers:
Thank you Felipe 😃 it's been a busy week but I look forward to checking out your PR soon!

SKuipers (approved):
Hi Felipe, thanks for your contributions and your patience as we work on wrapping up the current version of Gibbon. I'm going to merge this in and test it out. Ideally, I think we could aim to move this type of code into a class method somewhere, so that it is reusable. Something to consider for future versions. 

I think the tricky part about cropping automatically is maintaining the correct focal point of the image, so that it keeps the person centred in frame. I wonder if a more front-end approach might be useful in the future, such as a tool that lets users upload and preview cropped images before submitting the form.

PR #1564, opened 2021-11-15
Select Where - Update TableAware.php 

powertashton:
**Description**
This PR adds a simple public function to TableAware called selectWhere, that's based off of the selectBy function.

**Motivation and Context**
There's been several instances where I've had to create an entire new function in a gateway to achieve something that could've been done using this TableAware function, more specifically I'm currently working on a [reprographics module](https://github.com/powertashton/module-reprographics) where this is to be used. 

**How Has This Been Tested?**
Locally, within the reprographics module.

**Usage**
e.g. where different comparative operators are used to do a simple select beyond just the equals to operator.
`$orders = $orderGateway->selectWhere(['orderStatus = ' => 'Approved', 'orderDate >= ' => $startDate, 'orderDate <= ' => $endDate])->fetchAll();`


SKuipers:
My apologies for the wait! I've had a closer look and a think about this one, and I'm concerned that, if overused, it could lead to a reduction in the readability of the code. I see where you're looking to avoid creating gateway methods, but often this encapsulation of logic makes the code more readable and maintainable in the long run. 

Eg:
`$orders = $orderGateway->selectApprovedOrdersInDateRange($startDate, $endDate);`
vs.
`$orders = $orderGateway->selectWhere(['orderStatus = ' => 'Approved', 'orderDate >= ' => $startDate, 'orderDate <= ' => $endDate])->fetchAll();`

It also means if there's ever a new field introduced, such as an Active Y/N flag, or Reportable Y/N, then you just need to update the SQL logic in the gateway, rather than hunt down every instance of ->selectWhere() and update those too.

powertashton:
I see your points, I think my logic behind it was largely for cases where a query that could be done with a selectWhere is only made once, as it felt cumbersome to be creating more and more gateways just for one instance of code. 

I'll concede that you're probably right though, and simply bite the bullet and bother to make gateways 😅

PR #1571, opened 2021-11-19
Library - null coalesce in interface pages

hmerrettICHK:
**Description**
Adding null coalesce to interface pages in Library module

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**

### DIFF
--- a/modules/Library/library_browse.php
+++ b/modules/Library/library_browse.php
@@ -137,13 +137,13 @@ if (isActionAccessible($guid, $connection2, '/modules/Library/library_browse.php
     echo '</table>';
 
     //Get current filter values
-    $name = isset($_REQUEST['name']) ? trim($_REQUEST['name']) : null;
-    $producer = isset($_REQUEST['producer']) ? trim($_REQUEST['producer']) : null;
-    $category = isset($_REQUEST['category']) ? trim($_REQUEST['category']) : null;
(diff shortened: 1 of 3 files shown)


modules/Library/library_browse.php
```diff
@@ -137,13 +137,13 @@
...
-    $collection = isset($_REQUEST['collection']) ? trim($_REQUEST['collection']) : null;
-    $everything = isset($_REQUEST['everything']) ? trim($_REQUEST['everything']) : null;
+    $name = trim($_REQUEST['name'] ?? null);
+    $producer = trim($_REQUEST['producer'] ?? null);
+    $category = trim($_REQUEST['category'] ?? null);
+    $collection = trim($_REQUEST['collection'] ?? null);
+    $everything = trim($_REQUEST['everything'] ?? null);
 
-    $gibbonLibraryItemID = isset($_GET['gibbonLibraryItemID']) ? trim($_GET['gibbonLibraryItemID']) : null;
+    $gibbonLibraryItemID = trim($_GET['gibbonLibraryItemID'] ?? null);
```
yookoala, line 147: Doing `trim()` on null value seems odd. Also that would give you empty string `''` instead of `null` (as the original code does). Please make sure that the code below treat these value the same if they are empty string `''` as if they are `null`.

SKuipers (approved):

PR #1576, opened 2021-11-27
Fixed parameter (student_view_details instead of staff_view_details)

leofreitas:
**Description**
Fixed parameter (student_view_details instead of staff_view_details)

**Motivation and Context**
Wrong parameter causes broken link 

**How Has This Been Tested?**
Local and Travis


yookoala (approved):
Thanks for spotting the issue.

SKuipers (approved):
Well caught!

PR #1578, opened 2021-11-28
Identifying internal urls in the nameLinked function

leofreitas:
**Description**
Added function withAbsoluteUrl, because this way the internal links don't open a new window.

**Motivation and Context**
Name links are identified as external links

**How Has This Been Tested?**
Local and Travis


yookoala:
Strange. Why was the links opening new window in the first place?
Can you provide the step we can reproduce the original behaviour?

leofreitas:
Anywhere the nameListed function is called.

SKuipers (approved):

PR #1579, opened 2021-11-28
Added translation support for sortable attributes to the multi-select

leofreitas:
**Motivation and Context**
Added translation support for sortable attributes to the multi-select and column Role Category

**How Has This Been Tested?**
Local and Travis

**Screenshots**
Before
bug
After fixed
fixed

### DIFF
--- a/src/Forms/Input/MultiSelect.php
+++ b/src/Forms/Input/MultiSelect.php
@@ -93,7 +93,7 @@ class MultiSelect implements OutputableInterface, ValidatableInterface
     public function addSortableAttribute($attribute, $values)
     {
         $this->sortableAttributes[$attribute] = $values;
-        $this->sortBySelect->fromArray(array($attribute => __("Sort by " . $attribute)));
+        $this->sortBySelect->fromArray(array($attribute => __("Sort by ") . $attribute));
         return $this;
     }
 
--- a/src/Forms/Input/MultiSelect.php
+++ b/src/Forms/Input/MultiSelect.php
@@ -93,7 +93,7 @@ class MultiSelect implements OutputableInterface, ValidatableInterface
     public function addSortableAttribute($attribute, $values)
     {
         $this->sortableAttributes[$attribute] = $values;
-        $this->sortBySelect->fromArray(array($attribute => __("Sort by ") . $attribute));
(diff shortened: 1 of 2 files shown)


src/Forms/Input/MultiSelect.php
```diff
@@ -93,7 +93,7 @@ public function getID()
     public function addSortableAttribute($attribute, $values)
     {
         $this->sortableAttributes[$attribute] = $values;
-        $this->sortBySelect->fromArray(array($attribute => __("Sort by " . $attribute)));
+        $this->sortBySelect->fromArray(array($attribute => __("Sort by ") . $attribute));
```
yookoala, line 96: If we're to modify this to create a new translation string anyway, we should give a placeholder to attribute instead of just concat string:

```php
$this->sortBySelect->fromArray(array($attribute => __("Sort by {attribute}", ['attribute' => $attribute])));
```

See the `Locale::translate` function for [how it works](https://github.com/GibbonEdu/core/blob/4b7c173e03ebe65b1852f1a1fb3efe591827315c/src/Gibbon/Locale.php#L233-L283).

yookoala (changes requested):

src/Forms/Input/MultiSelect.php
leofreitas: I know Locale::translate. The question is that the term "Sort by " already exists and you don't need to create a new one.  Creating a new term would impact all language packs.

yookoala: That's OK then. I didn't notice the translation string "Sort by " exists. Thanks.

yookoala (approved):

src/Forms/Input/MultiSelect.php
leofreitas: I think we will need to register a task to refactor this also in the future.

yookoala: We'll probably need to mechanically create the new translation string, if possible.

SKuipers (approved):
I took a look in POEditor, and since there was a "Sort By" and a "Sort by " (with a space) I decided to update this one, so that we can correct it and not create a new string. So now there are "Sort By" and a "Sort by {attribute}" strings, and I've updated the code here so we don't have to remember to circle back to it in the future. Thanks Leo and Koala for the improvements!

PR #1580, opened 2021-11-29
Component listOptions fixed

leofreitas:
**Description**
In template the variable "absoluteURL"  generates a broken link and in the PaginatedView the "view" parameter is not necessary

**Motivation and Context**
OptionsList component doesn't work (e.g Staff/staff_view.php and Students/student_view_details.php)

**How Has This Been Tested?**
Local and Travis


yookoala (approved):
Removing the `['view' => '']` would give a different link to index.php, but I think it's still functionally the same.

SKuipers (approved):
Thanks for catching this! Will merge & test 👍

PR #1581, opened 2021-12-01
Alert levels review

leofreitas:
**Motivation and Context**
- In AlertLevel the 'name', 'description' and 'tag' fields don't need to be translatable, because they're editable
- The value of the 'name' field should not be used in conditionals because it's an editable field.

**How Has This Been Tested?**
Local and Travis

**Screenshots**
alertLevels


yookoala:
@SKuipers: I'm not familiar enough to review this part of the code, but I think there is a reason why 'name' field has a `__()` wrapped. Probably some fringe case you have encountered before?

SKuipers (commented):
I believe that the alerts are wrapped with translations so that the default values are still translated, as very few schools will change the alert level names. Once edited, yes, they will no longer translate, but before editing them they should translate to other languages. 

I like the fix for switching from the names to the IDs, as this is definitely an issue if someone does rename the alert levels. 

What are your thoughts on reverting the translation tags, but leaving the ID fix? Any translation issues that we haven't considered?

leofreitas:
Hi Sandra, I reverted the translation tags and left the ID fix. For  v24 i will think in a better solution for alert levels.

SKuipers (approved):
Thanks for the updated fix 👍

PR #1582, opened 2021-12-01
Autoimportappforms

Felipub:
When importing from file application forms of students, the official name of the parents was not taken into account. This resulted in a missing field later that had to be filled manually.

I added the missing fields in the yml file of the import.

The changes were tested in my cutting edge installation with good results.


SKuipers:
Thanks Felipe. No worries, we appreciate the intention either way 😃

PR #1586, opened 2021-12-12
Added Libyan Currency

azizly:
'LYD د.ل' => 'Libyan Dinar (د.ل)', 
added under currency list

**Description**

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


yookoala (approved):
LDY checks to be correct to the ISO 4217 standard code for [Libyan Dinar](https://en.wikipedia.org/wiki/Libyan_dinar). Seems reasonable.

PR #1587, opened 2021-12-13
Reports: Added default value for Yes/No criteria type

raynichc:
**Description**
Added a default value option when creating Yes/No reporting criteria.

**Motivation and Context**
Was requested by teachers to help streamline report writing process for certain criteria. 

**How Has This Been Tested?**
Locally.


SKuipers (approved):
Looking good, a nice and useful change, thanks! I can certainly see that, depending on the context of the yes/no, a default value could save some time. I bet this new field could possibly apply to other criteria types in the future, for any criteria that makes sense to have a default.

PR #1588, opened 2021-12-13
Timetable: Fixed calenders not persisting when timetable is changed

raynichc:
**Description**
Fixed calendars not persisting when timetable is changed.

**Motivation and Context**
Was requested by teacher because it can get annoying.

**How Has This Been Tested?**
Locally


SKuipers (approved):
That's odd that it would require ternary and not just the value of the variable itself 🤔 thanks for the fix, going to merge and test 👍

PR #1589, opened 2021-12-14
User Admin: User Status Log (WIP)

raynichc:
**Description**
This PR implements a status log for users. Status changes can be logged in a new table, with an attached reason. These can then be viewed in Manage Users.

Main functionality is implemented, just needs to be implemented in other areas where a person's status is changed.

**Motivation and Context**
A requested feature that came about due to many students coming and going because of covid.

**How Has This Been Tested?**
Locally.

**Screenshots**
image

Reason will only appear if status is changed.
image

### DIFF
--- a/CHANGEDB.php
+++ b/CHANGEDB.php
@@ -1058,10 +1058,12 @@ ALTER TABLE `gibbonPerson` CHANGE `genericAPIRefreshToken` `genericAPIRefreshTok
 ALTER TABLE `gibbonSession` CHANGE `sessionData` `sessionData` MEDIUMTEXT CHARACTER SET utf8 COLLATE utf8_general_ci NULL DEFAULT NULL;end
 ALTER TABLE `gibbonPerson` DROP `password`;end
 UPDATE `gibbonMarkbookColumn` SET gibbonPlannerEntryID=NULL WHERE gibbonPlannerEntryID=00000000000000;end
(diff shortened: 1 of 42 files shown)


SKuipers (changes requested):
This is looking nice and clean 👍  I wonder if we can get the PR merged sooner and work on refining the implementation, so that the strings are in for the string freeze. 

A couple things to change:
- We don't need to add tables to the gibbon.sql file, and editing it before the release can cause issues. All the changedb entries get applied to the gibbon.sql file during release. Can you revert those changes?
- The are already a number of actions in Manage Users, and this one, while useful, is less likely to be used than some of the other actions. I wonder about moving the View Status Log to a header action on the Edit User page, as well as adding the link to the student profile next to the school history.

Something to consider:
I wonder if the status log could also track student enrolment changes :thinking: Sometimes students move from one form group to another, and it would be nice to have this logged as well.

raynichc:
Hi Sandra, I've made the changes you've asked for and opened the PR for review.

SKuipers:
Brilliant, thanks for the quick fixes! I'll look at reviewing this and getting it merged in today 👍

SKuipers (approved):

SKuipers (approved):
Looking great, going to merge and test 👍

SKuipers:
Thanks Ray, tests out great! I wonder about pre-populating the table with the current status of each user, so that the status log isn't blank to begin with 🤔 @rossdotparker What do you think, useful or overkill?

rossdotparker:
This looks like a great addition @raynichc and overcomes an issue that has existed for a long time, which is people coming and going in schools. @SKuipers that could be useful, although is a the first status a "change" to be logged? I guess it would be useful to know when that first status was enacted...

PR #1593, opened 2021-12-30
School Admin: add option to configure activity types individually

SKuipers:
Currently, activity types are a comma separated list and other activity settings are applied globally. This PR creates a gibbonActivityType table, and moves the activity types into individual rows, so that settings such as enrolment type, waiting list, max per student, etc. can be configured per activity type.

Some schools run many different types of activities, and a single global setting isn't sufficient for each type of activity. This enables schools to open different types of activity registration at different times as well.

**Motivation and Context**
Getting some features in before the string freeze. I will review and merge this one to get the strings in, but we can still review and tweak the functionality as needed after the freeze.

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2021-12-30 at 11 33 34 AM


SKuipers:
I've done a cursory review and will merge this in so we can get the strings in, feel free to also review later when you have time.

rossdotparker:
Thanks for this Sandra. We've discussed a couple questions in Slack, and it looks good from my point of view, except the Type Add error, which you are looking into. Cheers!

PR #1599, opened 2022-01-11
System: sanitization improvements

SKuipers:
This PR improves sanitization of user values on each of the process pages in the core. This is another step in the improvement of how data is handled by the core, specifically in preventing stored/reflected XSS. The sanitize method has currently been used for external user data such as application forms, but via this PR it has now been applied to internal process pages as well.

Although large, this PR makes a nearly-identical change to the top of each page. The edge cases are adding flags to handle fields that need to have HTML data retained rather than removes. The validation class handles this sanitization of HTML using an allowableHTML setting and DOM traversal.

Even though it is late in the release cycle, Ross and I had a chat and felt it would be good to get these changes into the next version. We'll have to do some extra testing to ensure we haven't missed any of the edge cases, but overall it should be working well and ready to merge.

Edit: Sanitizing the whole POST array is a little heavy-handed, but given the scale of the code, this was a good first step, before we introduce more nuanced sanitization and validation methods.


rossdotparker:
I've reviewed the changes, and as you say, most of them are just a few changes to the top, some of which include reference to HTML content. This is definitely a good change to include at this point, given disclosures that have been made recently. Ideally issues will be found before the upcoming release, but if not we can always do a minor release once we round up and patch any issues. I am going to merge now and then test.

rossdotparker:
Thanks for all of the hard (and fast!) work you put into this fix.

PR #1600, opened 2022-01-16
Added translation support

leofreitas:
**Description**
Translation support for these terms has also been added in Poeditor (tag v23_review)

**How Has This Been Tested?**
Local and Travis


SKuipers (approved):
Well caught! 😄

PR #1601, opened 2022-01-19
Fixed HTML appearing in Age column in header row on export

leofreitas:
**Description**
- Fixed HTML appearing in Age column in header row on export
- Added translation support for the term "DOB"

**Motivation and Context**
Some HTML appearing in header row on export

**How Has This Been Tested?**
Local and Travis


SKuipers (approved):
Thanks, well caught!

PR #1603, opened 2022-02-02
Installer: fix homepage link generation

yookoala:
**Description**
* Fix the "go to Gibbon homepage" link in the message on step 4 of installation.
* Update InstallCept to check if the link goes to the home page.

**Motivation and Context**
* See issued reported by [tiekubd](https://ask.gibbonedu.org/profile/tiekubd) on [this Gibbon Support Forum thread](https://ask.gibbonedu.org/discussion/comment/9434)

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/installer/install.php
+++ b/installer/install.php
@@ -186,7 +186,7 @@ try {
 
         // Show success message if the installation is a complete success.
         if ($message === null) {
-            $absoluteURL = str_replace('/installer/install.php', '', $_SERVER['PHP_SELF']);
+            $absoluteURL = str_replace('/installer/install.php', '/', $_SERVER['PHP_SELF']);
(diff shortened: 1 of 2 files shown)


installer/install.php
```diff
@@ -186,7 +186,7 @@
 
         // Show success message if the installation is a complete success.
         if ($message === null) {
-            $absoluteURL = str_replace('/installer/install.php', '', $_SERVER['PHP_SELF']);
+            $absoluteURL = str_replace('/installer/install.php', '/', $_SERVER['PHP_SELF']);
```
SKuipers, line 189: At this point in the installation, we have an absoluteURL in the session (as well as the database). I wonder, would it be more robust to use that here, rather than trying to interpret the absoluteURL from the $_SERVER['PHP_SELF']?

yookoala: Added another commit for that. Please check.

SKuipers (approved):
Thanks for the fix Koala, I'm going to merge and test it out 👍

PR #1608, opened 2022-02-16
System Admin: add an Upload Photos & Files page, replacing the old Import User Photos

SKuipers:
This replaces the Import User Photos action with a more robust file uploader.

- Allows for uploading user photos, custom field files, and personal document files
- Presents users with a 3 step upload process, including a pre-check of the uploadable files
- Handles Zip files with nested folder structures
- Handles filenames with different unique identifiers (current options: username or studentID)
- Handles filenames with separator characters and identifier position
- Automatically scales and crops user photos on upload, saving lots of time and headache
- Optionally enables zooming and focusing user photos, to help when portraits are zoomed out too much
- Optionally skips replacing files that already exist

**Motivation and Context**
This gives schools much more flexibility in the types of files they can upload, as well as includes a pre-check before the upload completes.

**How Has This Been Tested?**
Locally and in production.

**Screenshots**
Screenshot 2022-02-16 at 16-51-37 DEMO - Gibbon - System Admin

Screenshot 2022-02-16 at 4 51 30 PM

Screenshot 2022-02-16 at 4 51 52 PM


modules/System Admin/file_upload.php
```diff
@@ -0,0 +1,133 @@
...
+
+        $row = $form->addRow()->addClass('userPhotos');
+            $row->addLabel('focalY', __('Vertical Focal Point'));
+            $row->addRange('focalY', 0, 100, 5)->setValue(50);
+
+        $row = $form->addRow();
+            $row->addFooter();
+            $row->addSubmit();
+
+    } else if ($step == 3) {
```
rossdotparker, line 126: Where did $step == 2 go?!

rossdotparker:
A great new feature for v24! I am going to merge it, even thought I dropped one question, because I know this has been tested in production already. My question was more out of curiosity. I guess the answer has to do with the *Preview.php script.

modules/System Admin/file_upload.php
SKuipers: Well spotted. Step 2 needed to be on a hybrid process/display page, in order to peek into the contents of the zip file and display the matches. Its enough of a chunk of code that I still wanted it in it's own page, rather than crammed into the initial form.

PR #1611, opened 2022-02-18
System: improved display of IDD country codes in phone number entry field

rossdotparker:
**Description**
Does what it says on the can. A simple fix to address a UX issue when selecting IDD country code in phone numbers.

**Motivation and Context**
This was partly address by placing the country of installation at the top, but this should offer an improved experience for other countries.

**How Has This Been Tested?**
Locally, including with changes to the country of installation

**Screenshots**
Screenshot 2022-02-18 at 11 49 51 AM


SKuipers (approved):
Looks good 👍

PR #1615, opened 2022-03-24
Pagination Refactoring

rossdotparker:
**Description**
printPagination in functions.php has been deprecated since v16. This removes the last two calls of this function by refactoring the tables they belong to. Next steps are to do the same for some additional modules, then this function can be removed in v25.

**Motivation and Context**
Keeping house.

**How Has This Been Tested?**
Locally

**Screenshots**
Not worth the bytes.


SKuipers (approved):
Looking great! Lovely to see another couple tables refactored 😃

PR #1622, opened 2022-05-16
Feature: Multifactor authentication.

powertashton:
**Description**

This PR creates the option for a user to enable MFA for logging into their account and changing their password
**Motivation and Context**

Higher security for users who want it, with potential for requiring MFA for admin and other users with sensitive data access, and to require MFA before allowing users to perform certain actions etc etc 

**How Has This Been Tested?**

Locally on a mac homebrew PHP 8 and Apache environment, as well as by the tests that will run once I make this PR. I suspect these tests might fail on validating the preferences page, but will make changes to the tests to not only fix this but also potentially enable the ability to check things are all working with the MFA later... Just want to get the big core changes in before the string freeze.

**Screenshots**

image
image


powertashton:
(The fail on the tests are to do with the new field in the forms. I will fix this hopefully by the end of tomorrow)

SKuipers (approved):
Finally had a moment to review and merge 😅 This is looking good! I notice there are still a handfull of TODO: items in the comments, but none of them appear to be urgent, so something to look at in v25. For instantiating the class, it'd be nice to move to using the container rather than using `new RobThree\Auth\TwoFactorAuth`, this would enable wiring up the name of the instance in a service container as well, so it's consistent each time the object is created. Another thing for v25, for now this is looking great and I'm going to merge and test. Thanks Ashton for the shiny new feature 😃

PR #1624, opened 2022-05-30
Admissions: new flexible Form Builder functionality and Admissions process

SKuipers:
It's here! This PR is a beast 😆 there was no way to really split it up smaller... I think we'll just have to review and test as we use it. The functionality is _functional_, there are various visual improvements on my to-do list, but the majority of the core admissions processes are in. 

It's late and I'm tired and it's hard to describe all the of the various features and changes. I don't usually skimp out on the PR descriptions, but this one will have to come in time, when I attempt the documentation 😅 

**Motivation and Context**
Flexible application forms and, in the future, other flexible forms such as public registration, staff application, etc.

**How Has This Been Tested?**
Tested locally with v24. I have created and tested a complete application with every field and function turned on, and all went well. I'm sure there will be minor issues to iron out.

**Screenshots**
Too many fancy things to get them all...
Screenshot 2022-05-30 at 7 48 52 PM


CHANGEDB.php
```diff
@@ -511,17 +511,40 @@
...
+INSERT INTO `gibbonAction` (`gibbonModuleID`, `name`, `precedence`, `category`, `description`, `URLList`, `entryURL`, `entrySidebar`, `menuShow`, `defaultPermissionAdmin`, `defaultPermissionTeacher`, `defaultPermissionStudent`, `defaultPermissionParent`, `defaultPermissionSupport`, `categoryPermissionStaff`, `categoryPermissionStudent`, `categoryPermissionParent`, `categoryPermissionOther`) VALUES((SELECT gibbonModuleID FROM gibbonModule WHERE name='Admissions'), 'Admissions Accounts', 0, 'Prospective Students', '', 'admissions_manage.php,admissions_manage_add.php,admissions_manage_edit.php,admissions_manage_delete.php', 'admissions_manage.php', 'Y', 'Y', 'Y', 'N', 'N', 'N', 'N', 'Y', 'N', 'N', 'N');end
```
rossdotparker, line 514: Do we not need some UPDATES to match these, at the end of the current lot of CHANGEDB.php for cutting edge installations that have already run the old 514 to 516?

rossdotparker:
Congratulations Sandra! You are right, this thing is a beast, and a very significant one at that. I've had a skim through, but as per your comments, it'll be something we review through use. It's labelled as "beta" for v24, so available for testing and refinement, but the old system remains in place too, so there ought to be no show stoppers. I'm going to merge and then test the update path and play with building a form. Looking forward to a tour, and perhaps this is something we can do on Zoom so that any interested parties might join us.

PR #1626, opened 2022-06-16
Translation support for new feature (formBuilder)

leofreitas:
**Description**
Added Translation support for new feature (formBuilder)

**Motivation and Context**
Translation support for new feature (formBuilder)

**How Has This Been Tested?**
Local


SKuipers:
Thanks Leo! My thinking for not translating the Beta message is that it will be removed again in v25, and it is only visible on admin-facing pages, so best to not have to translate the whole paragraph just to delete it so soon.

leofreitas:
Thanks Sandra. Translation support removed.

SKuipers:
Ah shoot, I could have swore I had merged this in before releasing v24, my apologies! I'll merge it in now 😅

PR #1635, opened 2022-09-11
Integration with bigbluebutton

cui-soft:
How are you, everybody?

I integrated [bigbluebutton](https://bigbluebutton.org/) on gibbon project.
So teacher and students can use video chat on planners.


fvlasie:
This is the real pull request for BigBlueButton with updated code so I will close the other one. Thank you !

modules/Staff/src/Messages/VideoChatRequest.php
```diff
@@ -0,0 +1,63 @@
...
+
+    public function getTitle() : string
+    {
+        return __('Video Chat Request');
+    }
+
+    public function getText() : string
+    {
+        return __("You have a meeting with ".$absence['from'].".");
+    }
```
yookoala, line 47: This use of the translation function doesn't work.

Suggested to change to:
```php
    public function getText() : string
    {
        return __('You have a meeting with {person}.', [
            'person' => $this->absence['from'],
        ]);
    }
```

yookoala (changes requested):

modules/Staff/src/Messages/VideoChatRequest.php
cui-soft: Ok, thank you for your suggestion.
Let me fix it.

modules/Staff/src/Messages/VideoChatRequest.php
```diff
@@ -1,7 +1,7 @@
 <?php
 /*
 Gibbon, Flexible & Open School System
-Copyright (C) 2010, Ross Parker
+Copyright (C) 2022, Cui wenlan
```
yookoala, line 4: Please maintain the original copy right owner.

cui-soft: OK.
I am also implementing Rocketchat integration now.
Should I insert the original copyright owner for new added files?

yookoala: 1. Please keep separated features in separated pull request. It's better for reviewers that way.
2. Please insert the original copyright owner on new files.

cui-soft: OK, I will follow you.
I've created v25.0.01 branch.
Let me delete current pull request and recreate it.

PR #1636, opened 2022-09-12
PHP 8.2 Compatibility

yookoala:
**Description**
* Add PHP 8.2 as a test target of Gibbon.
* Fix all issues that causes deprecation warning in the unit test.

**Motivation and Context**
* [PHP 8.2RC1](https://www.php.net/archive/2022.php#2022-09-01-4) has been released. It's perhaps time to include the version for testing.

**How Has This Been Tested?**
* CI environment.


SKuipers (approved):
Looks great, thanks for helping keep us compatible with the latest PHP versions 😃

PR #1641, opened 2022-09-19
System: update access denied message for non-logged in users

SKuipers:
This PR replaces the standard "you do not have access" message when users are not logged in. I tried a few approaches and decided to handle it as a string replacement because there are over 1300 instances of the string in the core and additional modules. This approach introduces the least code changes and enables fine-tuning the wording of this message in the future.

From:
Screen Shot 2022-09-19 at 11 29 09 AM

To:
Screen Shot 2022-09-19 at 11 34 34 AM

**Motivation and Context**
This helps improve the user experience when someone clicks a link externally, especially from emails, but needs to login first to access the page. Rather than telling them access is denied, it prompts them to login.

**How Has This Been Tested?**
Locally.


rossdotparker (approved):
Hi Sandra, this is a clever way to go, and definitely less likely to backfire than doing all the replacements. Do you think you would come back to do the replacements in due course, or is this the permanent solution? I'm happy to approve, as it is definitely an enhancement, and the code looks good. Ross.

PR #1642, opened 2022-09-22
Remove legacy class namespace compatibility

yookoala:
**Description**
* Remove autoloading support to autoload class like: "Gibbon\SomeModule\MyClass" -> "modules/Some Module/MyClass.php"

**Motivation and Context**
* The backward compatibility was introduced in f4fde610be (Nov 2018).
* As the comment stated, it was "Temporary backwards-compatibility for external modules (Query Builder)".
* I've checked the latest version of [Query Builder](https://github.com/GibbonEdu/module-queryBuilder/tree/806de79f3f19564637fae038fb87833b852486a2). The legacy class namespace style doesn't seem to exists anymore.

**How Has This Been Tested?**
* Locally


SKuipers (approved):
Looks good, happy to trim this out! Thanks Koala 👍

PR #1645, opened 2022-09-26
Fix various Markbook issues

yookoala:
**Description**
* Fix PDOException references: Missed the root namespace backslash.
* Replace obsoleted executeQuery calls in MarkbookView. Using select instead of deprecated executeQuery.
* Fix mistake in MarkbookView::error() created in previous refactoring.
* Remove obsoleted methods:
   * Remove MarkbookView::filterByFormOptions().
   * Remove MarkbookView::filterByQuery().

**Motivation and Context**
* Clean up several issues in the Markbook module.

**How Has This Been Tested?**
* CI Environment.


SKuipers (approved):
Thanks, looking good! The try-catch can also be removed from any of the old executeQuery statements because the pdo connection handles that internally now.

PR #1646, opened 2022-09-26
Fix Reports NullContext compliance with the DataContext interface

yookoala:
**Description**
* Fix the parameter list of NullContext::getIdentifiers().
* Improve documentation to the DataContext interface.

**Motivation and Context**
* NullContext::getIdentifiers() has 1 more parameter than the DataContext::getIdentifiers() and hence failed to implement DataContext. Reading the code and see the parameter list is wrong.

**How Has This Been Tested?**
* Locally
* CI environment.


SKuipers (approved):
👍

PR #1648, opened 2022-09-26
Refactor: make getEditor function obsoleted

yookoala:
**Description**
* Editor::getElement() is refactored to handle its own rendering instead of relying on getEditor.
* All getEditor calls can then be rewrite to use Editor directly.
* The getEditor() function is marked obsoleted. It is kept for module developers to adapt the change.

**Motivation and Context**
The function getEditor can be, and should be, completely replaced by the Editor class itself.

**How Has This Been Tested?**
* CI Environment

**Screenshots**

### DIFF
--- a/src/Forms/Input/Editor.php
+++ b/src/Forms/Input/Editor.php
@@ -27,7 +27,7 @@ namespace Gibbon\Forms\Input;
  */
 class Editor extends Input
 {
-    protected $guid;
+    protected $tinymceInit = true;
     protected $rows = 20;
     protected $showMedia = false;
     protected $initiallyHidden = false;
@@ -38,12 +38,10 @@ class Editor extends Input
     /**
      * Create a tinyMCE rich-text editor input.
      * @param  string  $name
-     * @param  string  $guid
      */
-    public function __construct($name, $guid)
+    public function __construct(string $name)
     {
         $this->setName($name);
-        $this->guid = $guid;
     }
 
     /**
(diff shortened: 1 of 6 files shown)


src/Forms/Input/Editor.php
```diff
@@ -38,12 +38,10 @@ class Editor extends Input
     /**
      * Create a tinyMCE rich-text editor input.
      * @param  string  $name
-     * @param  string  $guid
      */
-    public function __construct($name, $guid)
+    public function __construct(string $name)
```
SKuipers, line 42: If the constructor is changing we will need to update the [FormFactory ](https://github.com/GibbonEdu/core/blob/v25.0.00/src/Forms/FormFactory.php#L167)as well. Ideally, in the future, the planner scripts will be refactored to use the editor through the Form library rather than instantiating it on its own (those are some of the final areas in the core needing forms refactored).

SKuipers (changes requested):
Huge thanks for clearing out these old functions from the codebase! One day maybe we can get rid of functions.php entirely 😁

src/Forms/Input/Editor.php
yookoala: Sure :-)

yookoala:
Issues about this PR has been resolved.

SKuipers (approved):
Thanks, looks great! Ideally, as even more of these areas are refactored, they'll instantiate an editor through the FormFactory rather than manually, but this is a great next step to deprecate yet another function call. Nicely done!

PR #1649, opened 2022-09-26
Rubrics: moved ajax script location from src to module root

yookoala:
**Description**
Move the AJAX script of the Rubrics module up one level from "src".

**Motivation and Context**
* The src folder should only contain class files for autoloading.

**How Has This Been Tested?**
* Locally
* CI Environment


SKuipers (approved):
👍

Issue #1643, opened 2022-09-22
Discussion: retire the use of strftime from code base


yookoala:
### Problem

Some of the code is still using the deprecated function [strftime](https://www.php.net/manual/en/function.strftime.php). Includes:

* [modules/Formal Assessment/moduleFunctions.php](https://github.com/GibbonEdu/core/blob/9879b5c8511229e7acd39500fe47f9d6999e19eb/modules/Formal%20Assessment/moduleFunctions.php)
* [modules/Timetable/moduleFunctions.php](https://github.com/GibbonEdu/core/blob/9879b5c8511229e7acd39500fe47f9d6999e19eb/modules/Timetable/moduleFunctions.php)
* [modules/Finance/src/Forms/FinanceFormFactory.php](https://github.com/GibbonEdu/core/blob/9879b5c8511229e7acd39500fe47f9d6999e19eb/modules/Finance/src/Forms/FinanceFormFactory.php)
* [src/Services/Format.php](https://github.com/GibbonEdu/core/blob/9879b5c8511229e7acd39500fe47f9d6999e19eb/src/Services/Format.php)

The function strftime() and gmtstrftime() are [marked deprecated in 8.1 and set to be removed in 9.0](https://php.watch/versions/8.1/strftime-gmstrftime-deprecated). The usage should be removed.

### Proposed Solution

The recommended way is to rewrite them:
* If the date format does not need to be localized, it can be replaced with [DateTimeInterface::format()](https://www.php.net/manual/en/datetime.format.php).
* If the date format needs localization, it can be replaced with [IntlDateFormatter::format()](https://www.php.net/manual/en/intldateformatter.format.php). (see [this PHP Watch article](https://www.php.net/manual/en/intldateformatter.format.php) for details). The class is a part of the [Intl](https://www.php.net/manual/en/book.intl.php) bundle available since PHP 5.3.0. It's installed by default or available as packages like "php-intl" for many Linux distros.

One big issue is [Gibbon\Services\Format](https://github.com/GibbonEdu/core/blob/9879b5c8511229e7acd39500fe47f9d6999e19eb/src/Services/Format.php) relies heavily on the strftime() time formatting string. That means all usage of the methods of this class would rely on strftime(). That means the usages will all needed to be converted to date format of either DateTimeInterface::format() or IntlDateFormatter::format().

### Alternatives

* We can leave this be and re-visit later. Thou it is not my recommendation.
[...]

PR #1650, opened 2022-09-26
Format: replace strftime with IntlDateFormatter

yookoala:
**Description**
* Replace deprecated strftime function calls from internal uses in:
   - modules/Formal Assessment/moduleFunctions.php
   - modules/Timetable/moduleFunctions.php
   - modules/Finance/src/Forms/FinanceFormFactory.php
   - src/Services/Format.php

**Motivation and Context**
* To partially address #1643.

**How Has This Been Tested?**
* CI Environment.


composer.json
```diff
@@ -40,6 +40,7 @@
     "require": {
         "php" : "^7.3",
         "ext-curl": "*",
+        "ext-intl": "*",
```
SKuipers, line 43: Hmm. I wonder how we can handle cases where this is not available? The required PHP extensions here will only be noticeable by devs who are running composer, rather than the average stable install. 🤔

modules/Timetable/moduleFunctions.php
```diff
@@ -2054,6 +2054,11 @@ function renderTTSpace($guid, $connection2, $gibbonSpaceID, $gibbonTTID, $title
             $output .= "<span style='font-weight: normal; font-style: italic;'>".__('Time').'<span>';
             $output .= '</th>';
             $count = 0;
+
+            // Formatter to format date as day of week (full name based on locale).
+            $dayOfWeekNameFormatter = new \IntlDateFormatter(null);
```
SKuipers, line 2059: I wonder if the Format class can handle this internally, so we are not using IntlDateFormatter in various places in the codebase 🤔 This way, we can possibly have a fallback case inside the Format class. What are your thoughts on Format::dayOfWeekName and Format::monthName methods, to then keep the implementation of IntlDateFormatter internally in the Format class?

SKuipers (commented):
Looking great, thanks for catching these function deprecations and implementing a solution. I wonder about keeping the implementation of IntlDateFormatter internal to the Format class, checking for it and creating it in the constructor, so that the Format class can internally handle cases where ext-intl isn't available. This way, if it ever changes again, the implementation is contained within the Format class, and the rest of the codebase just relies on whatever the Format class provides as methods. What do you think?

composer.json
SKuipers: We could add it to the version.php list of extensions, but it'd be nice to be able to not require the extension if possible, and handle a fallback case internally in the Format class if possible.

yookoala: The [strftime](https://www.php.net/manual/en/function.strftime.php) documentation cited [date()](https://www.php.net/manual/en/function.date.php) and [intldateformatter::format()](https://www.php.net/manual/en/intldateformatter.format.php) as the alternatives. The class [IntlDateFormatter](https://www.php.net/manual/en/class.intldateformatter.php) comes with the the [intl](https://www.php.net/manual/en/book.intl.php) extension. So it seems to be the safest path to replace strftime for us.

SKuipers: Yes, I think it's a good replacement, I wonder about how to gracefully handle the additional library without needing to requite ext-intl for all systems (especially those with Gibbon already installed that are updating, particularly XAMPP systems seem to be very minimal on PHP extensions by default).

yookoala: [This article](https://arjunphp.com/install-intl-extension-xampp/) states that XAMPP has already shipped with Intl extension. You just need to edit php.ini and uncomment a certain line to have it enabled. It's a bit of trouble, but its not very hard to do so.

I've also checked Packagist. There is one [strftime replacement library](https://packagist.org/packages/php81_bc/strftime) but it's also written with IntlDateFormatter.

P.S. We can use the [php81_bc/strftime](https://packagist.org/packages/php81_bc/strftime) library for all the functions exposing the strftime formatting string. That way we can be lazy with the parameter rewrite.

modules/Timetable/moduleFunctions.php
yookoala: That sounds reasonable. Will work on those.

yookoala: Done.

SKuipers (approved):
👍

PR #1654, opened 2022-09-28
System: gracefully handle empty HTTP_HOST

yookoala:
**Description**
* $_SERVER['HTTP_HOST'] would be unset for CLI access (e.g. unit test) and thus should be used with fallback value coercion.

**Motivation and Context**
* To suppress warning message in test envrionment.

**How Has This Been Tested?**
* Locally with phpunit.


SKuipers (approved):
👍

PR #1655, opened 2022-09-28
Locale: improved handling for unsupported locale

yookoala:
**Description**
* setlocale might fail for some system lacking the locale to set.
* Gibbon should gracefully handle these case and tell administrator / user about it.

**Motivation and Context**
* Some tests are unable to run locally because my system do not have Spanish locale.
* PHPUnit should gracefully skip those test instead of complaining of incorrect translation.
* Further investigation found issue with core logic in handling potential issue(s) with system update and missing locale.

**How Has This Been Tested?**
* Locally.
* CI Environment.

**Screenshots**
Screenshot from 2022-09-29 00-26-46


yookoala:
Will work on this. Close for now.

PR #1657, opened 2022-09-29
System: extend ModuleGateway functionality

yookoala:
**Description**
* Add phpdocs to existing ModuleGateway methods.
* Add method `getActiveAdditional()`.
* Add ModuleGatewayInterface for flexibility of testing against it.
* Add ModuleGatewayInterface as a core service.

**Motivation and Context**
* Plan to make Locale a service loaded by container. Wanted to delegate the [database query in Locale::setTextDomain](https://github.com/GibbonEdu/core/blob/3862ccd1da00a0acca3e81b114097d54e026db1f/src/Gibbon/Locale.php#L108-L116) to an interface.

**How Has This Been Tested?**
* Locally.


SKuipers (commented):
Hi Koala, I can see where you're going here, however I'd like to avoid adding interfaces for the gateway classes if possible, as this will duplicate a large number of classes and I feel it could make the gateways much more cumbersome to maintain. The container is already able to instantiate the ModuleGateway without needing an interface because of auto-wiring, so it also simplifies the CoreServiceProvider by not having to add every gateway class here too. What do you think?

yookoala:
No real objection here. I'll remove the interface then.

SKuipers (approved):
Thanks! 👍

PR #1658, opened 2022-09-29
Markbook: added Term filter to student view, when Group Columns by Term enabled

rossdotparker:
**Description**
Adds a Term filter to the student view of the Markbook, which defaults to the current term when possible. This is only made available when the Markbook setting *Group Columns by Term* is set to Yes.

**Motivation and Context**
Requested by a school whose assessment tends to be oriented to term, and who already uses Group Columns by Term.

**How Has This Been Tested?**
Locally, with a limited set of testing data.

**Screenshots**
Screenshot 2022-09-29 at 7 52 40 PM


SKuipers (approved):
Looking good, I can see where this would be useful for schools that make heavy use of the markbook.

PR #1661, opened 2022-10-03
Tests: moved service behaviour away from FormTest

yookoala:
**Description**
* The behaviour of From::create was actually testing the behaviour of container with ViewServiceProvider. Moved the unit test out to its own thing to improve the tests.

**Motivation and Context**
* To remove dependency of test that requires gibbon.php setup.

**How Has This Been Tested?**
* Locally
* CI environment.


SKuipers (approved):
👍

PR #1662, opened 2022-10-03
System: replace archiveNotification with gateway method

yookoala:
**Description**
* Mark archiveNotification() deprecated.
* Add NotificationGateway::archiveNotificationForPersonAction() that provdes the exact feature.
* Rewrite Finance module usage of archiveNotification() to NotificationGateway::archiveNotificationForPersonAction().

**Motivation and Context**

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -200,7 +200,20 @@ function renderGradeScaleSelect($connection2, $guid, $gibbonScaleID, $fieldName,
     return $return;
 }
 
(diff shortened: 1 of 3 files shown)


SKuipers (approved):
Thanks for catching this strange edge-case, I'd never even noticed the archiveNotification function before. Seems to be only in use in one area, but I guess other areas could also auto-archive their notifications now a lot easier. 👍

PR #1663, opened 2022-10-03
System: replace getAlert with AlertLevelGateway method

yookoala:
**Description**
* Create AlertLevelGateway class with a getByID() method.
* Mark getAlert() deprecated.
* Rewrite all getAlert() calls with AlertLevelGateway::getByID().

**Motivation and Context**

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/src/Domain/System/AlertLevelGateway.php
+++ b/src/Domain/System/AlertLevelGateway.php
@@ -1,0 +1,76 @@
+<?php
+/*
+Gibbon, Flexible & Open School System
+Copyright (C) 2010, Ross Parker
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with this program. If not, see <http://www.gnu.org/licenses/>.
+*/
+
+namespace Gibbon\Domain\System;
+
+use Gibbon\Domain\QueryableGateway;
+use Gibbon\Domain\Traits\TableAware;
+
+/**
+ * AlarmLevel Gateway.
+ *
+ * @version v25
+ * @since   v25
+ */
+class AlertLevelGateway extends QueryableGateway
+{
+    use TableAware;
+
+    private static $tableName = 'gibbonAlertLevel';
+    private static $primaryKey = 'gibbonAlertLevelID';
+
+    const LEVEL_HIGH = 1;
+    const LEVEL_MEDIUM = 2;
+    const LEVEL_LOW = 3;
+
+    /**
+     * Get the specified alert level.
+     *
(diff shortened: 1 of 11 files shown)


src/UI/Dashboard/ParentDashboard.php
```diff
@@ -145,6 +146,7 @@ public function getOutput()
 
     protected function renderChildDashboard($gibbonPersonID, $dateStart)
     {
+        global $container;
```
SKuipers, line 149:
```suggestion
```

src/UI/Dashboard/ParentDashboard.php
```diff
@@ -153,7 +155,11 @@ protected function renderChildDashboard($gibbonPersonID, $dateStart)
 
         $return = false;
 
-        $alert = getAlert($guid, $connection2, 002);
+        /**
+         * @var AlertLevelGateway
+         */
+        $alertLevelGateway = $container->get(AlertLevelGateway::class);
```
SKuipers, line 161: This class has a container-aware interface, which auto-injects a reference to the container and enables the getContainer() method internally. This is a bit cleaner way for classes that need access to the container to use it, rather than globals.

```suggestion
        $alertLevelGateway = $this->getContainer()->get(AlertLevelGateway::class);
```

src/Domain/System/AlertLevelGateway.php
```diff
@@ -0,0 +1,76 @@
...
+     *                      translated unless $translated is false.
+     */
+    public function getByID(int $gibbonAlertLevelID, bool $translated = true)
+    {
+        $sql = 'SELECT * FROM gibbonAlertLevel WHERE gibbonAlertLevelID=:gibbonAlertLevelID';
+        $row = $this->db()
+            ->selectOne($sql, [
+                'gibbonAlertLevelID' => $gibbonAlertLevelID,
+            ]);
+        return ($row !== false && $translated)
```
SKuipers, line 64:
```suggestion
        return (!empty($row) && $translated)
```

src/Domain/System/AlertLevelGateway.php
```diff
@@ -0,0 +1,76 @@
...
+ * @since   v25
+ */
+class AlertLevelGateway extends QueryableGateway
+{
+    use TableAware;
+
+    private static $tableName = 'gibbonAlertLevel';
+    private static $primaryKey = 'gibbonAlertLevelID';
+
+    const LEVEL_HIGH = 1;
```
SKuipers, line 38: I like the readability here that using constants adds 👍

SKuipers (changes requested):
Looking great, thanks for tidying up yet another generic function 😃 I've added a couple minor suggested changes, but otherwise looks ready to merge.

src/UI/Dashboard/ParentDashboard.php
yookoala: Removed.

src/Domain/System/AlertLevelGateway.php
yookoala: Changed.

src/Domain/System/AlertLevelGateway.php
yookoala: :-)

SKuipers (approved):
Thanks, looking great, going to merge it in and test it out 👍

PR #1664, opened 2022-10-03
School: replace getTerms with SchoolYearTermGateway methods

yookoala:
**Description**
* Create SchoolYearTermGateway::getBySchoolYear() and SchoolYearTermGateway::mapNames().
* Mark getTerms() deprecated.
* Rewrite all getTerms() usage with SchoolYearTermGateway::getBySchoolYear() and SchoolYearTermGateway::mapNames().

**Motivation and Context**

**How Has This Been Tested?**
* CI Environment


SKuipers (approved):
Thanks, going to merge and test 👍

SKuipers:
Thanks Koala! On testing and reviewing the code, I noticed that most of the calls to the old function were using the term list to get the names based on a list of IDs, so I've made a couple post-merge tweaks to simplify the code a bit by adding a getTermNamesByID method, and renamed the getBySchoolYear to selectTermsBySchoolYear, which returns a result set making fetchKetPair a quick way to map IDs to names. For gateways the naming convention I've been using is queryThing for critera-based queries resulting in a DataSet, selectThing for selects that return a Result, and getThing for selects that return a plain array (a single record or list of primitive values).

PR #1667, opened 2022-10-03
Students: replace getHighestMedicalRisk with MedicalGateway method

yookoala:
**Description**
* Create MedicalGateway::getHighestMedicalRisk() method.
* Mark getHighestMedicalRisk() deprecated.
* Replace all getHighestMedicalRisk() calls with MedicalGateway::getHighestMedicalRisk() calls.

**Motivation and Context**

**How Has This Been Tested?**
* CI environment


src/Domain/Students/MedicalGateway.php
```diff
@@ -116,4 +116,25 @@ public function getMedicalConditionByID($gibbonPersonMedicalConditionID)
...
+     */
+    public function getHighestMedicalRisk(int $gibbonPersonID)
+    {
+        $sql = 'SELECT * FROM gibbonPersonMedical JOIN gibbonPersonMedicalCondition ON (gibbonPersonMedical.gibbonPersonMedicalID=gibbonPersonMedicalCondition.gibbonPersonMedicalID) JOIN gibbonAlertLevel ON (gibbonPersonMedicalCondition.gibbonAlertLevelID=gibbonAlertLevel.gibbonAlertLevelID) WHERE gibbonPersonID=:gibbonPersonID ORDER BY gibbonAlertLevel.sequenceNumber DESC';
+        $result = $this->db()->select($sql, ['gibbonPersonID' => $gibbonPersonID]);
+        if ($result->isNotEmpty()) {
+            $row = $result->fetch();
+            return $row;
+        }
+        return false;
```
SKuipers, line 138: The Connection & Result classes can internally handle whether to fetch or return an empty set using the selectOne method. This case ensures that the return type is always an array, simply an empty array for non-results, making it easier than having to test for false vs an array. Eg:

```suggestion
        return $this->db()->selectOne($sql, ['gibbonPersonID' => $gibbonPersonID]);
```

yookoala: I didn't take this path because this would change the contract of the function. I tried to write the new method as a drop-in replacement so module developers wouldn't need to change much of their code (I don't really know how this is used by other modules). 

I'm OK to change the code along with [student_view_details.php](https://github.com/GibbonEdu/core/pull/1667/files#diff-0a3f66851febb1b1ded942b1b63fb7f232e5cc8a6fc7e576696f87673bc72a41) if this is preferable.

SKuipers: That's a fair point. I'm okay with changing the contract in this case, ideally for the better. If we can ensure that all Gateway methods return an array as a result, either empty or full, it will likely keep things cleaner and easier for developers in the future. Then using empty() and !empty() rather than == false or == '' in the code should help to keep it backwards compatible.

yookoala: > That means empty() is essentially the concise equivalent to `!isset($var) || $var == false`. 

I've never noticed this about [empty()](https://www.php.net/manual/en/function.empty.php). Will change accordingly.

yookoala: Change has been made.

SKuipers (approved):
Thanks, looking good! 👍

SKuipers (approved):

PR #1668, opened 2022-10-03
Students: fix minor code style issue

yookoala:
**Description**
* Minor rewrite to fix issue of IDE type inference.

**Motivation and Context**
* Previously written code calls `setCriteria` on a `RendererInterface`, which does not have a `setCriteria` method.
* Causes IDE static analysis warning.

**How Has This Been Tested?**
* Locally with VSCode and [Intelephense](https://intelephense.com/)

**Screenshots**

Before
Screenshot from 2022-10-04 00-12-57

After
Screenshot from 2022-10-04 00-13-08

### DIFF
--- a/modules/Students/student_view_details.php
+++ b/modules/Students/student_view_details.php
@@ -504,8 +504,9 @@ if (isActionAccessible($guid, $connection2, '/modules/Students/student_view_deta
 
                             $view = $_GET['view'] ?? 'grid';
                             if ($view == 'grid') {
-                                $table->setRenderer(new GridView($container->get('twig')));
-                                $table->getRenderer()->setCriteria($criteria);
+                                /** @var GridView */
+                                $gridView = $container->get(GridView::class);
+                                $table->setRenderer($gridView->setCriteria($criteria));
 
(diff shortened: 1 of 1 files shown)


modules/Students/student_view_details.php
```diff
@@ -504,8 +504,10 @@
 
                             $view = $_GET['view'] ?? 'grid';
                             if ($view == 'grid') {
-                                $table->setRenderer(new GridView($container->get('twig')));
-                                $table->getRenderer()->setCriteria($criteria);
+                                $table->setRenderer(
+                                    (new GridView($container->get('twig')))
```
SKuipers, line 508: Thanks, well caught! Are we able to use the $container->get() method here to create the class, that way the constructor remains flexible and we can avoid using the new keyword to create classes.

SKuipers (changes requested):

yookoala:
Issue resolved.

modules/Students/student_view_details.php
yookoala: Rewritten.

PR #1671, opened 2022-10-05
FormGroups: replace getNextFormGroupID with method

yookoala:
**Description**
* Mark getNextFormGroupID() deprecated.
* Add FormGroupGateway::getNextFormGroupID() as a replacement.
* **Note:** unable to find any occurance of usage of getNextFormGroupID().

**Motivation and Context**
* Replace functions.php function with gateway method.
* Not sure if this is the way to handle because I found no usage of it. Alternatively, can remove it altogether.

**How Has This Been Tested?**
* CI environment.


src/Domain/FormGroups/FormGroupGateway.php
```diff
@@ -143,4 +143,26 @@ public function getFormGroupByID($gibbonFormGroupID)
...
+    public function getNextFormGroupID($gibbonFormGroupID)
+    {
+        $sql = 'SELECT gibbonFormGroupIDNext FROM gibbonFormGroup WHERE gibbonFormGroupID=:gibbonFormGroupID';
+        $result = $this->db()->select($sql, [
+            'gibbonFormGroupID' => $gibbonFormGroupID,
+        ]);
+        if ($result->rowCount() == 1) {
+            return $result->fetchColumn() ?? false;
+        }
+        return false;
```
SKuipers, line 166:
```suggestion
        return $this->db()->selectOne($sql, [
            'gibbonFormGroupID' => $gibbonFormGroupID,
        ]);
```

yookoala: The method here returns either the (int) ID or FALSE if not found. It's not exactly match what selectOne returns (an array of all columns or an empty array).

SKuipers (approved):
👍

PR #1673, opened 2022-10-07
System: new AccessManager for access control

yookoala:
**Description**
* A new method ModuleGateway::selectRoleModuleActionNames().
* A new class Action to replace simple string $address parameter in isActionAccessible().
* A new class AccessManager to provide feature of isActionAccessible().
* Make AccessManager accessible globally as $accessManager in index.php routed scripts.
* Mark isActionAccessible deprecated.

**Motivation and Context**
* To replace isActionAccessible() function.
* Remove unnecessary use of global variables, $guid and $connection2.

**How Has This Been Tested?**
* CI environment.
* Locally.


yookoala:
I've changed the Action usage to use `Action::fromRoute()` instead of `new Action()` to initiate the instance. Should be more flexible for future changes.

SKuipers:
Closed, as this class is now part of the core. Thanks again for laying the groundwork Koala 🙏

PR #1674, opened 2022-10-07
Timetable Admin: refactor to use Url class for paths and url generation

yookoala:
**Description**
* Use `__DIR__` and require_once for a more stable behaviour when routed through index.php.
* Refactor to use Url class for paths and url generation.

**Motivation and Context**
* Remove the use of `$session->get('absoluteURL')` from the module.
* Fix all issue when using Timetable Admin through index.php routing. Some feature was still relying on direct routing on the script path.
* Pave way for future router routing efforts.

**How Has This Been Tested?**
* CI environment.

### DIFF
--- a/modules/Timetable Admin/courseEnrolment_manage.php
+++ b/modules/Timetable Admin/courseEnrolment_manage.php
@@ -22,6 +22,7 @@ use Gibbon\Services\Format;
 use Gibbon\Tables\DataTable;
 use Gibbon\Forms\DatabaseFormFactory;
 use Gibbon\Domain\Timetable\CourseGateway;
+use Gibbon\Http\Url;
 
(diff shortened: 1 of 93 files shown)


SKuipers (changes requested):
Hi Koala, you've done a lot of great work here, I think unfortunatly any of the forms using Url::fromModuleRoute will face the same problem highlighted in the Crowd Assessment PR: the links are prepending the index.php/ front-controller route rather than a direct link to the script. Curious to hear what your thoughts are in terms of handling those two different types of routes in the URL class: a new method, or a chained method modifier?

yookoala:
I think in long term, all the corner cases of the problem scripts need to be refactor so they works within index.php routing environment. Only then we can really migrate to path based routing (instead of script file based).

I'm going to use the Crowd Assessment module as a test bed to try things out first. It is one of the smallest modules in Gibbon. Things should be simpler with it than the bigger ones. And before that, I'll need a better test coverage to the module.

PR #1675, opened 2022-10-10
System: remove obsoleted include in index.php

yookoala:
**Description**
* Remove Messenger module's moduleFunctions.php include statement from index.php. According to comment, the statement was needed because the function getNotificationTray() needs it. The function was removed since v18.

**Motivation and Context**
Checked "TODO"s in the code base and noticed this.

**How Has This Been Tested?**
* CI environment


yookoala:
Depends on #1676.

PR #1679, opened 2022-10-24
System: translation function fallback

yookoala:
**Description**
* If a global $gibbon->locale does not exists, the function __ can fallback on Locale::formatString. Prevent unit test using __ to rely on specific global $gibbon setup.

**Motivation and Context**
* Prevent error if the global variable is not setup properly.

**How Has This Been Tested?**
* Locally.
* CI environment.


yookoala:
Not sure if this is the way to go. Hence tagged as a discussion.

SKuipers (approved):
Looks like a nice fallback condition, I can certainly see where this is useful in testing.

PR #1682, opened 2022-10-26
Replace getRoleCategory with gateway method

yookoala:
**Description**
* Migrate getRoleCategory() function to RoleGateway::getRoleCategory().
* Mark getRoleCategory() as deprecated.
* Rewrite all usage of getRoleCategory() into RoleGateway::getRoleCategory().

**Motivation and Context**
* Rewrite database relevant functions into gateway methods.

**How Has This Been Tested?**
* CI environment.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -818,7 +818,20 @@ function getHighestGroupedAction($guid, $address, $connection2)
     return $output;
 }
 
-//Returns the category of the specified role
+/**
+ * Returns the category of the specified role.
+ *
+ * Deprecated. Use RoleGateway::getRoleCategory() instead.
+ *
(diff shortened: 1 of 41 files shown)


SKuipers (approved):
This is looking great, I'm going to merge and test. Another old function deprecated 🎉 

I've checked the session data, and it looks like gibbonRoleIDCurrentCategory is already loaded at the same time as gibbonRoleIDCurrent, so one tweak I'll make post-merge is to remove all the cases where `$session->get('gibbonRoleIDCurrent')` is being used to check getCurrentCategory, since this is already available in the session and doesn't need extra database queries. Should help clean up even more old code.

PR #1683, opened 2022-10-26
Services: replace emailBodyConvert calls with Format::emailBodyToPlain

yookoala:
**Description**
* Add Format::emailBodyConvert to replace emailBodyConvert() in
  functions.php.
* Mark emailBodyConvert() deprecated.
* Replace emailBodyConvert() calls with Format::emailBodyToPlain().

**Motivation and Context**
* Deprecated global functions in favour of class methods.

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -47,7 +47,19 @@ function getIPAddress() {
     return $return;
 }
 
-//Convert an HTML email body into a plain text email body
+/**
+ * Convert an HTML email body into a plain text email body.
+ *
+ * Deprecated. Use \Gibbon\Services\Format::emailBodyToPlain() instead.
+ *
+ * @deprecated v25
+ * @version v12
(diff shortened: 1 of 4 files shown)


SKuipers (approved):
Thanks Koala for even more deprecations! In this case, the [Mailer](https://github.com/GibbonEdu/core/blob/v25.0.00/src/Comms/Mailer.php) class has a renderBody method which handles the strip-tags automatically, and these two scripts appear to still be using the old-style mailing. I'm going to merge this PR and update them, as I think we can keep the deprecation and then use the Mailer to handle these two cases, without needing the Format class this time.

yookoala:
That would be an even cleaner solution. Thanks for merging.

PR #1684, opened 2022-10-26
System: refactor birthday calculation logics and functions

yookoala:
**Description**
* Mark daysUntilNextBirthday and is_leap_year deprecated.
* A new Format::daysUntilNextBirthday protected method to replace daysUntilNextBirthday() in functions.php. The new method is re-implemented with PHP core's \DateTime.

**Motivation and Context**
* The function daysUntilNextBirthday() is only used by the Format class.
* The function is_leap_year() is only used by daysUntilNextBirthday().
* The logics are based on explicit leap year calculation, which is also done internally with PHP's own DateTime implementation.
* Functions that do not needed to be public should be kept as implementation details within a class.

**How Has This Been Tested?**
* Locally.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -237,7 +237,20 @@ function archiveNotification($connection2, $guid, $gibbonPersonID, $actionLink)
     return $return;
 }
 
-//Accepts birthday in mysql date (YYYY-MM-DD) ;
+/**
+ * Calculate the number of days before next birthday.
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
Thanks! Amazing to see more and more of these old functions being refactored 😃

PR #1685, opened 2022-10-26
Refactor password policy

yookoala:
**Description**
* A new abstraction PasswordPolicy to describe password policies with.
* Have CoreServiceProvider setup to initialize PasswordPolicy from settings gateway.
* Rewrite doesPasswordMatchPolicy and getPasswordPolicy with PasswordPolicy. Also mark both as deprecated.

**Motivation and Context**
* To simplify code related to password policies.
* Pave way to better error message with password policies.

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -21,7 +21,7 @@ use Gibbon\Http\Url;
 use Gibbon\Forms\Form;
 use Gibbon\Services\Format;
 use Gibbon\Contracts\Comms\Mailer;
-use Gibbon\Data\PasswordPolicies;
+use Gibbon\Data\PasswordPolicy;
 use Gibbon\Domain\Students\MedicalGateway;
 use Gibbon\Domain\System\AlertLevelGateway;
 use Gibbon\Domain\System\LogGateway;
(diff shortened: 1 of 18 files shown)


SKuipers (changes requested):
This is looking great, I love to see more code with unit tests 😃 My only suggestion is a very minor semantic change to rename the class `PasswordPolicies` to `PasswordPolicy`, to change it from plural to singular (since a policy already implies multiple rules within itself).

yookoala:
@SKuipers: Renamed.

SKuipers (approved):
Thanks, looking great! Historically, the `randomPassword` function has generated passwords that did not match the policy. Now that we have this class, when we go to refactor randomPassword, we could add a `PasswordPolicy::generatePassword` method, so we could ensure that the passwords generated satisfied the policy 😄

yookoala:
> Thanks, looking great! Historically, the `randomPassword` function has generated passwords that did not match the policy. Now that we have this class, when we go to refactor randomPassword, we could add a `PasswordPolicy::generatePassword` method, so we could ensure that the passwords generated satisfied the policy smile

Sounds like a cool addition to the class. Will work on that.

PR #1687, opened 2022-10-26
Install: fix phpdocs of Installer

yookoala:
**Description**
* Fix return type in phpdocs for getConnection() and getPDO().

**Motivation and Context**
* Incorrect return type documented for getConnection() and getPDO().

**How Has This Been Tested?**
* Locally in VSCode.


SKuipers (approved):
👍

PR #1669, opened 2022-10-05
Refactor: SchoolYearGateway methods to replace functions.php relevant functions

yookoala:
**Description**
* Mark setCurrentSchoolYear() deprecated.
* Add SchoolYearGateway::setCurrentSchoolYear() that provdes the exact feature.
* Rewrite Finance module usage of setCurrentSchoolYear() to SchoolYearGateway::setCurrentSchoolYear().
* Mark getPreviousSchoolYearID() and getNextSchoolYearID() deprecated.
* Replace all getPreviousSchoolYearByID() calls with SchoolYearGateway::getPreviousSchoolYearByID().
* Replace all getNextSchoolYearByID() calls with SchoolYearGateway::getNextSchoolYearByID().

**Motivation and Context**
* Replace functions.php functions with relevant SchoolYearGateway method.

**How Has This Been Tested?**
* CI environment.

PR #1688, opened 2022-10-28
System: move school year session setup to SessionFactory

SKuipers:
After merging #1669 I got looking at the code and there appeared to be some school year setup code that ended up scattered throughout CLI scripts and other locations, rather than in one initialization step.

@yookoala I've moved the setCurrentSchoolYear method out of the SchoolYearGateway into the SessionFactory, as this prevents the Gateway class from needing session dependencies, and puts more of the session initialization code in the SessionFactory, which is then called during the Gibbon initializeCore method. This enabled using a single try-catch for exception handling, instead of needing several throughout the system.

This also has fixed an esoteric bug where the current school year wasn't available directly after logging out unless setCurrentSchoolYear was called manually, so I've been able to remove the code from the sidebar.

**Motivation and Context**
Enabled us to delete even more old code and centralize the session initialization.

**How Has This Been Tested?**
Locally.

### DIFF
--- a/cli/attendance_dailyIncompleteEmail.php
+++ b/cli/attendance_dailyIncompleteEmail.php
@@ -21,27 +21,9 @@ use Gibbon\Domain\System\SettingGateway;
 use Gibbon\Services\Format;
 use Gibbon\Comms\NotificationEvent;
 use Gibbon\Comms\NotificationSender;
-use Gibbon\Domain\School\SchoolYearGateway;
 use Gibbon\Domain\System\NotificationGateway;
 
 require getcwd().'/../gibbon.php';
-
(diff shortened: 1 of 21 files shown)


yookoala (dismissed):

yookoala:
Totally cool change. If fact it's more reasonable than my changes.

You might want to mention this in the phpdocs of the deprecated setCurrentSchoolYear() function.
https://github.com/GibbonEdu/core/blob/4a0472077bff1687fd9da87556a1218d6f9d76c4/functions.php#L1311-L1326

SKuipers:
Thanks! Well caught, I've updated the docs in functions.php 👍

PR #1691, opened 2022-10-30
System Admin / Authentication: LDAP Implementation

powertashton:
This is an almost complete LDAP implementation using the Aura Auth library that is already used in Gibbon for other forms of authentication.

Still TODO for this PR:
 - Use LDAP domain and bind details from system admin configuration (rather than the current method that uses hardcoded details (that have been redacted))
 - Make it possible to set the LDAP login as the default for login attempts from the normal login form, potentially allowing it to be controlled on a role/individual user basis rather than just a blanket setting across all users. 
 - Better exception catching

TODO in future PRs:
- Enable creation of accounts upon new user LDAP credentials, mapping different fields to each other and probably basing roles off of OUs

Initially made this for use at my old workplace to facilitate easy login to Gibbon, as users were used to only ever needing to use their AD credentials that were linked to most other systems used in the school. Still some configuration that needs working on before merging but wanted to make sure this project doesn't just get lost in my piles of files.

### DIFF
--- a/login.php
+++ b/login.php
@@ -28,6 +28,7 @@ use Gibbon\Auth\Adapter\OAuthAdapterInterface;
 use Gibbon\Auth\Adapter\OAuthGoogleAdapter;
 use Gibbon\Auth\Adapter\OAuthMicrosoftAdapter;
 use Gibbon\Auth\Adapter\OAuthGenericAdapter;
+use Gibbon\Auth\Adapter\LDAPAdapter;
 use Gibbon\Domain\System\LogGateway;
 use League\Container\Exception\NotFoundException;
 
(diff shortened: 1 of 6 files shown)


rossdotparker:
This looks very cool! I'll leave it to Sandra to review.

SKuipers:
Hi @powertashton Looks like this PR is shaping up nicely! It appears to be a nice clean implementation so-far, give a shout with any questions as you get the PR ready to review & merge 😄

SKuipers:
Hi @powertashton I've added the "in development" label to this PR for now, since it's in draft mode. Feel free to give a shout if you'd like anything reviewed or tested at this point. I think there will definitely be some schools very happy to see this in the core 😃

PR #1696, opened 2022-11-03
Deprecate nl2brr

yookoala:
**Description**
* Replace nl2brr() calls with nl2br().
* Mark nl2brr() as deprecated.
* Changed behaviour: "\r\n" would have been converted to 2 `<br/>` will now be converted to 1 single `<br/>`.

**Motivation and Context**
* From the source code, nl2brr turns \r\n, \r and \n into HTML `<br/>` tag. And PHP's native nl2br does that and also converts \n\r.
So it seems nl2brr specifically single out \n\r and not handle it. This doesn't seem to really matter in most cases.
* nl2br() is a C native implementation while nl2brr() is written with preg_replace(), which is much slower.
* Reduce unnecessary function would reduce effort if handling the code base.

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -1328,7 +1328,7 @@ function getModuleCategory($address, $connection2)
  *
  * @deprecated v25
  *             This happens in SessionFactory::setCurrentSchoolYear instead,
- *             which is called by Core::initializeCore, so shouldn't need to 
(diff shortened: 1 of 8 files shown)


SKuipers (approved):
Lovely to see more old functions deprecated 😄

PR #1700, opened 2022-11-04
System: deprecate randomPassword

yookoala:
**Description**
* Add a new PasswordPolicy::generate method to generate a random password that's conforming to the rules defined for the instance.
* Rewrite all randomPassword usages with PasswordPolicy::generate().

**Motivation and Context**
* Fix issue when randomly generated password doesn't meet site password policy settings.
* Deprecate usage of randomPassword.

**How Has This Been Tested?**
* Locally.
* CI environment.

### DIFF
--- a/cli/schoolAdmin_parentWeeklyEmailSummary.php
+++ b/cli/schoolAdmin_parentWeeklyEmailSummary.php
@@ -21,6 +21,7 @@ use Gibbon\View\View;
 use Gibbon\Services\Format;
 use Gibbon\Contracts\Comms\Mailer;
 use Gibbon\Comms\NotificationEvent;
+use Gibbon\Data\PasswordPolicy;
 use Gibbon\Domain\User\FamilyGateway;
 use Gibbon\Domain\System\SettingGateway;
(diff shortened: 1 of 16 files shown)


SKuipers (approved):
Interesting, I didn't realize how many places in the codebase were using randomPassword to generate random strings, rather than the password itself 🤔 I guess in the future this may be better handled by a function that creates strings for the purpose of being uuids. 

I appreciate the time you've put into this, and don't want to leave the PR lingering, so will approve it and test it out, and continue to think on how to deal with uuids vs passwords 🤔

yookoala:
@SKuipers: Just found an issue with phpstan and pushed 1 more commit. Please check and approve this again. Thanks.

SKuipers:
> @SKuipers: Just found an issue with phpstan and pushed 1 more commit. Please check and approve this again. Thanks.

Excellent, love that static analysis 😁 Thanks for catching this!

SKuipers (approved):

SKuipers:
🎉 PR number 1700 🎉

PR #1701, opened 2022-11-04
Data: explicit type-cast in ImportType

yookoala:
**Description**
* Use explicit type-cast on ImportType::getField() return values in various methods of ImportType to supress type warning or error when using those variables.

**Motivation and Context**
* PHP 8+ deprecates the use of implicit type casting on some core function parameters. Explicit type-casting will solve the potential issues in the future.

**How Has This Been Tested?**
* Checks with VSCode IDE.


SKuipers (approved):
👍

PR #1704, opened 2022-11-11
GitHub Actions: make phpstan (level 0) test essential

yookoala:
**Description**
* phpstan test will be essential to pass for every CI test.
* That would also make the test essential for all PRs.

**Motivation and Context**
* phpstan starts to catch some issue in PRs that format test don't.
* Level 0 is mild and easy to handle for developers. Should be a nice start.

**How Has This Been Tested?**
* CI environment.


SKuipers (approved):
Looks like a nice step forward, thanks 😃

PR #1705, opened 2022-11-14
Attendance: fix attendance for classes timetabled multiple times on the same day

SKuipers:
Currently, if a class is timetabled for multiple lessons on the same day, it only has one attendance record, and submitting the attendance again with update the existing attendance. There was also a bug where students with timetable exceptions from one class were also not being listed in other classes on the same day.

This PR adds a gibbonTTDayRowClassID field to the attendance log for class attendance to track exactly which timetable period the attendance was taken for. It is backwards compatible, handling null values in the field by default, and has been prefilled in the background for lesson plan attendance.

When a class is timetabled multiple times, the user will be prompted to select which timetable period they are taking attendance for.

Also updates the fromArray method on input classes to enable specifying the name, value and groupBy fields, the same way that other MultipleOptionsTrait inputs do. This makes it easier to pass in a set of data that has more than just value => name key pairs.

**Motivation and Context**
Fixes issues with multiple classes on the same date and adds more relational data to the attendance log to enable future queries to refine their results.

**How Has This Been Tested?**
Locally.

**Screenshots**
Screen Shot 2022-11-14 at 2 45 37 PM


rossdotparker (approved):

rossdotparker:
This one was a pleasure to review and merge, as it addresses an edge case that we've been aware of for a while, combined with a new bug. Nice : )

PR #1713, opened 2023-01-16
Selection component bug fix

leofreitas:
**Motivation and Context**
When a user uses another language some functionalities have a strange behavior. For example, the feature that make form fields hidden while form type is not selected. Because in the code the value should be 'Please select...'

**How Has This Been Tested?**
Local

**Screenshots**
Captura de Tela 2023-01-15 às 22 34 58

### DIFF
--- a/src/Forms/Input/Select.php
+++ b/src/Forms/Input/Select.php
@@ -198,13 +198,13 @@ class Select extends Input
         $output .= '<select '.$this->getAttributeString().'>';
 
         if ($this->getRequired() && $this->placeholder === '') {
-            $this->placeholder(__('Please select...'));
+            $this->placeholder('Please select...');
         }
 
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Thanks, I can see where this would cause errors for different languages. I did a quick search and there are some other areas that are checking for Please select... as a translated string, I'm going to merge your PR and follow up with some related fixes. Thanks again!! 😃

SKuipers:
Hi Leo, here is the follow up commit: https://github.com/GibbonEdu/core/commit/d782cc1c04b5f4f1724cd4b59be90358d7eb8808

There were two additional cases that I fixed, one where the script was manually defining the placeholder string, which wasn't necessary as this is handled internally. The other was the visibility triggers whenNot() that needed to now check using the untranslated string. Have a quick look and let me know if you spot any issues. Thanks!

PR #1715, opened 2023-01-26
Fix page sizes

Devon-Olivier:
**Description**

Change 'letter' pageSize to 'LETTER' for TCPDF Renderer to work for US Letter page size.

**Motivation and Context**

When "US Letter" page size is chosen in `Home > Reports > Template Builder > Edit Template `  for a template using the TCPDF renderer, generated reports still remain in the A4 page size. Things work as expected with the mPDF renderer.

**How Has This Been Tested?**

Enviroment:
Gibbon v25.0.00
PHP 8.1.13
Mariadb 10.10.2

These are run in docker containers on a linux host.

**Screenshots**


SKuipers:
Hi Devon-Olivier, huge thanks for submitting this PR! My apologies for the wait, I generally try to look at PRs sooner than this, but it's been a busy Chinese New Year season here, right on the heels of a busy release.

Your changes are looking good, overall. I wonder about existing data though, as the string is being compared case sensitively, so it looks like it wouldn't catch any existing templates with the page size already lowercase.

An alternative approach is to use the `strtoupper()` function on the value that is passed into the TCPDF renderer, this way regardless of the source case, it will end up the correct case. In the template, the `|upper` twig filter could be added, to ensure that it's always checking the same case,  for example `{% if pageSize|upper == 'LETTER' %}`. Do you want to have a look at making these changes, then I'm happy to merge this one in. Thanks! 😃

modules/Reports/templates/preview.twig.html
```diff
@@ -70,31 +74,30 @@ <h1 class="text-center text-lg text-gray-700 font-bold my-4">{{ name }}</h1>
     </details>
     {% endif %}
 
-    {% if pageSize == 'letter' %}
+    {% if pageSize == 'LETTER' %}
```
SKuipers, line 77:
```suggestion
    {% if pageSize|upper == 'LETTER' %}
```

modules/Reports/src/Renderer/MpdfRenderer.php
```diff
@@ -191,11 +191,11 @@ protected function setupDocument()
 
         $config = [
             'mode' => 'utf-8',
-            'format' => $this->template->getData('pageSize', 'A4') == 'letter' ? [215.9, 279.4] : [210, 297],
+            'format' => $this->template->getData('pageSize', 'A4') == 'LETTER' ? [215.9, 279.4] : [210, 297],
```
SKuipers, line 194:
```suggestion
            'format' => strtoupper($this->template->getData('pageSize', 'A4')) == 'LETTER' ? [215.9, 279.4] : [210, 297],
```

modules/Reports/src/Renderer/TcpdfRenderer.php
```diff
@@ -170,30 +170,30 @@ protected function renderSectionToHTML(ReportSection &$section, ReportData &$rep
...
+            return include 'templates/' . $section->template;
         }
 
         return '';
     }
 
     protected function setupDocument()
     {
-        $this->pdf = new ReportTCPDF($this->template->getData('orientation', 'P'), 'mm', $this->template->getData('pageSize', 'a4'), true, 'UTF-8', false);
+        $this->pdf = new ReportTCPDF($this->template->getData('orientation', 'P'), 'mm', $this->template->getData('pageSize', 'A4'), true, 'UTF-8', false);
```
SKuipers, line 190:
```suggestion
        $this->pdf = new ReportTCPDF($this->template->getData('orientation', 'P'), 'mm', strtoupper($this->template->getData('pageSize', 'A4')), true, 'UTF-8', false);
```
```suggestion
        $this->pdf = new ReportTCPDF($this->template->getData('orientation', 'P'), 'mm', $this->template->getData('pageSize', 'A4'), true, 'UTF-8', false);
```

SKuipers (changes requested):

PR #1717, opened 2023-03-09
added Botswana Pula as currency option

kalungia:
There are schools and tuition centers using Gibbon in Botswana and every time I have to add the Botswana currency if am installing or updating.

If the Botswana Currency is in Gibbon I belive their will be more schools that use it.

**How Has This Been Tested?**
Yes

**Screenshots**


rossdotparker:
Thanks for this @kalungia. I'm going to make a slight adjustment, moving the entry up one in the list to keep the alphabetisation consistent, and I'll add a changelog entry too.

rossdotparker:
You can see these two changes in the following commit: https://github.com/GibbonEdu/core/commit/f8ad4255c098d98f282cc5fb1f4dd5fca25a692a

Issue #1689, opened 2022-10-28
Meta: retire old functions in functions.php


yookoala:
### Problem

There are 3 types of functions in functions.php:

1. Old style database reading functions with old school `$guid` and/or `$connection2` (\PDO) as parameter.
2. Utility functions for formatting (e.g. email body sanitation) or other helper functions (e.g. resolving client IP address).
3. Shortcut for accessing core service feature (e.g. __(), __n(), __m()).

The first type of function cause usage of all the global `$guid` and `$connection2`, which is an anti-pattern when transiting to service container architecture.

### Proposed Solution

1. The first type should be rewritten into Gateway service method (or, in some case, rewritten as service provider).
2. The second type can be rewritten as static methods of some utility classes.
3. The third type can be left as-is for now.

In the rewrite, all obsoleted functions should not be removed for now. Instead, they are to be marked deprecated in their phpdocs doc-block. Modern IDE (e.g. VSCode) will identify the deprecated function calls and warn developers.

### Alternatives

Leaving things as-is.

SKuipers:
Amazing, thanks for getting this initiative underway and documenting it so well 😃

SKuipers:
All the deprecated functions from v25 have been removed in v30, and this has been picked up again as a v30 development goal, to keep removing functions.

PR #1752, opened 2023-08-31
Refactor: Deprecate msort

yookoala:
**Description**
* Rewrite all usages of msort in core to usort with documentation.
* Properly document msort with phpdoc and mark it deprecated.

**Motivation and Context**
* Part of the effort to retire old functions. See #1689.
* msort is a really old function in Gibbon (First introduced to git repo in 2286c9b0b, which was when Gibbon first migrated to git). I'm not sure if it is used elsewhere in modules (although I've searched all modules to found no reference to it). Might be safer to make it deprecated here.

**How Has This Been Tested?**
* Locally.
* CI.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -666,7 +666,15 @@ function getTerms($connection2, $gibbonSchoolYearID, $short = false)
     return $output;
 }
 
-//Array sort for multidimensional arrays
+/**
+ * Array sort for multidimensional arrays.
+ *
+ * Deprecated in favor of native usort.
+ *
+ * @since 2013
+ * @version v12.0.00
+ * @deprecated v26.0.00
+ */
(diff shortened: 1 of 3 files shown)


yookoala:
Test failed because arrow function used in this is a PHP 7.4 feature, but the composer PHP version config is now at 7.3. phpstan installed does not support PHP 7.4 syntax, hence the strange error message. #1753 will fix this issue.

SKuipers:
Thanks, can you rerun the checks now that #1753 is merged? It usually gives me the option to, but I'm not seeing it this time.

yookoala:
@SKuipers: All tests are past now :-)

SKuipers (approved):
Thanks! 👍

PR #1753, opened 2023-08-31
composer: update php version in config

yookoala:
**Description**
* Update php version config in composer to 7.4. We have used 7.4 as the lowest supported version for a while now. We just haven't updated composer.json accordingly.

* Update libraries to their latest version possible according to the new PHP version requirement.

* Added new "php-http/discovery" composer extension as a dependency to the latest "omnipay/common" version (v3.2.1).

**Motivation and Context**
* Working on #1752 involves rewriting old code with [Arrow Functions](https://www.php.net/manual/en/functions.arrow.php) syntax, which was introduced in PHP 7.4.
* phpstan reports an error that doesn't make sense:
  > ```
  > Error: Function getResourcesTagCloud not found.
  > ------ --------------------------------------------------------------------- 
  > Line   UI/Components/Sidebar.php                                            
  > ------ --------------------------------------------------------------------- 
  > 615    Function getResourcesTagCloud not found.                             
  >         💡 Learn more at https://phpstan.org/user-guide/discovering-symbols  
  > ------ --------------------------------------------------------------------- 
  > ```
* Investigation reveals that by updating the PHP version in composer config, composer will update phpstan to a more appropriate version for PHP 7.4 support. Then the arrow function wouldn't cause an issue anymore.

**How Has This Been Tested?**
* Locally.
* CI Platform.

**Screenshots**

Before
image

After
image

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -38,7 +38,7 @@
         "test:codesniffer": "Check code complies with PSR-2 standards"
     },
     "require": {
-        "php" : "^7.3",
+        "php" : "^7.4",
         "ext-curl": "*",
         "ext-intl": "*",
         "ext-mbstring": "*",
@@ -82,7 +82,10 @@
     },
     "config": {
         "platform": {
-            "php": "7.3"
+            "php": "7.4"
+        },
+        "allow-plugins": {
+            "php-http/discovery": true
         }
     },
     "autoload": {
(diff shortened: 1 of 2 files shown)


SKuipers (approved):
Thanks, looks good. I've updated the v26 minimum requirement to PHP 7.4 in version.php as well.

PR #1759, opened 2023-11-21
Fix CI environments language setup

yookoala:
**Description**
* Add en_GB and en_US to CI environment.

**Motivation and Context**
* Found out that the setlocale was incorrect in testing environment, which makes test results unstable for some of the test I'd want to add (in another PR).

**How Has This Been Tested?**
* GitHub Workflows
* GitLab CI


SKuipers (approved):
Thanks for discovering this and fixing it! 😃

Issue #1644, opened 2022-09-26
Discussion: Semi-auto installation


yookoala:
### Problem

* If you install a Gibbon copy, Gibbon requires to have no configuration file in the directory. So before installation, the config file need to be gone.
* For dockerized applications, we don't usually expose the root directory to volume (i.e. stored outside of docker). But unless a folder is exposed as a volume, files inside it will be removed on reboot. Clearly the config file need to be there.
* Usual practice includes:
  * Have config file located in a subfolder instead. So the folder can be exposed to volume.
  * Have a config file exposing variables to environment variables. So docker user can set them when initializing the docker.
* To fit this paradigm, we need to update Gibbon installer to accept installation with config file already in the folder.

### Proposed Solution

1. Following the [12 factors](https://12factor.net/) approach, accepts configurations from environment variables.
2. Rewrite installer to not block installation when the config file exists.
3. That means the installer should determine the installation steps by detecting the installation status:
   1. First step: setup the config file if none exists.
   2. Second step: install the database tables if the tables are not there.
   3. Third step: Install the user configurations if it was not yet done.
   4. Forth step: Installation is done. Some congratulation message and link to home page.
5. If installer detects a config file is there, it will detect which step the installation is. It will help user to proceed to the appropriate step:
   1. If the installation is at the second step, proceeding to second step and install database directly (without database information form)
   2. If the installation is at the third step, proceeding to user configuration form.
   3. If the installation is already at the forth step (installation done), it will show some warning and prompt user to clear the database / config file.
7. To prevent unexpected access to installer, the first visit to installer would create some kind of lock-file with the first visit browser information (e.g. IP address, browser, OS, cookie secret) to ensure only 1 person can do the installation within a reasonable time (e.g. 30 minutes).

### Alternatives

* Migrating the config file into some sub-folder and expose the folder as a docker volume. But that means a rewrite of bootstrap procedure of Gibbon.

### Additional Context

* [Use volumes | Docker Docs](https://docs.docker.com/storage/volumes/)
* [The Twelve-Factor App](https://12factor.net/)

PR #1762, opened 2023-12-04
Add Docker / docker-compose option to deploy gibbon

mrestivill:
**Description**
The main idea is to simplify deployment of gibbon on servers using containers.

**Motivation and Context**
This will also be deployed using docker-compose.yml using the original mysql/mariadb instance with users/password.

**How Has This Been Tested?**
I've tested the deployment of the application using this command:
```bash
docker-compose -p gibbon up -d
```
Accessing the port  `http://localhost:80` also accessing to mysql using `localhost:3306`.


yookoala:
Some thoughts:

1. Exposing the root folder of Gibbon would defeat the purpose of containerize the site. I think the default image should only mount the uploads folder as volume, if possible.
2. For better deployment flow, the hardcoded config file should be replaced with environment variables (i.e. dot env). Also the installer should support a semi-auto installation (i.e. with a docker specific `config.php` that takes environment variables, and the environment variable set correctly, the installer should start installing without user input of database information).

fvlasie:
We may want to expose modules and themes folders too.....

mrestivill:
I've changed the base docker image to apache, to simplify generation of the docker image.
I've also added an entrypoint file to activate debug mode based on environment variables.

mrestivill:
Next steps will be identify what variables we need on the runtime of the container to write config.php file.

mrestivill:
I've added the 4 variables from config.php, I generated the file during startup when no config.php is found, also deletes install directory.

mrestivill:
The config.php is generated with the environment variables but the database is empty, should I use gibbon.sql to initializate the database or some parameters on config.php to autoinitialize the app.

Thanks

mrestivill:
I just added the sql file in docker-compose to initialize the bbdd.

mrestivill:
@yookoala Hi, Could you have a look on the PR, thanks

kleo:
Hi @mrestivill I've recently tried it out and this skips the installer. 

While system requirements is ok and database settings are already setup using environment variables and docker-gibbon-entrypoint,
for this setup how would I go on to create my administrator user account?

yookoala:
@mrestivill: Working on a headless setup for this. May have something to add later. Sorry for the hold up (very busy these days).

@SKuipers if you find this good enough, please merge without my update. Thanks.

yookoala:
> Thank you so much @mrestivill for your work on this and @yookoala for your feedback and ideas. My sincere apologies for the lack of a response, it's not for a lack of appreciation, but simply a lack of capacity. Its been an incredibly busy school year and I'm a full-time teacher on top of maintaining Gibbon. My school has, luckily, recently hired a university intern to help build capacity for Gibbon development. I am going to ask him to take a look at this and test it out for us in the coming week. It is interesting to see that it skips the installer. I wonder, longer term, what we could develop to help support headless installation?

I think it would help to have a semi-auto installation, such that user can:
1. Setup a docker with existing database config injected through environment variable.
2. Visit the docker and follow up the installation (i.e. site name, account, email, and other settings).

## Why?

Site name and site informations should probably be filled by hand (unless we're talking about really massive deployment / some automatic Gibbon SaaS hosting). But database information should probably be automated or at least scriptable by a degree, especially for a docker container. 

A containerized setup will probably have everything setup with scripts. The database server hostname, username, password would probably be setup automatically in the process. There is no need for the user to find out then manually fill-in those information by hands. Also, those information might be dynamically changed later.

In general, application following [12 factor methodology](https://12factor.net/) will be easier to automatically deploy and scale up. I think it will be good to make Gibbon docker feasible for such setup ([I also described here before](https://github.com/GibbonEdu/core/issues/1644)).

## How Things would Work?

My thinking is like:
1. Add a config (template) variant / example that would setup things with [getenv](https://www.php.net/manual/en/function.getenv.php).
2. Have an extra flag variable in config that accepts automatic setup. Installer will be run if
   (a) database config is correctly setup; and
   (b) there are no Gibbon tables in the database; and
   (c) the "acceptHeadlessSetup" flag (or something like this) is explicitly set to "true".
3. After setup, the flag can explicitly set to "false" to prevent installer fired up for some sort of database failure.
4. Maybe add support to [.env file](https://www.dotenv.org/docs/security/env.html) for development. (Optional)

mrestivill:
The current proposed docker-compose.yml creates the database, and also executes the ./gibbon.sql file when the database is empty (see /docker-entrypoint-initdb.d folder used inside mysql docker documentation: https://hub.docker.com/_/mysql) . The admin user will be the default one inside gibbon.sql. Docker-compose also supports [the use of .env file](https://docs.docker.com/compose/environment-variables/set-environment-variables/#substitute-with-an-env-file) as a source of variables. The init script can be configured to obtain, as a variable, the admin user and generate a random password printed inside the logs. For now the password can be changed inside the application by admin (using default credentials) during setup as a poststep.

yookoala:
The current docker-composer.yml setup is not bad. What I'm proposing __should not block this PR__. It is meant to be improvements to the docker file.

Ideally, the docker file should not depend on docker-compose. It should be useful on its own. A headless installation with environment variable support customized deployments different from one single default setup. And thus I think is more ideal than the current proposed changes. All that said, I don't yet have the time to finish my headless installer PR.

mrestivill:
This Dockerfile does not depend on docker-compose. Docker-compose is an example of deployment with a standard database docker and gibbonEdu docker.

AbbadiAhmad:
Thank you for this initiative. 
I have already created simple and working docker compose that is running for production. but yours is much more professional :)

What I want to share with you is the development docker where I use it for testing new update or developing new modules using VS code. 
I use xdebug library in connection with VScode.
maybe you can add this as option to run the docker container in testing mode. which allow developer to debug the code or view the database when needed.
[gibbon_docker_testEnv.zip](https://github.com/user-attachments/files/18102639/gibbon_docker_testEnv.zip)

tendomart (commented):
Thanks for this PR, been wondering why a wonderful system like Gibbon is not containerised..Looking out for this

ali-ichk:
Hi @mrestivill @yookoala @fvlasie @AbbadiAhmad 

Thank you for developing the Docker feature for Gibbon. I'm eager to test it, as I believe it will be a valuable tool for the Gibbon community. I attempted to use Docker Desktop with the .zip file from the latest commit of the pull request, but I encountered issues running the Dockerfile. Since I already have Gibbon installed on my machine, I'm unsure if the problem stems from a conflict or server settings. As a Docker beginner, I'm finding it difficult to diagnose the issue.

To facilitate testing and deployment, could you provide a detailed, step-by-step guide on deploying Gibbon using the Dockerfile? This documentation would be extremely beneficial, allowing me to test the feature on different machines and configurations. Thank you in advance for your time. Looking forward to test it out in the coming weeks.

SKuipers:
Thank you @mrestivill for your past contributions and suggestions, Docker compose files have now been added to Gibbon in v31

PR #1765, opened 2023-12-08
Activities: add a Left status for activity enrolment and notification events for enrolment changes

SKuipers:
This PR improved the activity enrolment process with the following changes:

 - Notification events for activity enrolment changes (add, edit, delete)
 - Adds a Left status to activity enrolment to track past students in activities
 - Refactors several scripts and adds bulk actions to the activity enrolment page

**Motivation and Context**
Makes it easier to track changes and historical data in activity enrolment.

**How Has This Been Tested?**
Tested locally.


rossdotparker:
This looks a lot like refactoring to me! Nice to see another table get kicked into touch. I'm surprised that after all this time we are only just adding getUserDetails() to the user gateway!

PR #1767, opened 2023-12-13
System: remove raw exception message output from the interface

SKuipers:
Currently, there's many areas of the code where exception messages are output directly to the user-facing interface. In particular, when handling SQL queries. This is problematic for two reasons, one in that the Connection class already internally handles exceptions so this is a duplication of code, and another because exception messages shouldn't be displayed raw to the user except when in Development mode. Instead, in Production systems they should be logged, which is another set of refactoring we have planned for the future.

**Description**
Removes all raw $e->getMessage() output from try/catch blocks. Scripts that use $e->getMessage() internally and properly handle the exception message have not been changed.

**Motivation and Context**
Refactoring and security.

**How Has This Been Tested?**
Locally and CI.

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -28,6 +28,7 @@ v27.0.00
 
     Tweaks & Additions
         System: automatically hyperlink any urls included in Custom Field descriptions
+        System: removed raw exception message output from the interface
         Activities: added notification events for activity enrolment changes
         Activities: added a Left status to activity enrolment
         Activities: added bulk actions to the activity enrolment page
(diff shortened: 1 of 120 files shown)


yookoala:
Should the error message be logged somewhere instead of just ignored?

rossdotparker:
I've merged this for now, but @yookoala raises an interesting point. Would these errors not show up in the MySQL or PHP logs?

SKuipers:
Thanks Koala and Ross. Yes, the Connection class already catches and handles exception output for queries, so these were for the most part duplicates from old code. For any of the other ones, on our list is to go through and clean up all these try/catch statements, and will add proper logging to the cases that need it. Focused on cleanup only for this PR.

PR #1768, opened 2023-12-13
System: refactoring error messages in activities and attendance.

felixOlesen:
**Description**
Refactoring error messages in activities and attendance modules.

**Motivation and Context**
Refactoring echo-ed html `<div class='error'>` into `$page->addError();`.

**How Has This Been Tested?**
Local testing.


rossdotparker:
Looking good Felix. Welcome to the team!

PR #1776, opened 2024-01-10
Refactoring $gibbon->session variable to $session

felixOlesen:
**Description**
Replaced session call through $gibbon->session to $session. `$gibbon->session` is now deprecated. Changes were made taking php functions and classes into account such that `$gibbon->$session` should no longer exist in the core code base.

**How Has This Been Tested?**
Tested locally.

### DIFF
--- a/modules/Attendance/src/AttendanceView.php
+++ b/modules/Attendance/src/AttendanceView.php
@@ -78,7 +78,7 @@ class AttendanceView
         $this->settingGateway = $settingGateway;
 
         // Collect the current IDs of the user
-        $this->userRoleIDs = array_filter(array_column($gibbon->session->get('gibbonRoleIDAll'), 0));
+        $this->userRoleIDs = array_filter(array_column($session->get('gibbonRoleIDAll'), 0));
         // Get the current date
         $currentDate = (isset($_GET['currentDate'])) ? Format::dateConvert($_GET['currentDate']) : date('Y-m-d');
 
--- a/modules/Attendance/src/AttendanceView.php
+++ b/modules/Attendance/src/AttendanceView.php
@@ -74,11 +74,11 @@ class AttendanceView
      */
     public function __construct(Core $gibbon, Connection $pdo, SettingGateway $settingGateway)
     {
+        global $session;
         $this->pdo = $pdo;
         $this->settingGateway = $settingGateway;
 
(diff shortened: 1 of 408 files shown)


SKuipers:
Thanks Felix, looks like the CI robots have caught some issues when testing, see if you can trace them back to the files in question: https://github.com/GibbonEdu/core/actions/runs/7471398671/job/20331619554?pr=1776

yookoala:
@felixOlesen: Some advice in debugging.

At the [summary page](https://github.com/GibbonEdu/core/actions/runs/7471398671?pr=1776) of each job, you may find artifacts produced in run time:

圖片

Within the zip file (folder: `tests/_output`), there are files generated for the failed tests. If you have a local development server running Gibbon at http://127.0.0.1:8888, you may even see the proper render output when opening the HTML files. 

The files are the page(s) when CI test failed to get the expected result. There are usually some error message(s) there useful for debugging.

圖片

If not, you may check the render output to find something missing (or found an unexpected page rendered).

modules/Attendance/src/AttendanceView.php
```diff
@@ -78,6 +78,7 @@ public function __construct(Core $gibbon, Connection $pdo, SettingGateway $setti
         $this->settingGateway = $settingGateway;
 
         // Collect the current IDs of the user
+        global $session;
```
yookoala, line 81: For readability sake, the defacto standard is to place the "global" statement at the top of a function.

modules/Markbook/src/MarkbookView.php
```diff
@@ -130,6 +130,7 @@ public function __construct(Core $gibbon, Connection $pdo, $gibbonCourseClassID,
         $this->pdo = $pdo;
 
         $this->gibbonCourseClassID = $gibbonCourseClassID;
+        global $session;
```
yookoala, line 133: Same as above.

For readability sake, the defacto standard is to place the "global" statement at the top of a function.

yookoala:
Looks good to me :-)

SKuipers (approved):
Looks good, I can see a couple moduleFunctions.php files that were edited, but in each case, the $session variable has already been added as a global, so no issues there. Thanks for another big tedious PR to improve the code consistency. Now, on to some more interesting tasks 😄

PR #1781, opened 2024-01-23
Updated Format file to remove in-baked styling

felixOlesen:
**Description**
Updated Format file to remove in-baked styling of Format::photo uses. The styling will now only be used as a default param.

**Motivation and Context**
This change allows for more modular usage of the Format::photo method stylistically.

**How Has This Been Tested?**
Tested Locally

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -40,6 +40,7 @@ v27.0.00
         School Admin: added the ability to toggle facilities between active and inactive, removing inactive facilities from general use
         Students: add an Upcoming Students option to the Students by House report
         Timetable: added basic information table to the View Timetable by Facility report
+        System: Updated Format file to remove in-baked styling
 
     Bug Fixes
         System: fixed PHP 8+ compatibility in CustomFieldIDs migration file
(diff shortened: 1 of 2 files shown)


yookoala:
This would change the `$class` parameter behaviour from appending to overriding classes.

I've just search. We have 3 usages in core that uses the `Format::photo()` method:

* ./modules/Library/report_studentBorrowingRecord.php:97:
    ```php
                return Format::photo($item['imageLocation'], 75);
    ```
* ./modules/Library/library_browse.php:237:
    ```php
        return Format::photo($item['imageLocation']);
    ```
* ./modules/Students/student_view_details.php:2296:
    ```php
                                return Format::photo($item['imageLocation'], 75);
    ```

None of these use the `$class` parameter to override the photo class. Not sure if there is any usage in the modules.

SKuipers:
Thanks @yookoala for checking! I didn't recall this method even existing, so did some looking around in the code. It appears like this method was created by a community member, as part of Library refactoring work they contributed to the core, and has only been used in that context, and not in additional modules. Given this limited use, it looks like we're good to proceed in cleaning up this method a bit, as it also has some duplicate code that was intended only for the Format::userPhoto method.

SKuipers (approved):
👍 Thanks Felix

PR #1782, opened 2024-01-24
Added Status Filter to Coverage and Absence Summary Forms

felixOlesen:
**Description**
Added status filter to Staff Absence and Coverage Summaries to display only 'Full' status staff by default. This was done by implementing a `$status` variable that get's changed by a checkbox form input. The filter then gets applied by the value the checkbox sets the `$status` variable to.

**Motivation and Context**
This fix was required to de-clutter the default viewing experience for the tables in each summary page.

**How Has This Been Tested?**
Tested Locally

**Screenshots**
Screenshot 2024-01-24 at 10 48 12
Screenshot 2024-01-24 at 10 48 22
Screenshot 2024-01-24 at 12 57 15
Screenshot 2024-01-24 at 12 57 27

### DIFF
--- a/src/Domain/Staff/StaffAbsenceGateway.php
+++ b/src/Domain/Staff/StaffAbsenceGateway.php
@@ -173,7 +173,6 @@ class StaffAbsenceGateway extends QueryableGateway implements ScrubbableGateway
             ->leftJoin('gibbonPerson AS coverage', 'gibbonStaffCoverage.gibbonPersonIDCoverage=coverage.gibbonPersonID')
             ->where('gibbonStaffAbsenceDate.date BETWEEN :dateStart AND :dateEnd')
             ->where("gibbonStaffAbsence.status = 'Approved'")
-            //->where("gibbonPerson.status = 'Full'")
             ->bindValue('dateStart', $dateStart)
             ->bindValue('dateEnd', $dateEnd)
             ->groupBy(['gibbonStaffAbsence.gibbonStaffAbsenceID', 'gibbonStaffCoverageDate.gibbonStaffCoverageDateID']);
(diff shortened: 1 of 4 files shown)


src/Domain/Staff/StaffAbsenceGateway.php
```diff
@@ -173,7 +173,7 @@ public function queryApprovedAbsencesByDateRange(QueryCriteria $criteria, $dateS
             ->leftJoin('gibbonPerson AS coverage', 'gibbonStaffCoverage.gibbonPersonIDCoverage=coverage.gibbonPersonID')
             ->where('gibbonStaffAbsenceDate.date BETWEEN :dateStart AND :dateEnd')
             ->where("gibbonStaffAbsence.status = 'Approved'")
-            ->where("gibbonPerson.status = 'Full'")
+            //->where("gibbonPerson.status = 'Full'")
```
SKuipers, line 176: It's okay to remove this line, rather than commenting it out. We'll never remember why it was left there in the future, and if we do ever need it, all the past code is in the git commit history : )

SKuipers, line 176:
```suggestion
```

SKuipers (commented):
For consistency, in Staff Absence Summary, are the left staff also highlighted in red, the same as Staff Absence Summary? I can't see from the screenshot if this is the case, but if not, this would be good to add. Then, everything else looks good and ready to merge.

felixOlesen:
Ah I see, they aren't highlighted in red for the absence summary so I'll add that change in.

felixOlesen:
Screenshot 2024-01-24 at 15 55 57

All sorted 👍

SKuipers (approved):
Looking good, thanks for making these changes 👍

SKuipers:
Tested out perfectly in production.

PR #1786, opened 2024-02-19
System: Refactoring code and editing error message by adding blank state

ali-ichk:
**Description**
Edited the error message by adding the blank slides (step-1)


SKuipers (approved):
Thanks, I've reviewed this and it is looking good. For the $pageCount variable changes, I had a look at those three scripts, and the silly thing is that those variables are declared but not even used, I think they're leftover from a previous refactoring. I'll add a note to Trello to cleanup these lines of code.

SKuipers:
Actually, in order to resolve the conflicts with the null coalesce changes, I just cleaned up these lines, so nothing else needs done here (for now).

PR #1787, opened 2024-02-19
GitHub Actions: CI to check PHP Notice

yookoala:
**Description**
* CI to check and output all php "Notice" in server_log generated in test. More hint to improve code with.
* Allow the step to fail at the moment. Only provide verbose hint but not enforcing developer to clear all notices.

**Motivation and Context**
Sometimes changes in code would generate notice that are expose incorrect assumption and hence potential problem, for example:

圖片

(from: https://github.com/GibbonEdu/core/actions/runs/7955085821)

Extracting and showing verbose log about it can provide vital hints for code improvements.

**How Has This Been Tested?**

Tested the base script locally against the log with notice:
圖片

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -137,6 +137,21 @@ jobs:
           fi
         continue-on-error: ${{ matrix.experimental }}
 
+      - name: Check PHP notice in server_log
+        run: |
+          # Check number of notice.
+          export NUM_NOTICE=$(grep --perl-regexp 'Notice: ' ./tests/_logs/server_log | wc -l)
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Thanks Koala for the change, and for so quickly catching those unintentional errors in the previous refactoring. Where should I be looking to see these notices before merging a PR, will they show up in the interface under "All checks have passed", or would I need to go into details and check the logs for the action? Thanks!

yookoala:
@SKuipers: The GitHub Action for CI Test will now show a new "Check PHP notice in server_log" section that will have the log message extracted. But currently the step will be skipped if the test failed. Need one more update to have it visible. 

See #1790.

PR #1792, opened 2024-02-20
System: Refactoring code to retrieve object from container

ali-ichk:
**Description**

Refactoring code for changing the notification object from manual to container.


SKuipers (approved):
👍

PR #1796, opened 2024-02-23
Staff: creating cli script for sending first aid expiry notifications

ali-ichk:
**Description**

Staff: Created a cli to send notification informing the names of the staff whose First Aid Qualification are due to expire within three months.


SKuipers (changes requested):
Thanks Ali, this is looking good. It looks like this branch may have pulled in some changes from your other recent PRs. Can you fetch and pull your v27.0.00 locally to make sure it's up to date, then rebase this branch onto v27, and force-push your rebased branch. This should clean up the commit changes. Thanks!

PR #1800, opened 2024-02-28
FileUploader & addQuick: Bug fix so that filenames include Chinese cha…

ali-ichk:
**Description**
FileUploader & addQuick: Fixed the big so that filenames that include Chinese character can also be uploaded in the messages in Gibbons.


SKuipers (changes requested):
Thanks for the fix, this is looking good. Can you also add a CHANGELOG.txt entry, something like `System: fixed bug in FileUploader that did not allow Chinese and Latin characters`

SKuipers (approved):

PR #1801, opened 2024-02-29
Bug Fix: Fixed the bug of messageWall widget and messageWall messages…

ali-ichk:
**Description**

System: Fixed the bug of messageWall widget overflowing the table and larger messages in messageWall goibg beyond the table. Added the feature of overflow behaviour.

### DIFF
--- a/src/Domain/Messenger/MessengerGateway.php
+++ b/src/Domain/Messenger/MessengerGateway.php
@@ -791,7 +791,6 @@ class MessengerGateway extends QueryableGateway
                     ->width('650px')
                     ->addClass('align-top overflow-x-scroll max-w-lg')
                     ->format(function ($message) {
-                        //$output = '<div style="max-width:600px;overflow-x: scroll;">';
                         $output = '<h3 style="margin-top: 3px">';
                         $output .= $this->validator->sanitizePlainText($message['subject']);
                         $output .= '</h3>';
--- a/src/Domain/Messenger/MessengerGateway.php
+++ b/src/Domain/Messenger/MessengerGateway.php
@@ -788,9 +788,10 @@ class MessengerGateway extends QueryableGateway
                     });
 
                 $table->addColumn('message', __('Message'))
-                    ->width('640px')
-                    ->addClass('align-top')
+                    ->width('650px')
+                    ->addClass('align-top overflow-x-scroll max-w-lg')
                     ->format(function ($message) {
+                        //$output = '<div style="max-width:600px;overflow-x: scroll;">';
                         $output = '<h3 style="margin-top: 3px">';
                         $output .= $this->validator->sanitizePlainText($message['subject']);
                         $output .= '</h3>';
@@ -798,6 +799,7 @@ class MessengerGateway extends QueryableGateway
                         $output .= '</p>';
                         $output .= $this->validator->sanitizeRichText($message['details']);
                         $output .= '</p>';
+                        //$output .= '</div>';
 
                         return $output;
                     });
--- a/src/Domain/Messenger/MessengerGateway.php
+++ b/src/Domain/Messenger/MessengerGateway.php
@@ -788,7 +788,7 @@ class MessengerGateway extends QueryableGateway
                     });
 
                 $table->addColumn('message', __('Message'))
-                    ->width('650px')
+                    ->width('640px')
                     ->addClass('align-top overflow-x-scroll max-w-lg')
                     ->format(function ($message) {
                         $output = '<h3 style="margin-top: 3px">';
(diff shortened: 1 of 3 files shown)


src/UI/Components/Sidebar.php
```diff
@@ -326,7 +326,7 @@ public function getOutput()
                             return $group;
                         }, []);
 
-                        echo '<div class="column-no-break">';
+                        echo '<div class="overflow-x-scroll max-w-xs" >';
```
SKuipers, line 329: It looks like the column-no-break class has been removed, however this needs to be there for mobile responsiveness in the sidebar. 
```suggestion
                        echo '<div class="column-no-break overflow-x-scroll max-w-xs" >';
```

src/UI/Components/Sidebar.php
```diff
@@ -346,7 +346,7 @@ public function getOutput()
                             } elseif (count($unpinnedMessages) == 2) {
                                 $height = 197;
                             }
-                            echo "<table id='messageWallWidget' style='width: 100%; height: ".$height."px; border: 1px solid grey; padding: 6px; background-color: #eeeeee'>";
+                            echo "<table id='messageWallWidget' style='width: w-auto; height: ".$height."px; border: 1px solid grey; padding: 6px; background-color: #eeeeee'>";
```
SKuipers, line 349: w-auto is a class name rather than a CSS value, so it won't work here in this case.

```suggestion
                            echo "<table id='messageWallWidget' style='width: 100%; height: ".$height."px; border: 1px solid grey; padding: 6px; background-color: #eeeeee'>";
```

src/UI/Components/Sidebar.php
```diff
@@ -360,7 +360,7 @@ public function getOutput()
                                 echo "<tr id='messageWall".$pos."' style='z-index: 1;'>";
                                 echo "<td style='font-size: 95%; letter-spacing: 85%;'>";
                                 //Image
-                                $style = "style='width: 45px; height: 60px; float: right; margin-left: 6px; border: 1px solid black'";
+                                $style = "style='width: 45px; height: 50px; float: right; margin-left: 6px; border: 1px solid black'";
```
SKuipers, line 363: Was the height change here intentional? Since we're working with a width-related bug, it doesn't look like the height should have changed.
```suggestion
                                $style = "style='width: 45px; height: 60px; float: right; margin-left: 6px; border: 1px solid black'";
```

SKuipers (changes requested):
Hi Ali, this fix is looking good overall, with a few items I've flagged in code review. Can you also add a message to the CHANGELOG.txt file, something like `Messenger: fixed visual bug when messages are wider than the message wall`

Thanks!

src/Domain/Messenger/MessengerGateway.php
```diff
@@ -788,16 +788,18 @@ public function getMessages(string $mode = 'print', string $date = '')
                     });
 
                 $table->addColumn('message', __('Message'))
-                    ->width('640px')
-                    ->addClass('align-top')
+                    ->width('650px')
+                    ->addClass('align-top overflow-x-scroll max-w-lg')
                     ->format(function ($message) {
+                        //$output = '<div style="max-width:600px;overflow-x: scroll;">';
```
SKuipers, line 794: Ah, just noticed that this commented out code seems to have made its way into the commit. Can you clean these up? Thanks.
```suggestion
```

src/Domain/Messenger/MessengerGateway.php
```diff
@@ -788,16 +788,18 @@ public function getMessages(string $mode = 'print', string $date = '')
...
                         $output = '<h3 style="margin-top: 3px">';
                         $output .= $this->validator->sanitizePlainText($message['subject']);
                         $output .= '</h3>';
 
                         $output .= '</p>';
                         $output .= $this->validator->sanitizeRichText($message['details']);
                         $output .= '</p>';
+                        //$output .= '</div>';
```
SKuipers, line 802:
```suggestion
```

src/Domain/Messenger/MessengerGateway.php
```diff
@@ -788,16 +788,18 @@ public function getMessages(string $mode = 'print', string $date = '')
                     });
 
                 $table->addColumn('message', __('Message'))
-                    ->width('640px')
-                    ->addClass('align-top')
+                    ->width('650px')
```
SKuipers, line 791:
```suggestion
                    ->width('640px')
```

SKuipers (changes requested):

SKuipers (approved):

PR #1802, opened 2024-03-01
Activities: gateway class refactoring part 1

ali-ichk:
**Description**
Activities: Refactoring code in the activities module to update the SELECT queries into respective gateways.


modules/Activities/activities_attendance.php
```diff
@@ -73,40 +75,27 @@
...
+    $attendanceResult = $container->get(ActivityAttendanceGateway::class)->selectStudentAttendanceByActivity($gibbonActivityID);
 
     // Gather the existing attendance data (by date and not index, should the time slots change)
     $sessionAttendanceData = array();
 
     while ($attendance = $attendanceResult->fetch()) {
         $sessionAttendanceData[ $attendance['date'] ] = array(
             'data' => (!empty($attendance['attendance'])) ? unserialize($attendance['attendance']) : array(),
-            'info' => sprintf(__('Recorded at %1$s on %2$s by %3$s.'), substr($attendance['timestampTaken'], 11), Format::date(substr($attendance['timestampTaken'], 0, 10)), Format::name('', $attendance['preferredName'], $attendance['surname'], 'Staff', false, true)),
+            'info' => __('Recorded at %1$s on %2$s by %3$s.'), substr($attendance['timestampTaken'], 11), Format::date(substr($attendance['timestampTaken'], 0, 10)), Format::name('', $attendance['preferredName'], $attendance['surname'], 'Staff', false, true),
```
SKuipers, line 98: It looks like this line has removed the sprintf, but not updated the syntax, so this string will likely cause an error. In this case, lets restore the sprintf and refactor strings like this in a future PR.

ali-ichk: noted and add the sprintf back

PR #1803, opened 2024-03-02
Change magnifier/plus icon to eye icon

fvlasie:
Some users have reported being confused about how to open lessons and suggested that the magnifying glass/+ symbol did not seem intuitive. I am suggesting changing the symbol to an eye. This eye symbol is free from FontAwesome.


SKuipers:
Hmmm, interesting. This is a tricky one. I can see where the magnifier might be confusing at first, but we also now have many thousands of users already using Gibbon who have gotten used to the icon, so switching it now could also cause confusion 😆

SKuipers:
I've tested out the suggested icon change on Ross but he found it a bit too strange of a change 😆 I think you'd be welcome to make this change on your systems, and we do ideally hope to update to a font or svg based icon set in the future. But, for now, although we appreciate the suggestion, I think it may be too much of a visual change for the current Gibbon team to adapt to 😅

PR #1806, opened 2024-03-07
Messenger- Created a new index to reduce the run time for nested quer…

ali-ichk:
**Description**
Messenger: In the messenger_manage page, the page takes a considerable time to load the table especially because of the column "Recipients" which are the target audience for the message. In order to reduce the run time, we have created a new key index which reduces the runtime in the nested loop in the code. However, this is not a permanent solution and we will need to come back to it later for a better solution.

**Testing**
Locally, with production data


SKuipers (approved):
Thanks Ali, looking good, I'm going to merge and test it out.

SKuipers (approved):

PR #1807, opened 2024-03-11
Improved Library Shelf Automatic Generation, Updating, and Shuffling

felixOlesen:
**Description**
Library: added automatic shelf generation by search term, automatic shelf updates, and auto-shuffling.
- Added automatic-updates to auto-filled shelves every time the library_browse.php page is reloaded.
- Added an auto-fill category to fill by Search Term. (Available after selecting "Print Publication" in the Catalog Type field)
- Added a new column in the table for auto-shuffling on reload, this can be toggled by the form inputs in the Edit and Add pages for specific shelves.

**Motivation and Context**
These improvements were implemented in response to some feedback given by a librarian on useful auto-fill options.

**How Has This Been Tested?**
Tested Locally

**Screenshots**
Screenshot 2024-03-11 at 20 58 53


modules/Library/library_browse.php
```diff
@@ -161,8 +161,16 @@
             ->fromPOST();
 
         foreach($activeShelves as $shelf) {
-            $libraryShelves[$shelf['gibbonLibraryShelfID']] = $itemGateway->queryItemsByShelf($shelf['gibbonLibraryShelfID'], $criteria);
+            if($shelf['type'] == 'Automatic') {
+                $itemGateway->updateShelfContents($shelf['gibbonLibraryShelfID'], $shelf['field'], $shelf['fieldValue']);
```
SKuipers, line 165: It looks like this is happening every time someone loads the page. I've looked at the code and wonder what logic it's using to prevent duplicates in the shelf? 🤔

SKuipers (commented):
Amazing, thanks Felix for continuing to polish of this feature set with a very useful addition of automatically updating shelves, and search terms for shelves. The changes are looking good, just one question I've added about the updateShelfContents method. Looking forward to merging and checking it out 😃

modules/Library/library_browse.php
felixOlesen: Hi Sandra, it's my pleasure to help polish out the stuff I couldn't finish before leaving. As for the updateShelfContents gateway I tried to take duplicates into account with the:

`WHERE gibbonLibraryItemID NOT IN
            (SELECT gibbonLibraryItemID 
             FROM gibbonLibraryShelfItem
            WHERE gibbonLibraryShelfID = :libraryShelfID)`

Portion of the SQL statement, which seems to work from my testing. Although I'm not sure if two people were to simultaneously reload what the outcome would be. It is also slower if you have shelves that have content in the thousands range. If you have a couple pointers I'd appreciate the help on improving this 👍

SKuipers: Aha, I see that now. Interesting. I could see where it might end up causing some load on the server if the table is constantly being written to in bulk on each page load 🤔 I wonder if the Automatic shelves are always a live query, and don't write to the table at all, instead returning their list of books for each page load. I'm happy to tinker with that idea post-merge, but curious to hear what you'd think.

PR #1808, opened 2024-03-13
Behaviour: create multiple records and show all the students involved

ali-ichk:
**Description**

Behaviour: When a behaviour incident involves multiple students, it is shown who are the students involved , so that the related incidents can easily be referenced.


modules/Behaviour/behaviour_manage_edit.php
```diff
@@ -52,22 +53,18 @@
...
         $gibbonYearGroupID = $_GET['gibbonYearGroupID'] ?? '';
         $type = $_GET['type'] ?? '';
-
+        
         //Check if gibbonBehaviourID specified
         $gibbonBehaviourID = $_GET['gibbonBehaviourID'] ?? '';
         if ($gibbonBehaviourID == '') {
             $page->addError(__('You have not specified one or more required parameters.'));
         } else {
             try {
```
SKuipers, line 62: We can remove the try/catch statement anytime a query is coming from a gateway.

modules/Behaviour/behaviour_manage_edit.php
```diff
@@ -100,18 +97,42 @@
                 $form->addHiddenValue('address', "/modules/Behaviour/behaviour_manage_add.php");
                 $form->addRow()->addClass('hidden')->addHeading('Step 1', __('Step 1'));
 
+
+                //To show other students involved in the incident
+                if(!empty($values['gibbonMultiIncidentID'])) {
+
+                    $resultList = $container->get(BehaviourGateway::class)->selectMultipleStudentsOfOneIncident($values['gibbonMultiIncidentID']);
```
SKuipers, line 104: When a gateway is used multiple times in a script, it can be more efficient to add a `$behaviourGateway = $container->get(BehaviourGateway::class);` at the top of the script, and then reuse the gateway with that variable.

SKuipers, line 104: Using result variables is part of the legacy style that we're refactoring out. In this case, a nice concise way to fetch the students would be: 
```suggestion
                    $students = $container->get(BehaviourGateway::class)->selectMultipleStudentsOfOneIncident($values['gibbonMultiIncidentID'])->fetchAll();
```
The result object internally handles checking for no results and will return an empty array from the fetchAll() method.

src/Domain/Behaviour/BehaviourGateway.php
```diff
@@ -230,4 +230,28 @@ public function queryBehaviourRecordsByPerson(QueryCriteria $criteria, $gibbonSc
 
             return $this->runQuery($query, $criteria);
     }
+
+    public function getBehaviourID($gibbonSchoolYearID, $gibbonBehaviourID)
```
SKuipers, line 234: It looks like, since more than just an ID is being retrieved, a good name for this method might be `getBehaviourDetails` along with `getBehaviourDetailsByCreator` below.

src/Domain/Behaviour/BehaviourGateway.php
```diff
@@ -230,4 +230,28 @@ public function queryBehaviourRecordsByPerson(QueryCriteria $criteria, $gibbonSc
...
+    {
+        $data = ['gibbonSchoolYearID' => $gibbonSchoolYearID, 'gibbonBehaviourID' => $gibbonBehaviourID];
+        $sql = 'SELECT gibbonBehaviour.*, student.surname AS surnameStudent, student.preferredName AS preferredNameStudent, creator.surname AS surnameCreator, creator.preferredName AS preferredNameCreator, creator.title FROM gibbonBehaviour JOIN gibbonPerson AS student ON (gibbonBehaviour.gibbonPersonID=student.gibbonPersonID) JOIN gibbonPerson AS creator ON (gibbonBehaviour.gibbonPersonIDCreator=creator.gibbonPersonID) WHERE gibbonSchoolYearID=:gibbonSchoolYearID AND gibbonBehaviourID=:gibbonBehaviourID ORDER BY date DESC';
+        
+        return $this->db()->select($sql, $data);
```
SKuipers, line 239: Since these are both get methods, it's best to use selectOne, for consistency. It also makes the code in the script easier, as you don't need to call fetch on it.

modules/Behaviour/behaviour_manage_edit.php
```diff
@@ -52,22 +53,18 @@
...
         //Check if gibbonBehaviourID specified
         $gibbonBehaviourID = $_GET['gibbonBehaviourID'] ?? '';
         if ($gibbonBehaviourID == '') {
             $page->addError(__('You have not specified one or more required parameters.'));
         } else {
             try {
                 if ($highestAction == 'Manage Behaviour Records_all') {
-                    $data = array('gibbonSchoolYearID' => $session->get('gibbonSchoolYearID'), 'gibbonBehaviourID' => $gibbonBehaviourID);
-                    $sql = 'SELECT gibbonBehaviour.*, student.surname AS surnameStudent, student.preferredName AS preferredNameStudent, creator.surname AS surnameCreator, creator.preferredName AS preferredNameCreator, creator.title FROM gibbonBehaviour JOIN gibbonPerson AS student ON (gibbonBehaviour.gibbonPersonID=student.gibbonPersonID) JOIN gibbonPerson AS creator ON (gibbonBehaviour.gibbonPersonIDCreator=creator.gibbonPersonID) WHERE gibbonSchoolYearID=:gibbonSchoolYearID AND gibbonBehaviourID=:gibbonBehaviourID ORDER BY date DESC';
+                    $result = $container->get(BehaviourGateway::class)->getBehaviourID($session->get('gibbonSchoolYearID'), $gibbonBehaviourID);
```
SKuipers, line 64: Thanks for refactoring these as part of the new feature 👍

modules/Behaviour/behaviour_manage_edit.php
```diff
@@ -100,18 +97,42 @@
...
+                //Other Students
+                if (!empty($values['gibbonMultiIncidentID'])) {
+                $row = $form->addRow();
+                    $row->addLabel('otherStudents0', __('Other Students Involved'));
+                    $col = $row->addColumn()->addClass('flex flex-col');
+
+                    foreach ($students as $i => $student) {
+
+                        if ($student['gibbonPersonID'] != $values['gibbonPersonID']) {
+                        $url = $session->get('absoluteURL').'/index.php?q=/modules/Students/student_view_details.php&gibbonPersonID='.$student['gibbonPersonID'].'&subpage=Behaviour&search=&allStudents=&sort=surname,preferredName';
```
SKuipers, line 124: This is correct, and you'll see lots of links in the codebase like this. We are refactoring towards using the Url class, which is a PHP standard now. Can you see about implementing this link with the Url::fromModuleRoute method, eg: `$url = Url::fromModuleRoute('Behaviour', 'behaviour_manage_edit')->withQueryParams([... etc.]);`

SKuipers (changes requested):
Hi Ali, you're off to a great start, having worked out the back-end and front-end changes. I've added some comments and suggested changes, all just habits and consistency that comes with practice in the Gibbon codebase. Thanks!

modules/Behaviour/behaviour_manage_edit.php
```diff
@@ -100,18 +97,42 @@
                 $form->addHiddenValue('address', "/modules/Behaviour/behaviour_manage_add.php");
                 $form->addRow()->addClass('hidden')->addHeading('Step 1', __('Step 1'));
 
+
+                //To show other students involved in the incident
+                if(!empty($values['gibbonMultiIncidentID'])) {
+
+                    $students = $container->get(BehaviourGateway::class)->selectMultipleStudentsOfOneIncident($values['gibbonMultiIncidentID'])->fetchAll();
+                }
+                $students = $resultList->fetchAll();
```
SKuipers, line 106: It looks like you've auto accepted the suggested change, but haven't updated the rest of the file in context. Be sure to check this out, as this code will now create an error.

SKuipers (approved):
Thanks Ali, these changes are looking good, I'm going to merge them in and test them out. For your other behaviour-related changes in the works, be sure to rebase your branch onto the updated v27 branch before PRing them. Thanks!

SKuipers (changes requested):
Almost there, except some lines from CHANGEDB and Changelog have been inadvertently deleted in this branch. Can you add them back in? Thanks.

SKuipers (approved):

PR #1810, opened 2024-03-14
Adding comment log to behaviour incidents

ali-ichk:
**Description**

Behaviour: Added a follow-up log to behaviour incidents so that any staff member can add their comment, and all comments are displayed in a chronological order with names and timestamps. Also created a new table in the DB called - gibbonBehaviourFollowUp to store all the follow-up comments for respective behaviour reports.

### DIFF
--- a/CHANGEDB.php
+++ b/CHANGEDB.php
@@ -800,6 +800,7 @@ INSERT INTO `gibbonPermission` (`gibbonRoleID` ,`gibbonActionID`) VALUES ('001',
 CREATE TABLE `gibbonLibraryShelfItem` ( `gibbonLibraryShelfItemID` INT(6) UNSIGNED ZEROFILL NOT NULL AUTO_INCREMENT , `gibbonLibraryItemID` INT(10) UNSIGNED ZEROFILL NOT NULL, `gibbonLibraryShelfID` INT(6) UNSIGNED ZEROFILL NOT NULL, PRIMARY KEY (`gibbonLibraryShelfItemID`)) ENGINE = InnoDB CHARSET=utf8;end
 ALTER TABLE `gibbonLibraryShelf` ADD `shuffle` ENUM('N','Y') NOT NULL DEFAULT 'N' AFTER `active`;end
 CREATE INDEX `gibbonMessengerID` ON gibbonMessengerTarget(`gibbonMessengerID`,`gibbonMessengerTargetID`);end
-";
 ALTER TABLE `gibbonBehaviour` ADD `gibbonMultiIncidentID` varchar(64) DEFAULT NULL;end
+CREATE TABLE `gibbonBehaviourFollowUp` ( `gibbonBehaviourFollowUpID` int(11) unsigned zerofill NOT NULL AUTO_INCREMENT,
+ `gibbonBehaviourID` int(10) unsigned zerofill NOT NULL, `gibbonPersonID` int(10) unsigned zerofill NOT NULL, `followUp` text, `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP, PRIMARY KEY (`gibbonBehaviourFollowUpID`), KEY `gibbonBehaviourID` (`gibbonBehaviourID`));end
 ";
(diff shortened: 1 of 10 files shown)


CHANGEDB.php
```diff
@@ -801,4 +801,6 @@
 ALTER TABLE `gibbonLibraryShelf` ADD `shuffle` ENUM('N','Y') NOT NULL DEFAULT 'N' AFTER `active`;end
 CREATE INDEX `gibbonMessengerID` ON gibbonMessengerTarget(`gibbonMessengerID`,`gibbonMessengerTargetID`);end
 ALTER TABLE `gibbonBehaviour` ADD `gibbonMultiIncidentID` varchar(64) DEFAULT NULL;end
+CREATE TABLE `gibbonBehaviourFollowUp` ( `gibbonBehaviourFollowUpID` int(11) unsigned zerofill NOT NULL AUTO_INCREMENT,
+ `gibbonBehaviourID` int(10) unsigned zerofill NOT NULL, `gibbonPersonID` int(10) unsigned zerofill NOT NULL, `followUp` text, `timestamp` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP, PRIMARY KEY (`gibbonBehaviourFollowUpID`), KEY `gibbonBehaviourID` (`gibbonBehaviourID`));end
```
SKuipers, line 805:
```suggestion
CREATE TABLE `gibbonBehaviourFollowUp` ( `gibbonBehaviourFollowUpID` INT(11) UNSIGNED ZEROFILL NOT NULL AUTO_INCREMENT,
 `gibbonBehaviourID` INT(10) UNSIGNED ZEROFILL NOT NULL, `gibbonPersonID` INT(10) UNSIGNED ZEROFILL NOT NULL, `followUp` TEXT, `timestamp` TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP, PRIMARY KEY (`gibbonBehaviourFollowUpID`), KEY `gibbonBehaviourID` (`gibbonBehaviourID`));end
```

modules/Behaviour/behaviour_manage_edit.php
```diff
@@ -171,12 +170,33 @@
...
+                ]));
+                }
+
+                //Allow entry of fresh followup
                 $row = $form->addRow();
                     $column = $row->addColumn();
-                    $column->addLabel('followup', __('Follow Up'));
-                    $column->addTextArea('followup')->setRows(5)->setClass('fullWidth')->setValue($values['followup']);
-
+                    $column->addLabel('followUp', (empty($logs) ? __('Follow Up') : __('Further Follow Up')) .' / '.__('Notes'))->description(__('If you are the student\'s teacher, please include details such as: the location & lesson, what was the incident, what did you do.'));
```
SKuipers, line 196: It looks like the description has come across from the first aid feature.
```suggestion
                    $column->addLabel('followUp', (empty($logs) ? __('Follow Up') : __('Further Follow Up')));
```

src/Domain/Behaviour/BehaviourGateway.php
```diff
@@ -254,4 +254,23 @@ public function selectMultipleStudentsOfOneIncident($gibbonMultiIncidentID)
 
         return $this->db()->select($sql, $data);
     }
+    public function queryFollowUpByBehaviourID($gibbonBehaviourID)
```
SKuipers, line 257: Since this relates to the behaviour follow up, it'd be best located in the new BehaviourFollowupGateway you've created. For consistency, since it's not using a runQuery() method or query $criteria value, selectFollowUpByBehaviourID would be a better method name.

SKuipers (changes requested):
Hi Ali, this is coming together nicely. I've added a couple code review comments, and there are two additional considerations for this feature:

- The Add and Add Multi pages for behaviour also contain a follow up text box, so these process scripts need updated to write this follow up to the new table.
- Since we're no longer using the follow up field, the Manage Behaviour Records table will need updated to display the follow up from the new table. This is part of addExpandableColumn() method on the table, and you'll likely need to fetch & append the follow up text to the string that is returned.

SKuipers (approved):
Thanks Ali, this is looking good, I'm going to merge it in and test it out. Be sure to rebase any current branches on these new changes.

PR #1811, opened 2024-03-14
Auto save feature for modules/Planner/planner_add.php

hsh2001:
**Description**

Auto saving planner data!

**Motivation and Context**

It solves the problem of losing the data we're working on when we navigate to another page while writing information.

**How Has This Been Tested?**

Write some data, and I check these datas are saved in session.

**Screenshots**

스크린샷 2024-03-18 오후 5 56 54


SKuipers (commented):
Hi @hsh2001 Thanks for sharing your draft PR, looks like it's off to an excellent start. I like how you've used the existing FormSessionStorage class to good effect. It's shaping up to be a nice clean implementation of an auto-save feature.

I wonder about the onKeyDown event causing a higher than necessary amount of server load, since keypresses can happen quite frequently, and each one would be making an AJAX call to the server. Perhaps a setInterval of 15 seconds would be sufficient (possibly even checking for a keypress in that time, to reduce calls even further). The potential volume of work lost in 15 seconds is likely less than one or two sentences, so this could find a good balance between functionality and efficiency. What do you think?

hsh2001:
Thank you for comment @SKuipers :) 

Yes, I need to use debounce for onKeyDown callback! I will update it and request a review again!

SKuipers (commented):
Hi @hsh2001 Thanks for your continued work on this feature, I know there will be teachers who will be very happy to see this added to Gibbon 😃 

It's looking very close to merging, just two final thoughts from me:

- The debounce code looks great. It does seem a bit short at 300ms. Since this is a larger legacy application, for the benefit of server load, I think somewhere between 5 to 15 seconds might be better. It would be enough that teachers wouldn't lose much work, but would balance server load (with many people accessing Gibbon at once).
- I'd love to add the autosave to some other key forms in the system, and it looks like your implementation will be nice and flexible for this. Rather than exposing the specific implementation through `setOnKeyDown`, which would be less reusable, I wonder if we can rename it to `enableAutoSave(string $formID, string $url)`, and have the implementation of the js function handled internally in the enableAutoSave method. This way, a form could add autosave by calling this method + creating an AJAX. What do you think?

Thanks again! 👍

hsh2001:
Hello @SKuipers !

Nice comment, I 100% agree with you! I will fix it!

hsh2001:
@SKuipers 
First of all, thank you for your thorough review of the code and for your nice comments. However, I think you are mistaken about something. 

This is Debounce, so it doesn't send a request every 300 milliseconds. According to Debounce logic, it shouldn't send a request until 300 milliseconds of inactivity have passed. Sending a request every 300 milliseconds is throttling logic. If we increased the duration to 5-15 seconds, the user would have to wait 5-15 seconds without taking any action before the autosave would be triggered. (I think this would be very rare.) I don't think this would fulfill our intention of improving the frustrating experience of accidentally hitting another link while writing and losing what you're writing. 

What do you think?

src/Forms/Form.php
```diff
@@ -335,6 +335,18 @@ public function addConfirmation($message)
         return $this;
     }
 
+    /**
+     * setOnKeyDown
+     * Add a javascript function to the form's onkeydown event.
+     * @param string $function
+     * @return self
+     */
+    public function setOnKeyDown($function)
```
SKuipers, line 344: Thanks for making these changes. I think this method would also work best as `enableAutoSave(string $url, string $formId)`, with the `$keydownJS` being created and handled internally in the method. That way, if the JS logic or events it's bound to ever change, the implementation would only need to change in one place.

SKuipers (changes requested):
Thanks @hsh2001 for your quick and positive response to feedback. The debounce makes sense the way you've explained it, thanks, I hadn't quite picked up on that nuance when just reading the code itself. I've added one suggestion, for consistency/maintainability, and then it's ready to merge 😃 looking forward to testing it out.

hsh2001:
Thank you for check @SKuipers !

Please review again!

SKuipers:
Hi @hsh2001 Thanks again for your work on this! I'm just going to merge it in now. My apologies for the delays with code review, it's been a busy couple weeks leading up to Easter break here. Huge thanks!

SKuipers (approved):

hsh2001:
Thank you @SKuipers :)

PR #1812, opened 2024-03-21
Link a single behaviour record to another behaviour record

ali-ichk:
**Description**
Added the ability so that a behaviour record that is not linked to any other behaviour record can be manually linked to another behaviour record created by the teacher within 30 days.


SKuipers (approved):
Thanks Ali, I'm going to merge this in and test it out 👍

PR #1813, opened 2024-03-22
Behaviour: add notifications for positive behaviour records

SKuipers:
Currently, behaviour notifications are only sent for negative records. However, it would be good to draw attention to those times where students receive positive behaviour notes, so this PR enables notifications for both positive and negative records.

Thanks @Fongngtis for contributing these code changes to the core.

**How Has This Been Tested?**
Locally

**Screenshots**
Screen Shot 2024-03-22 at 9 49 52 AM


ali-ichk (commented):
Tested locally. The functions seems to work properly.

ali-ichk (approved):
Tested locally. The functions seems to work properly.

PR #1816, opened 2024-04-09
Adding the Study Year beside the form group in student history details

ali-ichk:
**Description**

Updated the Student History list to show the form group and Study Year, especially important for students in mixed form groups, as before it didn’t tell which year they were in.

### DIFF
--- a/src/Domain/Students/StudentGateway.php
+++ b/src/Domain/Students/StudentGateway.php
@@ -406,8 +406,8 @@ class StudentGateway extends QueryableGateway
 
     public function selectStudentEnrolmentHistory($gibbonPersonID)
     {
-        $dataSelect = ['gibbonPersonID' => $gibbonPersonID];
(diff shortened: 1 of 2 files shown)


src/Domain/Students/StudentGateway.php
```diff
@@ -403,4 +403,19 @@ public function selectActiveStudentNames($gibbonSchoolYearID)
 
         return $this->db()->select($sql, $data);
     }
+
+    public function selectStudentEnrolmentHistory($gibbonPersonID)
+    {
+        $dataSelect = ['gibbonPersonID' => $gibbonPersonID];
```
SKuipers, line 409: This is looking good, and great to see the SQL moved over into a Gateway class. Thanks for the quick work!

One very quick an minor change: when transplanting the SQL, be sure to simplify the $data and $sql variables, as they often have longer names like $dataSelect, etc because back in the day Ross was avoiding variable name conflicts with multiple queries in a script, but luckily Gateway classes solve this my using separate methods.

SKuipers (changes requested):

SKuipers (approved):

PR #1819, opened 2024-04-17
Markbook: Fixed the issue of undefined array key "gibbonDepartmentID"

ali-ichk:
**Description**

Fixed the issue of Warning: undefined array key "gibbonDepartmentID" in /Markbook/markbook_view_allClassesAllData.php


modules/Markbook/markbook_view_allClassesAllData.php
```diff
@@ -139,11 +139,15 @@
...
+    $departmentAccess = null;
+    if(isset($class['gibbonDepartmentID'])) 
+    {
+        $departmentAccess = $container->get(DepartmentGateway::class)->selectMemberOfDepartmentByRole($class['gibbonDepartmentID'], $session->get('gibbonPersonID'), ['Coordinator', 'Teacher (Curriculum)'])->fetch();
+    }
 
     //Get teacher list
     $teacherList = getTeacherList( $pdo, $gibbonCourseClassID );
-	$canEditThisClass = (isset($teacherList[ $session->get('gibbonPersonID') ]) || $highestAction2 == 'Edit Markbook_everything' || ($highestAction2 == 'Edit Markbook_multipleClassesInDepartment' && !empty($departmentAccess)));
+	$canEditThisClass = (isset($teacherList[$session->get('gibbonPersonID')])) ?? ($highestAction2 == 'Edit Markbook_everything') ?? ($highestAction2 == 'Edit Markbook_multipleClassesInDepartment' && !empty($departmentAccess));
```
yookoala, line 150: I'm not sure this is the correct way to use [Null coalescing operator](https://www.php.net/manual/en/migration70.new-features.php#migration70.new-features.null-coalesce-op) (`??`). Are you intend to use [Short-ternary](https://www.php.net/manual/en/language.operators.comparison.php#language.operators.comparison.ternary) (`?:`) instead.

SKuipers: Thanks Koala. Ali, it does look like replacing or `||` operators with `??` null coalesce operators isn't going to fix this particular issue, and appears to have introduced an error that the CI tests have caught.

In this case, having a moment to look closer at the code, it looks like the error is coming from a missing array key for `$class['gibbonDepartmentID']`, which we can see being used by line 142 in the original code. The next question is, why is this array key missing? To check this, we can trace it back to where the `$class` variable comes from, and see that it is set on line 102 `$class = getClass($pdo, $session->get('gibbonPersonID'), $gibbonCourseClassID, $highestAction );`

Tracing the getClass function back to it's source leads to an (albeit outdated) query in the Markbook/moduleFunctions.php file on line 202. If you look closely at this code, there appears to be a conditional statement in which one query returns a gibbonDepartmentID, and the other doesn't. Fixing this will likely fix the root of the issue. 

I know I did mention to check if it was an issue of a missing null coalesce, but in this case, tracking things back helps to show where the problem may be originating.

ali-ichk: Thanks koala for your feedback. Thanks Sandra for pointing out the root cause of the issue. Much appreciated.

SKuipers:
Thanks Ali, I'll merge this now. Now that the root cause is fixed, I've reverted the other changes, as they introduce additional complexity that should hopefully not be needed.

PR #1824, opened 2024-04-24
Library: Adding autocomplete to these two fields Vendor and Location Details

ali-ichk:
**Description**
Library: added autocomplete to the Vendor and Location Detail fields in Add/Edit item form under Manage Catalogue.


SKuipers (approved):
Thanks Ali, this is looking excellent and I know our librarian will be quite happy to see these added to Gibbon 😃

PR #1825, opened 2024-04-29
Behaviour: add a new permission so that teachers can only view behaviours created by them

ali-ichk:
**Description**
Added the _my permission in the behaviour_view and  behaviour_view_details page so that teachers and staff are restricted to only view the behaviour records created by them.


CHANGELOG.txt
```diff
@@ -47,6 +47,7 @@ v27.0.00
         Behaviour: updated the view so that followups are displayed as conversational logs and stored in a separate table
         Behaviour: added the ability to able to link a behaviour record to some other existing behaviour record
         Behaviour: added positive behaviour records to the Daily Behaviour Summary CLI script
+        Behaviour-Add-a-new-permission-so-that-teachers-can-only-view-behaviours-created-by-them
```
SKuipers, line 50: This has come through a bit odd, maybe a copy-paste thing?

src/Domain/Behaviour/BehaviourGateway.php
```diff
@@ -227,7 +229,56 @@ public function queryBehaviourRecordsByPerson(QueryCriteria $criteria, $gibbonSc
...
+
+        return $this->runQuery($query, $criteria);
+    }
+
+    public function getBehaviourRecordsByPersonAndCreator($gibbonSchoolYearID, $gibbonPersonID, $gibbonPersonIDCreator) {
+        $data = ['gibbonSchoolYearID' => $gibbonSchoolYearID, 'gibbonPersonID' => $gibbonPersonID, 'gibbonPersonIDCreator' => $gibbonPersonIDCreator];
+        $sql = 'SELECT gibbonBehaviour.*, creator.title as titleCreator, creator.surname AS surnameCreator, creator.preferredName AS preferredNameCreator FROM gibbonBehaviour JOIN gibbonPerson AS creator ON (gibbonBehaviour.gibbonPersonIDCreator=creator.gibbonPersonID) WHERE gibbonSchoolYearID=:gibbonSchoolYearID AND gibbonBehaviour.gibbonPersonID=:gibbonPersonID AND gibbonBehaviour.gibbonPersonIDCreator=:gibbonPersonIDCreator ORDER BY date DESC';
+        
+        return $this->db()->select($sql, $data);
     }
```
SKuipers, line 283: In the end it looks like this new method was not needed.
```suggestion
```

CHANGELOG.txt
ali-ichk: Updated!! Thanks...

modules/Behaviour/moduleFunctions.php
```diff
@@ -58,8 +58,12 @@ function getBehaviourRecord(ContainerInterface $container, $gibbonPersonID)
                 ->sortBy('timestamp', 'DESC')
                 ->fromPOST($schoolYear['gibbonSchoolYearID']);
 
-            $behaviourRecords = $behaviourGateway->queryBehaviourRecordsByPerson($criteria, $schoolYear['gibbonSchoolYearID'], $gibbonPersonID);
-
+            if (empty($gibbonPersonIDCreator)) {              
+                $behaviourRecords = $behaviourGateway->queryBehaviourRecordsByPerson($criteria, $schoolYear['gibbonSchoolYearID'], $gibbonPersonID);
+            } else {
+                $behaviourRecords = $behaviourGateway->queryBehaviourRecordsByPerson($criteria, $schoolYear['gibbonSchoolYearID'], $gibbonPersonID, $gibbonPersonIDCreator);
+            }
```
SKuipers, line 65: It looks like this conditional can be simplified even further, since if $gibbonPersonIDCreator is already null, it will be passed as a null value into the method, which does its own empty() check internally.
```suggestion  
            $behaviourRecords = $behaviourGateway->queryBehaviourRecordsByPerson($criteria, $schoolYear['gibbonSchoolYearID'], $gibbonPersonID, $gibbonPersonIDCreator);
```

modules/Behaviour/moduleFunctions.php
```diff
@@ -86,37 +90,37 @@ function getBehaviourRecord(ContainerInterface $container, $gibbonPersonID)
             $table->addMetaData('hidePagination', true);
 
             $table->addExpandableColumn('comment')
-                ->format(function($beahviour) {
+                ->format(function($behaviour) {
```
SKuipers, line 93: Thanks for catching the typo 😄

modules/Students/student_view_details.php
```diff
@@ -2426,9 +2426,15 @@
                             $page->addError(__('Your request failed because you do not have access to this action.'));
                         } else {
                             include './modules/Behaviour/moduleFunctions.php';
+                            
+                            $highestActionBehaviour = getHighestGroupedAction($guid, '/modules/Behaviour/behaviour_view.php', $connection2);
 
                             //Print assessments
-                            echo getBehaviourRecord($container, $gibbonPersonID);
+                            if ($highestActionBehaviour == 'View Behaviour Records_my') {
```
SKuipers, line 2433: In hindsight, can we flip this around, so it's checking the _all permission when it passes no third parameter, otherwise it always passes a third parameter? This is more future-proofed, should there ever be new permissions.

src/Domain/Behaviour/BehaviourGateway.php
```diff
@@ -23,6 +23,7 @@
 
 use Gibbon\Domain\QueryCriteria;
 use Gibbon\Domain\QueryableGateway;
+use Gibbon\Domain\Traits\SharedUserLogic;
```
SKuipers, line 26:
```suggestion
```

src/Domain/Behaviour/BehaviourGateway.php
```diff
@@ -39,11 +40,12 @@ class BehaviourGateway extends QueryableGateway implements ScrubbableGateway
     use TableAware;
     use Scrubbable;
     use ScrubByPerson;
+    use SharedUserLogic;
```
SKuipers, line 43: This accidentally came though from the StudentGateway.
```suggestion
```

SKuipers (changes requested):
Hi Ali, this is looking excellent, with a lot of good thinking and troubleshooting having gone into it. I've added some notes and suggestions, based on reading the code in more detail, but many are just ways to tidy it up some more, rather than logic changes. Thanks again 😄

SKuipers (approved):
Thanks! 👍

PR #1826, opened 2024-04-30
Improvement in Generic OAuth

leofreitas:
**Description**
Some systems require additional parameters for a login request in order to read the user's basic profile.

**Motivation and Context**
I needed to do SSO and the institution did not use the email field as a username. With this PR, the user can configure it without changing the code. Now it is more generic and flexible.

**How Has This Been Tested?**
Locally and Travis

**Screenshots**
image

### DIFF
--- a/modules/System Admin/thirdPartySettings_ssoEdit.php
+++ b/modules/System Admin/thirdPartySettings_ssoEdit.php
@@ -139,6 +139,20 @@ if (isActionAccessible($guid, $connection2, '/modules/System Admin/thirdPartySet
         $row = $form->addRow()->addClass('settingActive');
             $row->addLabel('userEndpoint', __('API User Endpoint'));
             $row->addURL('userEndpoint')->required();
+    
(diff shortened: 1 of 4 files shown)


SKuipers:
Hi Leo, thanks for suggesting these changes. I haven't been using the Generic OAuth in production, so it's good to hear you've been able to get it working with these changes. Before merging, I think I'm reading the code correctly, but just wanted to check: it looks like preferred_name is being used as an authentication field (which generally wouldn't be a unique field in Gibbon), but that's only for data on the other end of the OAuth authentication? And so, the returned value matching the `usernameAttribute` would still need to match a valid username in Gibbon, from what I can see? Just wanted to check, since the use of a non-unique value here could have unwanted side-effects. Thanks!

leofreitas:
Hi Sandra, that's right! The usernameAttribute in my example is preferred_name. It's on the other side of OAuth authentication, so the value returned needs to match a valid username in Gibbon.

SKuipers:
Thanks Leo for confirming! These changes look good, I've merged them into v27

PR #1827, opened 2024-05-03
Fix error that generates page breaks when adding a new adult to the f…

leofreitas:
**Description**
When adding a new adult to the family with Contact Priority != 1 it generates a page break error

**How Has This Been Tested?**
Local and travis

**Screenshots**

Captura de Tela 2024-05-03 às 16 06 21


SKuipers (approved):
Oh?! Well caught! It's odd that this has persisted, since we've been using the feature for so long in production and testing. Thanks for the fix!

PR #1828, opened 2024-05-06
Admissions: Changed the string of the withdraw notification

ali-ichk:
**Description**
Changed the tense of the string when a student is going to withdraw from the school in the future.


SKuipers (commented):
Thanks @ali-ichk, can you add a changelog entry for this as well? Then I'm happy to merge it and update it in production.

PR #1829, opened 2024-05-07
Messenger: detect emoji in text and display an error message

ali-ichk:
**Description**
Added the ability to detect if a message has emojis, and remove them while showing a warning to the user that the emojis have been removed from the message body.


SKuipers (approved):
Thanks Ali, looking good, I'm going to merge and test it out 👍

PR #1830, opened 2024-05-08
Messenger: Added the ability to post a Wall Message between two dates

ali-ichk:
**Description**
The Message Wall will use a start date and an end date, and display the message for any date between these two.


SKuipers (approved):
Looking great, I'm going to merge and test it out. Nice work tackling the evil MessageTarget script 😆

PR #1834, opened 2024-05-17
Add symfony/polyfill-intl-icu as dependency

yookoala:
**Description**
Provide fallback to older installations. Supports system without php's ext-intl installed.

**Motivation and Context**
Helps with moving forward the #1758 changes without the huge workload to write fallback pattern calls.

**How Has This Been Tested?**
* Not tested. But trusted the Symfony project get this right.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,8 @@
         "microsoft/microsoft-graph": "^1.35",
         "aura/auth": "^4.0",
         "eluceo/ical": "^0.16.0",
-        "robthree/twofactorauth": "^1.8"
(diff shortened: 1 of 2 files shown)


SKuipers:
Hi @yookoala thank you again for moving us forward with the date formats. I've come up with another solution in #1836 that I think will give us the best of forward and backwards compatibility without needing to introduce a new dependency to the project. By decoupling the implementation from the date formatting, using a set of constants that align with how IntlDateFormatter works, I think the solution will hopefully prove to be readable and maintainable.

PR #1838, opened 2024-05-31
Messenger: added the ability to limit individual messages to 50 people

ali-ichk:
**Description**

Added the ability to check if the Individual target is used and if the participant count is greater than 50. If yes, return a warning and ask the creator to select less than 50 individuals and try again.


SKuipers (approved):
Thanks Ali, I'm going to merge and test this. Generally, with one-off return messages, we can use the $page->return->addReturns() method directly on the page, rather than adding it to the global list. I'll make this change post-merge.

ali-ichk:
Thanks, Sandra. Will definitely keep a note of it. Cheers!!

PR #1844, opened 2024-09-03
Update FormFactory.php

kochinc:
Incorrect description for zh_HK

**Description**
Replace 體字 with 繁體字

**Motivation and Context**
zh_HK should be Traditional Chinese which is 繁體字.

**How Has This Been Tested?**
Checked on my local Gibbon installation.

**Screenshots**


SKuipers (approved):
Thanks for the PR! 👍

PR #1845, opened 2024-09-03
Update gibbon.sql

kochinc:
Incorrect zh_HK description

**Description**
Replace 體字 with 繁體字

**Motivation and Context**
zh_HK should be Traditional Chinese which is 繁體字.

**How Has This Been Tested?**
Checked on my local Gibbon installation.

**Screenshots**


SKuipers (approved):
Hi @kochinc Thanks for catching this and submitting a PR 👍

PR #1852, opened 2024-09-26
Rubrics: allowed HTML in Rubrics cells

oswaldkit:
**Description**
To allow user input HTML tags in Rubrics cells


SKuipers (approved):
Thanks! Looks good

PR #1862, opened 2024-10-17
New Mailing List Feature

rossdotparker:
**Description**
Adds a simple, single mailing list, for recording external email recipients.

**Motivation and Context**
A first step in adding a Mailing List target to Messenger. Next steps include adding the target, and then adding an unsubscribe link/feature to emails sent to the mailing list.

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2024-10-17 at 2 57 33 PM


SKuipers (approved):
Hi Ross, this is looking good. I suspect in the future we'll need to add some fields to this table, to track separate mailing list names, where the sign up came from, and likely bounces (to prevent future mailing). For now, this looks like a good start.

PR #1864, opened 2024-10-28
Added multiple lists to new Mailing List(s) feature

rossdotparker:
**Description**
Continuing evolution of this feature, hopefully the last that requires rewriting data structures already created (mature feature set, etc).

**Motivation and Context**
Why have one list, when you can have many lists!

**How Has This Been Tested?**
Locally

**Screenshots**
Screenshot 2024-10-28 at 3 54 20 PM
Screenshot 2024-10-28 at 3 54 33 PM
Screenshot 2024-10-28 at 3 54 53 PM


modules/Messenger/messenger_sendProcess.php
```diff
@@ -66,7 +66,10 @@
     }
 
     // Check which recipients who have been manually unchecked
-    $recipientList = $_POST['gibbonMessengerReceiptID'] ?? [];
+    $recipientList = $_POST['gibbonMessengerReceiptID'] ?? []; 
+    for ($i = 0; $i < count($recipientList); $i++) {
```
rossdotparker, line 70: @SKuipers this is what I mentioned in Slack. Without lines 70 to 72, none of the IDs match because the values in $recipientList have no zero fill, so the count of $unselected matches the count of $recipientList. The result is that the system thinks all values have been unselected.

Is this perhaps stricter matching in PHP? I'm running PHP 8.1..13? Or is it something to do with the new checkboxes?

SKuipers: Thank Ross. That's odd, as I'm seeing the strings with zero fill on my local copy of v28 and v27, and it's working in production. But, it couldn't hurt to add this to ensure it doesn't lose these zeroes.

SKuipers (approved):
This looks good, going to merge and test.

modules/Messenger/messenger_sendProcess.php
rossdotparker: Very odd indeed. What PHP versions are you running?

SKuipers: 8.2.0, and production is 8.0.27, so yours is right in the middle and shouldn't work differently 🤔

rossdotparker: In the end we tracked this one down to the behaviour of the newly-added union, which seems to strip zero-fills from IDs. Good to know that it does not reflect a system-wide change in behaviour!

PR #1876, opened 2025-01-12
Activities: fix bug in activity registration by parent

AbbadiAhmad:
**Description**
The registration button is not shown to the parent. 
there is a variable used before creation.

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**

### DIFF
--- a/modules/Activities/activities_view.php
+++ b/modules/Activities/activities_view.php
@@ -50,12 +50,15 @@ if (isActionAccessible($guid, $connection2, '/modules/Activities/activities_view
         //If student, set gibbonPersonID to self
         if ($roleCategory == 'Student' and $highestAction == 'View Activities_studentRegister') {
             $gibbonPersonID = $session->get('gibbonPersonID');
+        }else if ($roleCategory == 'Parent' and $highestAction == 'View Activities_studentRegisterByParent') {
+            $gibbonPersonID = $_GET['gibbonPersonID'] ?? '';
+
         }
         
         //Check access controls
         $settingGateway = $container->get(SettingGateway::class);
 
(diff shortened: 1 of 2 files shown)


modules/Activities/activities_view.php
```diff
@@ -50,12 +50,15 @@
         //If student, set gibbonPersonID to self
         if ($roleCategory == 'Student' and $highestAction == 'View Activities_studentRegister') {
             $gibbonPersonID = $session->get('gibbonPersonID');
+        }else if ($roleCategory == 'Parent' and $highestAction == 'View Activities_studentRegisterByParent') {
```
SKuipers, line 53: I've had a look, and it appears the `$gibbonPersonID = $_GET['gibbonPersonID'] ?? '';` code on line 83 (86 in your code) is working as expected, so I'm not sure this extra line would be needed. Are you able to share more details about this behaviour versus the existing code?

SKuipers (commented):
Hi @AbbadiAhmad Thanks for the PR, much appreciated 😃 I've added a comment to the code for lines 53-55. Otherwise, the other two changes look good, from what I can see they'd enable viewing even if the parent didn't have children to register.

modules/Activities/activities_view.php
AbbadiAhmad: The line 61 requires none empty gibbonPersonID. Which is filled in lines 53... 
Without that extra code the check fails and parent will get $canAccessRegistration =false regardless of other conditions on kine 61.

SKuipers: Thanks for the clarification, I can see the issue now. I'll merge your PR 👍

SKuipers (approved):

PR #1881, opened 2025-02-02
System: removed Date of Birth field from Public Registration when no minimum registration age is set

rossdotparker:
**Description**
Date of birth is superfluous when their is no age requirement in public registration, and so the form can be simplified by its removal.

**Motivation and Context**
Improving the public registration workflow

**How Has This Been Tested?**
Locally


publicRegistrationProcess.php
```diff
@@ -135,8 +132,6 @@
     }
 
     // Check publicRegistrationMinimumAge
-    $publicRegistrationMinimumAge = $settingGateway->getSettingByScope('User Admin', 'publicRegistrationMinimumAge');
-
     if (!empty($publicRegistrationMinimumAge) > 0 and $publicRegistrationMinimumAge > (new DateTime('@'.Format::timestamp($dob)))->diff(new DateTime())->y) {
```
SKuipers, line 135: This line, although not part of your changes, seems somewhat suspect 🤔 `!empty($publicRegistrationMinimumAge) > 0` is technically valid, but not how one would normally handle a boolean. The time conversion is interesting too ... perhaps a task for future refactoring.

SKuipers (approved):
Seems legit 👍

PR #1886, opened 2025-02-07
Fix css to add space and commas between names. of multiple behaviour …

ali-ichk:
Fixed css and added space between names.

### DIFF
--- a/modules/Behaviour/behaviour_manage_edit.php
+++ b/modules/Behaviour/behaviour_manage_edit.php
@@ -115,7 +115,6 @@ if (isActionAccessible($guid, $connection2, '/modules/Behaviour/behaviour_manage
                 if(!empty($values['gibbonMultiIncidentID'])) {
                     $students = $behaviourGateway->selectMultipleStudentsOfOneIncident($values['gibbonMultiIncidentID'])->fetchAll();
                 }
-            
 
                 //Student
                 $row = $form->addRow();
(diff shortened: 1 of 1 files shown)


modules/Behaviour/behaviour_manage_edit.php
```diff
@@ -128,11 +128,14 @@
...
+                    
                     foreach ($students as $i => $student) {
                         if ($student['gibbonPersonID'] != $values['gibbonPersonID']) {
-                            $url = Url::fromModuleRoute('Students', 'student_view_details')->withQueryParams(['gibbonPersonID' => $student['gibbonPersonID'], 'subpage' => 'Behaviour']);
-                            $col->addContent('<b>'.Format::link($url, Format::name('', $student['preferredNameStudent'], $student['surnameStudent'], 'Student', false, true)).'</b>');
+                            $url = Url::fromModuleRoute('Students', 'student_view_details')
+                                ->withQueryParams(['gibbonPersonID' => $student['gibbonPersonID'], 'subpage' => 'Behaviour']);
+                            
+                            // Add inline CSS for spacing
+                            $col->addContent('<span style="margin-right: 5px;"><b>' . Format::link($url, Format::name('', $student['preferredNameStudent'], $student['surnameStudent'], 'Student', false, true)) . '</b>,</span>');
```
SKuipers, line 138: This looks like a good spot for tailwind. The right margin and the bold could both be classes on the span.

PR #1890, opened 2025-02-11
Only add tinymcesetup event listener once

morne-entrant:
**Description**

When editing a Freelearning Unit (Home > Free Learning > Manage Units > Edit Unit), if there multiple smartblocks in the unit, the html tab in the tinymce editor does not _always_ work. Upon investigation, it seems that the event listener was not always correctly added for all of the `#{{ id }}edButtonHTML` elements. 

**Motivation and Context**
By only adding the eventlistener once, tinymce is set up correctly, and after that the click event listeners is added for each `#{{ id }}edButtonHTML` by loopindirectly, rather than through the `tinymceSetup` event listener.

**How Has This Been Tested?**
Tested on the MOS environment and confirmed by users that fix resolved the issue. Each of the smartblocks are not editable consistently.

### DIFF
--- a/resources/templates/components/editor.twig.html
+++ b/resources/templates/components/editor.twig.html
@@ -120,34 +120,40 @@ Rich Text Editor
 </div>
 
 <script type="text/javascript">
-document.addEventListener('tinymceSetup', function (event) {
-
-    {% if tinymceInit %}
+    function tinymceSetupListener(event) {
+        {% if tinymceInit %}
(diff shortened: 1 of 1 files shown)


rossdotparker:
@SKuipers are you OK to take a look at this PR for Morne?

SKuipers (approved):
Thanks @morne-entrant for finding and fixing this issue, this looks like a good solution to ensure that the event listener is only added once. I'm happy to merge this into v29.

PR #1893, opened 2025-02-12
Reports: add category name to criteria list in Manage Criteria

Harrison-Idornigie:
**Description**

**Motivation and Context**

**How Has This Been Tested?**

**Screenshots**


modules/Reports/reporting_criteria_manage.php
```diff
@@ -188,6 +188,10 @@
 
     $table->addColumn('name', __('Criteria'));
     $table->addColumn('criteriaType', __('Type'));
+    $table->addColumn('category', __('Category'))
+        ->format(function ($row) {
+            return $row['category'] ?? '';
+        });
```
SKuipers, line 194: Rather than adding a new column which could take up more horizontal space, especially if category is not used by a school, lets use a formatDetails to combine this with the Type column. What do you think?

```suggestion
    $table->addColumn('criteriaType', __('Type'))
        ->description(__('Category'))
        ->formatDetails(function ($row) {
            return $row['category'] ?? '';
        });
```

Harrison-Idornigie:
> Thanks for the PR, looks like a good idea based on your previous changes. I've added a comment to the code. Also, we've just fixed the breaking tests in v29, so can you rebase your branch onto the current commit of v29. Thanks! 😃

Thanks for your response. We just wanted to be able to visually categorize the report fields and it makes the generated reports look well formatted and organized

Harrison-Idornigie:
@SKuipers I need advice on how to add student number generation logic, similar to userID generation. We specify the format and when the students are being registered, the next number in the sequence is added for the student if no unique value is provided. 

Or at least separate the userID generation logic for students and staff. We could add the student ID generation logic here:
image

Or I was thinking of using a custom module but I'm not sure how that would work.

SKuipers (approved):

SKuipers:
Hi @Harrison-Idornigie I think what you're looking for may already exist. If you go into User Admin > User Settings, at the top you can define Username Formats separately for different roles, including prefixes and numerical counting. These usernames could also be used as student IDs, so that both are unique and auto-generated by the system.

PR #1894, opened 2025-02-17
fileUploaded: Fix the error of File not supported i

ali-ichk:
**Description**

Added a fix for some temporary image files especially in the Stream module for an error related to exif: exif_read_data(phpUf54qr): File not supported in FileUploader.php.

PHP Manuals referred to:
https://www.php.net/manual/en/function.exif-read-data.php
https://www.php.net/manual/en/function.getimagesize.php 

Also added a fix for the deprecation messages:  implicit conversion from float to int loses precision

### DIFF
--- a/src/Gibbon/FileUploader.php
+++ b/src/Gibbon/FileUploader.php
@@ -307,13 +307,34 @@ class FileUploader
             return $sourcePath;
         }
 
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Thanks Ali, looking forward to fixing this issue, going to merge and test.

PR #1895, opened 2025-02-18
Finance: Refactor the form on modules/Finance/expenses_manage_view.php

ali-ichk:
**Description**
Refactor the form on modules/Finance/expenses_manage_view.php to use the standard Form Class

Removed the duplicate file modules/Finance/expenses_manage_print.php as it is a duplicate of expenses_manage_view.php.


SKuipers (approved):
Looking good, I'm going to merge and test.

PR #1897, opened 2025-02-19
Finance: Refactor tables into DataTable class. 

ali-ichk:
**Description**
Refactor the table on modules/Finance/expenseRequest_manage.php.
Created new Gateways and Refactored similar sql scripts into these gateways.


SKuipers (approved):
Perfect, thanks for cleaning up the SQL into Gateway classes 🎉

PR #1902, opened 2025-02-25
Added missing country codes to the Database

ali-ichk:
**Description**

Added the missing country codes in the DB


SKuipers (approved):
Thanks! 👍

PR #1904, opened 2025-02-26
Replaced the Regex equation to separate variable or keywords with two ::

ali-ichk:
**Description**
Fixed the issue of using : in the search keyword.


SKuipers (approved):
👍

PR #1910, opened 2025-02-28
Fixed the issue of updating attainment percentage values from raw values

ali-ichk:
**Description**
When Raw Attainment is enabled, it’s possible to set a total value for a column, so teachers can enter arbitrary grades for that column. If they also have Percent set as the Attainment Scale, when they type a number and press enter, it is automatically calculating the percent.


SKuipers (approved):
Interesting fix, I'll merge and test 👍

PR #1914, opened 2025-03-03
Batch creation for internal assessments

AndroidOL:
**Description**
Added batch creation function for internal assessments.

**Motivation and Context**
Internal assessments cannot be added in bulk, which is very time-consuming. We may have some routine daily inspections, and this method allows batch additions. (The downside is that the corresponding tests need to be manually renamed, though duplication is possible).

**How Has This Been Tested?**
This is a minor change.

**Screenshots**

### DIFF
--- a/modules/Formal Assessment/internalAssessment_manage.php
+++ b/modules/Formal Assessment/internalAssessment_manage.php
@@ -75,6 +75,11 @@ if (isActionAccessible($guid, $connection2, '/modules/Formal Assessment/internal
             $page->navigator->addHeaderAction('addMultiple', __('Add Multiple Columns'))
                 ->setURL('/modules/Formal Assessment/internalAssessment_manage_add.php')
                 ->addParams($params)
+                ->setIcon('page_new')
+                ->displayLabel();
+            $page->navigator->addHeaderAction('addMultipleTimes', __('添加多个考试与多个时间列'))
+                ->setURL('/modules/Formal Assessment/internalAssessment_manage_addMulti.php')
+                ->addParams($params)
                 ->setIcon('page_new_multi')
                 ->displayLabel();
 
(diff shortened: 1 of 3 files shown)


SKuipers:
Hi @AndroidOL Thanks for this PR, I appreciate your enthusiasm to contribute code to Gibbon 😄 It's interesting to see you adding this feature, however, as bulk adding is the default behaviour for creating new Internal Assessments. Are you able to help explain these changes in more detail, compared to the existing functionality?

AndroidOL:
> Hi @AndroidOL Thanks for this PR, I appreciate your enthusiasm to contribute code to Gibbon 😄 It's interesting to see you adding this feature, however, as bulk adding is the default behaviour for creating new Internal Assessments. Are you able to help explain these changes in more detail, compared to the existing functionality?

@SKuipers Hello SKuipers, I apologize for not clearly explaining my request earlier, which may have taken up some of your time.

Let me clarify:
Our teachers conduct a simple review test for the entire class every day during lessons, which results in frequent creation of internal assessments.

For example:

    Y9 Mathematics – Topic: Trigonometry, Name: In-Class Test, Schedule: Monday to Friday.
    Y9 English – Topic: School Life, Name: In-Class Test, Schedule: Monday to Friday.

Using Gibbon, I believe this would require creating at least five separate gradebook entries:
    Monday (Math, English)
    Tuesday (Math, English)
    Wednesday... and so on.

However, with this PR, I can simply select both Math and English, **set repetition to 5 with an interval of 1**, and it will generate **10 blank gradebook entries** automatically.

Although my code is quite basic, mostly copied and combined from other files, it significantly reduces the repetitive workload for teachers when creating assessments. I hope you can evaluate our needs.

For instance, if Math runs Monday to Friday, and English runs only on Monday, Wednesday, and Friday, I can set:
    Math: repetition = 5, interval = 1
    English: repetition = 3, interval = 2

With just two operations, I can complete the entire setup.

SKuipers:
Hi @AndroidOL It sounds like the Markbook would be the ideal place for these kinds of frequent assessment, as it's a teacher-managed area for grades and comments, which they can create and edit as needed. Internal Assessment is a more formalised area for assessments, which is why is requires being set up by admin in advanced (for example, IB Internal Assessments, which are standardised for all IB courses and have set dates.)

PR #1916, opened 2025-03-09
Fix to avoid html error

ali-ichk:
**Description**
Fix to remove html error


SKuipers (approved):
Thanks! Going to merge and test

PR #1917, opened 2025-03-10
Bug: There was an error that was preventing the alarm from running

ali-ichk:
**Description**
It was using the old ThickBox css. Refactored the code to remove the tb_show and tb_remove which was causing a Reference Error.


SKuipers (approved):
Thanks Ali, I'm going to merge and test 👍

PR #1923, opened 2025-03-13
Refactored the code to support Chinese characters in file names and ids.

ali-ichk:
**Description**
Updated the regex statement inside the preg_replace statement to support Chinese characters in filenames and ids.


wellinwu:
THANKS!

SKuipers (approved):
👍

PR #1924, opened 2025-03-18
Bug: Fixed the issue of correctly extracting a URL Link from text.

ali-ichk:
**Description**

If there was a full stop at the end of an url (not part of the link), the full stop is added to the url link and this breaks the link.

For example: "there is the link [https://www.w3schools.com/php/func_string_nl2br.asp. Very helpful."

Now the link is actually https://www.w3schools.com/php/func_string_nl2br.asp and the full stop is added to finish the sentence. However, the link is displayed as https://www.w3schools.com/php/func_string_nl2br.asp. 
which is broken and when you click it, it gives a 404 error.

### DIFF
--- a/src/Services/Format.php
+++ b/src/Services/Format.php
@@ -653,7 +653,7 @@ class Format
      */
     public static function hyperlinkAll(string $value)
     {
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Wonderful, thanks for this fix!

PR #1930, opened 2025-04-07
System: removed the sendNotificationsAsBcc function

ali-ichk:
**Description**

System: the View Details button in email does not work when using sendNotificationsAsBcc. Thus, removed the use of sendNotificationsAsBcc and the function from the system


SKuipers (approved):
Thanks Ali, looks good. Ideally, when removing a function, we'd deprecate it a few versions ahead of time. However, in this case, it was only ever used in one place in the codebase and not documented elsewhere, so we should be good to proceed.

PR #1931, opened 2025-04-08
User Admin:  added a warning to the rollover page when max_input_vars is too low

ali-ichk:
**Description**

Added a check and warning on the Rollover page to prevent issues when max_input_vars is too low as it can cause the rollover to partially fail. It will also prevent from displaying the proceed button and moving on with the rollover.


SKuipers (approved):
Looks good, going to merge and test.

PR #1932, opened 2025-04-08
Students: add the option for parents and students to see activities in profile

ali-ichk:
**Description**
Added the option/permission for parents and students to be able to see activities in the view profile page.


SKuipers (approved):
Looks great, and nice idea to add the View action to this table. Going to merge and test 😃

PR #1933, opened 2025-04-15
Messenger: added the ability to add recipients to a sent message

ali-ichk:
**Description**
Added the ability to add recipients to a sent message, to enable easier re-sending. Now an owner can directly add new recipients to an earlier sent message and can resend the earlier message to them. This should make it easier to forward the original message to new participants if they are recently added.


SKuipers (changes requested):
Hmm, some Gateway refactoring seems to have tagged along with this PR, are you able to see if you can re-base it, or removed the changed gateways? Thanks!

SKuipers (approved):
Thanks Ali, you've done a wonderful job cleaning up some messy queries and implementing a new feature that will be very much appreciated by the activities team and Gibbon users in general.

PR #1934, opened 2025-04-15
Application Form: added the option to customise the success message w…

ali-ichk:
**Description**
Added the option to customise the success message when application forms are submitted. Added a new gibbon setting so that a school can add a customised success message to display to applicants.


SKuipers (approved):
Hi Ali, this is looking good. I've rebased it for you to remove some extra yml files that tagged along. I've also added default text for the setting, otherwise the popup will be blank for any systems that haven't configured the setting. There's also currently two application form systems in place (transitional 😅) so I've applied the setting to the new Admissions application form as well.

PR #1935, opened 2025-04-21
Reports:  added a way to archive templates that are no longer in use

ali-ichk:
**Description**

Templates should not be deleted even when not in use, in case reports from past years need to be re-generated. Added the option to make a template active or inactive. Only active records are shown by default.


SKuipers (approved):
👍

PR #1936, opened 2025-04-21
Reports: added an option to archive template assets

ali-ichk:
**Description**

Template Assets should not be deleted even when not in use, in case reports from past years need to be re-generated. However, keeping them around makes things messy and confusing.

For assets, there needs to be an Active field, and only active records are shown by default.


modules/Reports/templates_assets.php
```diff
@@ -37,9 +38,31 @@
     $prototypeGateway = $container->get(ReportPrototypeSectionGateway::class);
     $fontGateway = $container->get(ReportTemplateFontGateway::class);
 
+    $request = ['active' => $_GET['active'] ?? 'Y'];
```
SKuipers, line 41: Although we don't need this form for the filters anymore, the asset page desperately needs a search function. Are you able to change the active filter into a search field, which should be easy to connect to the existing DataTable. Thanks!

modules/Reports/templates_assets.php
```diff
@@ -108,6 +133,11 @@
                     return;
             }
 
+            $actions->addAction('edit', __('Change Active Status'))
+                ->setIcon('cross')
+                ->setURL('/modules/Reports/templates_assets_edit.php')
+                ->addParam('gibbonReportPrototypeSectionID', $template['gibbonReportPrototypeSectionID']);
+
```
SKuipers, line 140: Hmmm, this is interesting to add it as a stand-alone action, rather than part of the edit action? 🤔

SKuipers (changes requested):
Thanks Ali, I've added a couple thoughts, and one new request 😄

PR #1937, opened 2025-04-28
System: added a system log when a password reset is initiated and changed

ali-ichk:
**Description**
In the Forgot Password page, users can initiate a password reset. For security purposes, this request should be logged with the IP address, so that false requests can be more easily caught. Added a log entry every time a request is initiated and a password is changed via clicking the link in the email

### DIFF
--- a/passwordResetProcess.php
+++ b/passwordResetProcess.php
@@ -90,6 +90,8 @@ else {
         $gibbonPersonID = $row['gibbonPersonID'];
         $email = $row['email'];
         $username = $row['username'];
+        $preferredName = $row['preferredName'];
+        $surname = $row['surname'];
 
         if ($step == 1) { //This is the request phase
             // Use password policy to generate random string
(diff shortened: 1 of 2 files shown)


passwordResetProcess.php
```diff
@@ -145,6 +147,16 @@
...
             if ($mail->Send()) {
+                // Log this password reset request
+                $details = [
+                    'gibbonPersonID' => $gibbonPersonID,
+                    'name' => Format::name('', $row['preferredName'], $row['surname'], 'Staff', false, true),
+                    'changedByID' => $session->get('gibbonPersonID'),
+                    'changedBy' => Format::name('', $session->get('preferredName'), $session->get('surname'), 'Staff', false, true),
+                ];
+
+                $container->get(LogGateway::class)->addLog($session->get('gibbonSchoolYearID'), null, $session->get('gibbonPersonID'), 'Forgot User - Password Request Initiated ', $details, $_SERVER['REMOTE_ADDR']);
```
SKuipers, line 158:
```suggestion
                $container->get(LogGateway::class)->addLog($session->get('gibbonSchoolYearID'), null, $session->get('gibbonPersonID'), 'User - Forgot Password Request Initiated ', $details, $_SERVER['REMOTE_ADDR']);
```

passwordResetProcess.php
```diff
@@ -145,6 +147,16 @@
             ]);
 
             if ($mail->Send()) {
+                // Log this password reset request
+                $details = [
+                    'gibbonPersonID' => $gibbonPersonID,
+                    'name' => Format::name('', $row['preferredName'], $row['surname'], 'Staff', false, true),
+                    'changedByID' => $session->get('gibbonPersonID'),
```
SKuipers, line 154: This page is outside of the normal login session, so the gibbonPersonID and names will need to come from the $row value

passwordResetProcess.php
```diff
@@ -218,6 +230,16 @@
                                 $result = $connection2->prepare($sql);
                                 $result->execute($data);
 
+
+                                // Log this password change
+                                $details = [
+                                    'gibbonPersonID' => $gibbonPersonID,
+                                    'name' => Format::name('', $row['preferredName'], $row['surname'], 'Staff', false, true),
+                                    'changedByID' => $session->get('gibbonPersonID'),
```
SKuipers, line 238: Same here, cannot use $session because the user is not logged in.

SKuipers (approved):
Thanks, looks good, going to merge and test

PR #1940, opened 2025-05-08
Messenger: add a warning if [confirmLink] is in text but read receipts are not turned on

ali-ichk:
**Description**

This is aimed to prevent people from sending a message with [confirmLink] in the text. On the Messenger Preview page, added a big red warning message to the top of the page if there is [confirmLink] in the message body, but read receipts are not turned on. Then added the feature to remove [confirmLink] from the body text if read receipts are not turned on.


SKuipers (changes requested):
Wonderful, thanks Ali, the only thing missing is the changelog entry 😉

SKuipers (approved):

PR #1941, opened 2025-05-10
Timetable: complete UI and code refactor of the timetable

SKuipers:
This PR is a long time coming, started months ago, for something that has needed refactored for a long time...

- Replaces the old renderTT and renderSpaceTT functions with an object oriented set of classes.
- Separates timetable logic into a set of layers, which each load and process their own layer items.
- Adds user preferences for toggling timetable layers, as well as more detailed hover-over information.
- Adds a tabbed mobile-responsive view, making it easier to see a single day's classes.
- Enables viewing all classes from multiple timetables in the same view.
- Enables ad-hoc rendering of lessons that are not attached to timetabled classes.
- Supports custom timetable layers and extending the timetable view with different templates.
- Updates the rendering of the timetable to use a set of Twig templates.
- Many other small tweaks and improvements.

**How Has This Been Tested?**
Locally and soon to be in production.

**Screenshots**
Screenshot 2025-02-24 at 3 55 48 PM
Screenshot 2025-02-24 at 3 56 19 PM


SKuipers (commented):
Merging this in so we can proceed with the v29 string freeze (which is overdue! 😓 )

PR #1945, opened 2025-05-26
System: Used a try-catch block to catch any error/exception

ali-ichk:
**Description**
Added a try-catch block to account for any exception or error thrown if a function used does not exist or is deprecated.


SKuipers (approved):
Thanks Ali, this will help prevent errors in custom modules from breaking the dashboard 😅

PR #1948, opened 2025-05-28
Report: Added a way for tracking and displaying who last edited an as…

ali-ichk:
**Description**
As requested before, tracking and displaying who last edited an asset would be ideal. Thus, added a field in the DB to record and display who last edited an asset.


SKuipers (approved):
Wonderful, thanks for adding this.

PR #1952, opened 2025-06-23
New Core Module - Alerts

ali-ichk:
**Description**

We are creating a new core module called Alerts, which can manage the existing alerts system in Gibbon, while also adding the ability to add and track manually created alerts for students.
We have created the new Module, added new tables to store and record the manual alerts and normalize the types of Alerts. Added the _my and _all permission to add/edit and approve alerts. Refactored the getAlertBar function from functions.php into its own class under the UI/Components. This is the initial commit and further developments of this module will follow.


src/Tables/Prefab/FormGroupTable.php
```diff
@@ -53,6 +54,7 @@ public function __construct(GridView $renderer, StudentGateway $studentGateway,
 
     public function build($gibbonFormGroupID, $canViewConfidential, $canPrint, $sortBy = 'rollOrder, surname, preferredName')
     {
+        global $container;
```
SKuipers, line 57: Rather than using a global here, creating the object in the class constructor would be more maintainable. Ideally, we want to avoid global as much as possible.

src/Tables/Prefab/FormGroupTable.php
```diff
@@ -107,9 +109,12 @@ public function build($gibbonFormGroupID, $canViewConfidential, $canPrint, $sort
             $this->addMetaData('gridFooter', $this->getCheckboxScript($gibbonFormGroupID));
 
             $this->addColumn('alerts')
-                ->format(function ($person) use ($guid, $connection2, $gibbonFormGroupID) {
+                ->format(function ($person) use ($gibbonFormGroupID, $container) {
                     $divExtras = ' data-conf="confidential'.$gibbonFormGroupID.'"';
-                    return getAlertBar($guid, $connection2, $person['gibbonPersonID'], $person['privacy'], $divExtras);
+                    // $output = getAlertBar($guid, $connection2, $person['gibbonPersonID'], $person['privacy'], $divExtras);
```
SKuipers, line 114: Feel free to remove old code, if we ever need it again, it will be in version history.

SKuipers (approved):
Thanks Ali, I'm going to merge and test this 👍 

There are a few areas we'll want to keep developing, such as removing the hard-coded (24 * 60 * 60 * 60) calculation and moving to a setting for choosing the time range.

SKuipers:
It looks like we'll also need to update `tests/acceptance/School Admin/AlertLevelSettingsCept.php`

PR #1953, opened 2025-06-23
System: added public functions to support the refactoring of Higher Education

ali-ichk:
**Description**
Added some public functions to the hook and action gateway under system while refactoring Higher Education Write Reference form.

### DIFF
--- a/src/Domain/System/ActionGateway.php
+++ b/src/Domain/System/ActionGateway.php
@@ -127,6 +127,14 @@ class ActionGateway extends QueryableGateway
 
         return $this->db()->select($sql, $data);
     }
+
+    public function selectActionByModuleAndRole($gibbonModuleID, $gibbonRoleIDCurrent)
+    {
+        $data = ['gibbonModuleID' => $gibbonModuleID, 'actionName' => 'View CAS in Student Profile', 'gibbonRoleID' => $gibbonRoleIDCurrent];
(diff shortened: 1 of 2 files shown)


src/Domain/System/ActionGateway.php
```diff
@@ -127,6 +127,14 @@ public function selectModuleActionsByRole(
 
         return $this->db()->select($sql, $data);
     }
+
+    public function selectActionByModuleAndRole($gibbonModuleID, $gibbonRoleIDCurrent)
+    {
+        $data = ['gibbonModuleID' => $gibbonModuleID, 'actionName' => 'View CAS in Student Profile', 'gibbonRoleID' => $gibbonRoleIDCurrent];
```
SKuipers, line 133: It looks like this value may have been accidentally hard-coded. Is it possible to use the isActionAccessible or getHighestGroupedAction functions, rather than creating a new one?

src/Domain/System/HookGateway.php
```diff
@@ -91,4 +91,12 @@ public function getAccessibleHooksByType(string $type, string $gibbonRoleIDCurre
 
         return $hooks;
     }
+
+    public function selectPermissionByRoleToHook($gibbonRoleIDCurrent, $sourceModuleName, $sourceModuleAction)
+    {
+    $data = ['gibbonRoleIDCurrent' => $gibbonRoleIDCurrent, 'sourceModuleName' => $sourceModuleName];
+    $sql = "SELECT gibbonHook.name, gibbonModule.name AS module, gibbonAction.name AS action FROM gibbonHook JOIN gibbonModule ON (gibbonModule.name='".$sourceModuleName."') JOIN gibbonAction ON (gibbonAction.name='".$sourceModuleAction."') JOIN gibbonPermission ON (gibbonPermission.gibbonActionID=gibbonAction.gibbonActionID) WHERE gibbonAction.gibbonModuleID=(SELECT gibbonModuleID FROM gibbonModule WHERE gibbonPermission.gibbonRoleID=:gibbonRoleIDCurrent AND name=:sourceModuleName) AND gibbonHook.type='Student Profile' ORDER BY name";
```
SKuipers, line 98: Looks like this came from some really old code 😅 the sourceModuleName and sourceModuleAction need to be replaced with tokens and added to the $data array, to prevent SQL injection. Thanks!

PR #1954, opened 2025-06-28
Checks emailAlternate as well as email for matching SSO credentials

sgillott:
**Description**
Made a minor addition in \src\Domain\User\UserGateway.php function selectLoginDetailsByUsername($username) on line 104
changed from: 
(username=:username OR (LOCATE('@', :username)>0 AND email=:username)) 
to:
(username=:username OR (LOCATE('@', :username)>0 AND (email=:username OR emailAlternate=:username))

**Motivation and Context**
For (admittedly small) use case where a teacher is also a parent, and wished to have communications from Gibbon messenger to go to their home email address instead of their school email address, as messenger will only send to the primary email address and not the alternative email address supplied.
So, if you put their home email address in the primary email within Gibbon, and their work email address in the alternative email address, messenger sends to their home email address. Happy teacher!
However, this breaks the single sign on, as the SSO will only check against email, and not emailAlternate. So I need it to check against either the email field or emailAlternate field for the email address to resolve the credentials.

**How Has This Been Tested?**
Tested in our production server against Google SSO and Microsoft SSO
As this just adds an OR - none of the existing functionality is changed, but enhanced in a small but impactful way. And this means I don't need to refactor messenger code, or add a "home email" "work email" logic inside messenger or alter gibbonPerson. I thought this would be least disruptive.

### DIFF
--- a/src/Domain/User/UserGateway.php
+++ b/src/Domain/User/UserGateway.php
@@ -101,7 +101,7 @@ class UserGateway extends QueryableGateway implements ScrubbableGateway
                 FROM gibbonPerson 
                 LEFT JOIN gibbonRole ON (gibbonPerson.gibbonRoleIDPrimary=gibbonRole.gibbonRoleID) 
                 WHERE (
-                    (username=:username OR (LOCATE('@', :username)>0 AND email=:username)) 
+                    (username=:username OR (LOCATE('@', :username)>0 AND (email=:username OR emailAlternate=:username)) 
                     AND status='Full' 
                 )";
 


SKuipers:
Hi @sgillott I appreciate this conundrum with staff who are also parents. As you mentioned, it's a small yet annoying edge case. However, emailAlternate is currently not used in authentication because it cannot be guaranteed to be unique, and many schools use it as a way to attach a parents email to student accounts, or vice versa. It wouldn't be secure at this point for us to change the way Gibbon handles emailAlternate and use it as an authentication method, as this could open up serious security issues for existing schools, and unforseen complexity when using Forgot Password for account reset.

I wonder about an alternate but possibly more robust solution, to add an option to user profiles to "Also send communications to alternate email address". This could enable any user to opt-in to having Messenger send emails to both their primary and alternate email addresses, as Messenger currently only sends to the primary address. It would take a bit more development and settings, but could be a more useful long-term feature.

SKuipers:
Hi @sgillott Following up, for the security reasons noted above, we cannot use the alternate email for login credentials, however I've added a feature request to our list to be able to specify if users should receive email to their alternate email address. This should ideally help solve the original problem, of having staff who are parents who wish to recieve student-related email to their alternate email address.

PR #1955, opened 2025-07-07
Fixed grammar in alarm message

4-nthn:
**Description**

The message letting staff know that an alarm will stay on their screen until it is cancelled had a slight grammatical error, which was fixed.

**Motivation and Context**

It provides clarity into what the message meant.

**How Has This Been Tested?**

Tested to work after sending an alarm

**Screenshots**


SKuipers (approved):
Thanks @4-nthn I appreciate the fix for added clarity. 😃  I've updated the strings in POEditor as well, to ensure this change doesn't create new translation strings.

PR #1956, opened 2025-07-16
string % int in SimpleRenderer.php

fvlasie:
With some queries I was getting: TypeError: Unsupported operand types: string % int


SKuipers (approved):
Thanks for the fix, looks good 👍

PR #1958, opened 2025-08-21
Staff: updated the ordering of the department displayed in the Staff Directory

ali-ichk:
**Description**

Updated the ordering list to display 'Human Technologies' in the correct order in Staff Directory


rossdotparker:
Thanks for solving this one Ali! You've tracked the issue back to a setting (which I forgot that I programmed many years ago)! The issue is ICHK specific, so I'll adjust the setting in User Admin on our install, and thus there is no reason to update the value for other schools. Please roll back these changes on your branch, or delete the branch. Thanks!

PR #1960, opened 2025-08-23
fix(Planner): Hide class participants/guests sidebar in Planner module for Student/Parent roles (Privacy/Safeguarding)

sgillott:
**Description**
Hide the Participants & Attendance (and Guests) sidebar for users in the Student and Parent role categories when viewing lesson plans/homework in the Planner. Staff/teachers still see the panel and can take attendance as usual, so functionality remains as before.

Also adds a defensive check around a call to SchoolYearSpecialDayGateway::selectOffTimetableStudentsByClass() to prevent a fatal error where this method isn’t available.

**What changed**
In modules/Planner/planner_view_full.php, gate the sidebar by adding:
 - Only call $page->addSidebarExtra($form->getOutput()) if gibbonRoleIDCurrentCategory is not Student or Parent.
 - Apply the same gate to the Guests sidebar.

Wrapped the call to selectOffTimetableStudentsByClass() with a method_exists guard and fall back to an empty result to avoid crashes.

**Motivation and Context**
Safeguarding: prevents students/parents from seeing a full class roster (names + photos) on Planner pages, where children's photos or details may potentially be misused/abused.

**How Has This Been Tested?**
Tested on a staging and production server with:
 - Teacher account → roster panel visible; attendance works.
 - Student/Parent accounts → roster/guest panels hidden.
 - Regression check: page loads without error where selectOffTimetableStudentsByClass() is unavailable.

**Screenshots**
PlannerChanges

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -55,6 +55,7 @@ v30.0.00
         Data Updater: fixed attachments being cleared when new data updates are submitted
         Messenger: fixed missing From address on emails when using the Add Recipient action in Send Report
         Planner: fixed Teacher Notes in smart blocks not visible when using the Show Confidential Data toggle
+        Planner: fixed access to class list in Edit Unit when viewing past and upcoming years
         Staff: fixed coverage longer than one week not including staff duty and activity cover
         Timetable Admin: fixed missing Manage Exceptions button on the timetable in Course Enrolment by Person
 
--- a/modules/Planner/units_edit.php
+++ b/modules/Planner/units_edit.php
@@ -153,10 +153,11 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/units_edit.php') =
                         $currentRole = $container->get(RoleGateway::class)->getByID($session->get('gibbonRoleIDCurrent'));
                         $selectedSchoolYear = $container->get(SchoolYearGateway::class)->getByID($gibbonSchoolYearID);
 
-                        $isCurrentYear = $session->get('gibbonSchoolYearIDCurrent') == $gibbonSchoolYearID && $session->get('gibbonSchoolYearIDCurrent') == $session->get('gibbonSchoolYearID');
+                        $isCurrentYear = $session->get('gibbonSchoolYearIDCurrent') == $gibbonSchoolYearID;
                         $canAccessUpcomingYear = !empty($currentRole) && $currentRole['futureYearsLogin'] == 'Y';
+                        $canAccessPastYear = !empty($currentRole) && $currentRole['pastYearsLogin'] == 'Y';
 
-                        if ($isCurrentYear || ($canAccessUpcomingYear && $selectedSchoolYear['status'] == 'Upcoming')) {
+                        if ($isCurrentYear || ($canAccessUpcomingYear && $selectedSchoolYear['status'] == 'Upcoming') || ($canAccessPastYear && $selectedSchoolYear['status'] == 'Past')) {
                             $dataClass = array('gibbonUnitID' => $gibbonUnitID, 'gibbonCourseID' => $gibbonCourseID);
                             $sqlClass = "SELECT gibbonCourseClass.gibbonCourseClassID, gibbonCourseClass.nameShort, running, gibbonUnitClassID, gibbonUnitID
                                         FROM gibbonCourseClass
--- a/modules/Planner/units_edit_working.php
+++ b/modules/Planner/units_edit_working.php
@@ -126,7 +126,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Planner/units_edit_working
 
     $form->addHiddenValue('address', $session->get('address'));
 
-    $form->addHeaderAction('add', __('Add'))
+    $form->addHeaderAction('add', __('Add Lessons'))
         ->setURL('/modules/Planner/units_edit_working_add.php')
         ->addParams($urlParams)
         ->displayLabel();


sgillott:
Resolved the conflict in CHANGELOG.txt

modules/Planner/planner_view_full.php
```diff
@@ -1172,7 +1172,25 @@
...
+                            // BEGIN: Guard against a crash for the missing method
+							// Before (crashes if method doesn't exist):
+							// $offTimetableStudents = $container->get(\Gibbon\Domain\School\SchoolYearSpecialDayGateway::class)
+							//     ->selectOffTimetableStudentsByClass($session->get('gibbonSchoolYearID'), $gibbonCourseClassID, $values['date'])
+							//     ->fetchKeyPair();
+
+							$offTimetableStudents = [];
+							$specialDayGateway = $container->get(\Gibbon\Domain\School\SchoolYearSpecialDayGateway::class);
+
+							if (method_exists($specialDayGateway, 'selectOffTimetableStudentsByClass')) {
```
SKuipers: Hmm. If your Gibbon install is fully up to date, this method_exists check shouldn't be necessary, as the method does indeed exist. The SchoolYearSpecialDayGateway is included with a `use` statement at the top of the file, so it shouldn't need fully qualified.

modules/Planner/planner_view_full.php
```diff
@@ -1312,7 +1330,13 @@
                             $form->addRow()->addSubmit()->addClass('mt-2');
                         }
 
-                        $page->addSidebarExtra($form->getOutput());
+                        // Hide roster for Students & Parents
+						$roleCat = $session->get('gibbonRoleIDCurrentCategory'); // 'Staff' | 'Student' | 'Parent' | etc.
```
SKuipers: Thanks for the suggested change. In this case, rather than manually checking role categories, the built-in permission system would offer better control over view permissions.

SKuipers (commented):
Hi @sgillott, thanks for your suggested change. I can see where schools may wish to hide these participant details for privacy reasons. However, rather than making them hard-coded changes, I'd be inclined to add them as permission checks, that way a school can choose which permissions to give. For example, they may want students to be able to see participants but not parents. There could be cases such as parent helpers, where a school may want certain parent roles to see participants, and other parent roles to not see them. 

As you can see, the planner_view_full.php script is very legacy! I appreciate your efforts to wade into the script and make changes.

sgillott:
> Hi @sgillott, thanks for your suggested change. I can see where schools may wish to hide these participant details for privacy reasons. However, rather than making them hard-coded changes, I'd be inclined to add them as permission checks, that way a school can choose which permissions to give. For example, they may want students to be able to see participants but not parents. There could be cases such as parent helpers, where a school may want certain parent roles to see participants, and other parent roles to not see them.
> 
> As you can see, the planner_view_full.php script is very legacy! I appreciate your efforts to wade into the script and make changes.

Hi @SKuipers thank you for your time reviewing this and for the feedback. If it is okay, I'd like to try again and write something that adds some permissions into the Planner module for this to be toggle-able. Germany has a very strict implementation of GDPR surrounding consent for photographs, and it would be a showstopper for us (after just moving entirely to Gibbon for our SIS) if we aren't able to turn this panel off.

So if it is okay, I'd like to propose that I work on:
- Adding a new planner permission to "Show Participants Panel" , default on for Adm, Prt, Std, Tcr to not affect current behaviour for upgrades, but allowing schools to choose to turn it on or off for different user types.
- Gating the show roster code in modules/Planner/planner_view_full.php with the built-in permissions system as you suggested.

I'll leave out the method_exists check in the next pull request and just focus on the above - I'll need to investigate that method call missing further because it broke the page for me, even though it was a brand new install of v29.

Best wishes,
Steve.

SKuipers:
Hi Steve, thanks for your willingness to dig deeper and work on these changes. Your approach sounds good. These are the existing Lesson Planner permissions (which are grouped and accessed with `$highestAction`):
Screenshot 2025-09-02 at 2 56 16 PM

- Lesson Planner_viewAllEditMyClasses is used by teachers and Lesson Planner_viewEditAllClasses is used by teachers and admins with full planner access. These permissions should be able to see the full participant list.
- Lesson Planner_viewMyChildrensClasses is used for parent access. This should perhaps only be able to see the teacher but no one else. This would be the same as other areas of the system, like Departments and Classes.
- Lesson Planner_viewMyClasses is used by students, and should probably also only see the teacher.
- Lesson Planner_viewOnly should probably not see any participants, or perhaps only just the teacher.

Working with the existing permissions, it may be possible to make these tweaks and maintain backwards compatibility without needing to add a new permission. The only permissions that could see the full list would be the _viewAll and _viewEditAll ones. What do you think? Happy to discuss.

sgillott:
@SKuipers 
Thanks Sandra,
That's a good idea to keep it simple and not add any additional permissions to start off with. If you're happy, I'll work on those changes you've suggested. 😊
Steve.

SKuipers:
Amazing, thank you Steve. Give a shout with any questions. We also have a developer Slack channel you'd be welcome to join, just email support@gibbonedu.org and we'll send an invite. It's a very quiet Slack, but useful to have a place to ask coding questions.

modules/Planner/planner_view_full.php
sgillott: Ah - I see now. That was added in the v30 code, not the v29 I was testing against in production.

PR #1967, opened 2025-09-02
System: Fixed the issue of not being able to use SSO to login

ali-ichk:
**Description**
Fixed the issue to allow a user use Google sign in or SSO to login in a different language or login into a different academic year using the options toggle.


SKuipers (approved):
Thanks!

PR #1968, opened 2025-09-03
fix style on behaviour form

Kachulio1:
**Description**

This PR fixes an issue where, when adding multiple behaviour records, the Remove button would disappear behind the second select at certain browser window widths.

**Motivation and Context**

@rossdotparker noticed the Remove button was playing hide-and-seek behind the second select at certain widths. This change makes sure it stays visible and behaves.

**How Has This Been Tested?**

To test locally navigate to this page  [Home ](http://localhost:8888/core/index.php)> [Behaviour ](http://localhost:8888/core/index.php?q=/modules/Behaviour/behaviour_manage.php)> [Manage Behaviour Records ](http://localhost:8888/core/index.php?q=/modules/Behaviour/behaviour_manage.php)> Add Multiple

	1.	Open Inspect mode.
	2.	Switch to the device toolbar and set the dimension to Responsive.
	3.	Drag the width of the page narrower and wider.
	4.	 Confirm that the Remove button remains visible and does not disappear behind the second select.

**Screenshots**

# Before 
Screenshot 2025-09-03 at 10 26 05 PM

# After 
Screenshot 2025-09-03 at 10 26 21 PM

### DIFF
--- a/modules/Behaviour/css/module.css
+++ b/modules/Behaviour/css/module.css
@@ -18,3 +18,16 @@ You should have received a copy of the GNU General Public License
 along with this program. If not, see <http://www.gnu.org/licenses/>.
 
 */
+
+#gibbonPersonIDMultiContainer {
+    display: flex;
+    flex-wrap: wrap; 
+    gap: 1rem; 
+    align-items: flex-start;
+}
+
+#gibbonPersonIDMultiContainer > .w-full.sm\:w-1\/3 {
+    flex: 2;
+    min-width: 250px; 
+    padding: 0px;
+}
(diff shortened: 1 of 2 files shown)


SKuipers:
Hi @Kachulio1 Thanks for jumping in and making a start on these changes. It looks like your solution is hard-coded to this particular multi-select, however I suspect this visual issue would be common to all multi-select boxes. Would you be interested in tinkering some more and seeing if you can apply a global fix? Rather than adding new CSS, it would ideally be best to adjust the Tailwind classes in the src/Forms/Input/MultiSelect.php file. This set of fields already has a small-size mobile view which stacks:
Screenshot 2025-09-04 at 12 40 41 PM

It may need tweaked so that the medium size stacks with a row of buttons in the middle:
multiselect

Kachulio1:
I have updated the changes

SKuipers (approved):

SKuipers:
Thanks @Kachulio1 for the updated fix, I made a small tweak post-merge to keep the buttons side-by-side, and thanks to your changes they no longer overlap on smaller screens.

PR #1970, opened 2025-09-05
Activities - Added a description to cost fields

ali-ichk:
**Description**

Added another field to the cost section of each activity. Added a description under the estimated cost.


SKuipers (approved):
Thanks Ali, these changes look great.

PR #1972, opened 2025-09-09
System: fixed login issue after public registration

ali-ichk:
**Description** 
fixed login issue of user not getting redirected correctly after public registration


SKuipers (approved):
Thanks 👍

PR #1974, opened 2025-09-10
 Library: added a new Library Type for Sports Kit and added the optio…

ali-ichk:
**Description**
Library: added a new Library Type for Sports Kit and added the option to notify parents when a new item is lended out


SKuipers (approved):
Thanks Ali for these new features. I've made a tweak to remove the Sports Kit code, we will use that internally at ICHK but it may not apply to other schools. I've run that SQL on our server to add the new library item type. The rest looks good, I'll merge and test, thanks!

PR #1976, opened 2025-09-12
Activities: improve the duplicate activity bulk process

ali-ichk:
**Description**
Improved the duplication code to include photos when an activity is duplicated


SKuipers (approved):
Thanks Ali, this looks great, I'll merge and test.

PR #1978, opened 2025-09-15
System admin   new import yml file for gibbon library item event

ali-ichk:
**Description**
 System Admin: added new loaned items import for lIbrary and also a new importer for importing items of type sports kit


SKuipers (approved):
Thanks Ali, I've made a tweak to remove the Sports Kit importer from the core, but have uploaded it to our own server as a custom import.

PR #1984, opened 2025-09-22
System: Added a backup table to store person images

ali-ichk:
**Description**
Add a backup table to store person images
Add a gibbonPersonPhotos table to always have a record when a new picture is uploaded.

Fields:
gibbonPersonPhotosID, gibbonPersonID, gibbonSchoolYearID

Update the bulkimport and user photo uplo


SKuipers (changes requested):
Hi Ali, this is looking great, thanks for putting it together so quickly. Your code is looking great, and I noticed that you've added a unique key to the table, which is good. Because of this key, it's possible to simplify the code to remove the check to see if the record already exists, and use the `insertAndUpdate(array $insertData, array $updateData)` method to accomplish both at the same time. Do you want to see if you can update the code with this change, then request review again and I'll merge and test. Thanks!

ali-ichk:
Hi, Sandra. Thank you for the comments. I have updated the code accordingly. I have tried to look for if there are any other scripts in the system which can update a person's image but have not found anywhere else. Let me know if there are another places I need to add the code to. Thanks.

SKuipers (approved):
Thanks Ali, I'm going to merge and test.

PR #1985, opened 2025-09-24
fix(Forms): Define default $rtl in Toggle.template.php to avoid installer warning

sgillott:
**Description**
This updates src/Forms/Input/Toggle.template.php to ensure the $rtl variable is always defined. 
During installation, language settings (and therefore $rtl) are not yet available, which results in a PHP warning when the toggle component is rendered in the Gibbon installer. This fix adds a default value of false for $rtl, ensuring consistent behaviour in both the installer and at runtime when everything is set up.

**Motivation and Context**
Without this change, fresh installs of Gibbon trigger a warning which is rendered in the installer:
`Warning: Undefined variable $rtl in \src\Forms\Input\Toggle.template.php on line 12`
This should improve the user experience for new installs of Gibbon.

**How Has This Been Tested?**
Tested in a fresh install of Gibbon v30.0.00 on PHP 8.2 using Laragon (Windows).
Verified that the PHP warning no longer occurs.
Verified that toggles continue to render and behave correctly in both installer and normal application views.
Confirmed that RTL languages (when selected after install) still display the toggle correctly.

**Screenshots**
image
image

### DIFF
--- a/src/Forms/Input/Toggle.template.php
+++ b/src/Forms/Input/Toggle.template.php
@@ -9,6 +9,7 @@
     </label>
 
     <?php $disabledClass = $disabled || $readonly ? 'border-dashed cursor-not-allowed' : ''; ?>
+    <?php $rtl = $rtl ?? false; // During installer, $rtl may not be defined; default it to false ?>
     <?php $flip = $rtl ? '-1' : '1'; ?>
     
     <?php if ($toggleSize == 'sm') { ?>


SKuipers (commented):
Thank you, well caught! We hadn't started installation tests for the current version yet. However, generally a warning wouldn't break the flow of code in this way, it looks like you have some extra debugging set up.

In looking at the error, I think a solution in the `Component` class may be more robust, as we ideally want to start improving $rtl handling in all kinds of form fields. Adding `'rtl' => false` to the default $environment variable should help ensure there is a fallback when the session is uninitialized.

SKuipers:
I've made this change in the following commit, which should help resolve the issue going forward: https://github.com/GibbonEdu/core/commit/ea8be3395cc17651ccae1318ae69c0718603356c

sgillott:
Oh, I see! Thanks, Sandra. 😊

SKuipers:
Thanks again, much appreciated! Feel free to keep letting us know about any issues you find. Gibbon has a lot of surface area, so even with daily use and testing, bugs can slip by unnoticed.

PR #1986, opened 2025-09-25
Staff: add an option to accept a Staff Application without creating a…

ali-ichk:
**Description**
Staff: add an option to accept a Staff Application without creating an account


modules/Staff/applicationForm_manage_accept.php
```diff
@@ -121,250 +129,260 @@
...
-                    }
 
-                    //DETERMINE ROLE
-                    $gibbonRoleID = ($values['type'] == 'Teaching') ? '002' : '006';
+                $createStaffAccount = false;
+                if (isset($_POST['createStaffAccount'])) {
+                    if ($_POST['createStaffAccount'] == 'on') {
+                        $createStaffAccount = true;
+                    }
+                }
```
SKuipers: This could be simplified into a single line, since checkboxes are empty when not set.

```suggestion
                $createStaffAccount = !empty($_POST['createStaffAccount']);
```

PR #1990, opened 2025-10-02
Timetable Admin: fixed bug to allow timetable deletion

ali-ichk:
**Description**
Timetable Admin: fixed bug to allow timetable deletion


SKuipers (approved):
Thanks Ali for the fix 👍

PR #1993, opened 2025-10-08
Set Japanese and Portuguese as active language options

ali-ichk:
**Description**
Set the 2 languages as active


SKuipers (approved):
Thanks Ali.

PR #1999, opened 2025-10-17
User Admin: add an option to view images from gibbonPersonImage

ali-ichk:
**Description**
User Admin: add an option to view images from gibbonPersonImage. Users can download and re-upload them if they wish to.


SKuipers (approved):
Thanks Ali 👍

PR #2000, opened 2025-10-23
System: removed duplicate creation of gibbonAction records

ali-ichk:
**Description**

 removed duplicate creation of gibbonAction record for Manage Events_all for Calendar Module


SKuipers (changes requested):
Hi Ali, we can't remove lines from the CHANGEDB, because this will change the line numbering for everyone who has already updated. They need to be repalaced with a dummy statement, like `SELECT NULL;end` so that they do not run if they have not already been run.

PR #2011, opened 2025-10-28
Create studentFutureAbsence.yml

MrCSanders:
Importer for future student absences

**Description**
Import process to batch future non-consecutive absences.

**Motivation and Context**
Students are often absent for non-consecutive days, sometimes for weeks or months in advance. This allows administrators to set future absences quickly without needing to use the GUI.

**How Has This Been Tested?**
t's been run on 29.0, PHP 8.1.2, MySQL 8.0.42. No issues when testing.

**Screenshots**

Screenshot 2025-10-28 at 13 29 29
Screenshot 2025-10-28 at 13 29 58

### DIFF
--- a/resources/imports/studentFutureAbsence.yml
+++ b/resources/imports/studentFutureAbsence.yml
@@ -1,0 +1,92 @@
+details:
+  type: studentFutureAbsence
+  category: Attendance
+  name: Future Absences
+  table: gibbonAttendanceLogPerson
+  modes:
+    insert: true
+    update: false
+
+access:
+  module: Attendance
+  action: Set Future Absence
+
+primaryKey: gibbonAttendanceLogPersonID
+
+table:
+  gibbonPersonID:
+    name: "Student"
+    desc: "Username"
+    args:
+      filter: nospaces
+      required: true
+    relationship:
+      table: gibbonPerson
+      key: gibbonPersonID
+      field: username
+
+  date:
+    name: "Date"
+    desc: "YYYY-MM-DD"
+    args:
+      filter: date
+      required: true
+
+  gibbonAttendanceCodeID:
+    name: "Attendance Code (short, e.g. PS, A, etc.)"
+    desc: "Looks up the code ID"
+    args:
+      filter: string
+      required: true
+    lookup:
+      table: gibbonAttendanceCode
+      key: nameShort
+      value: gibbonAttendanceCodeID
+
+  type:
+    name: "Type"
+    desc: "Eg. Absent, Excused Absence"
+    args:
+      filter: string
+      required: true
+
+  reason:
+    name: "Reason"
+    desc: "From School Admin > Attendance Settings"
+    args:
+      filter: string
+      required: false
+
+  comment:
+    name: "Comment"
+    desc: "Optional note"
+    args:
+      filter: string
+      required: false
+
+  context:
+    name: "Context"
+    desc: "Must be Future"
+    args:
+      filter: string
+      required: true
+    default: Future
+
+  direction:
+    name: "Direction"
+    desc: "Must be Out for absences"
+    args:
+      filter: string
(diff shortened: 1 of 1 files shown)


SKuipers (changes requested):
Hi @MrCSanders, thanks for your contribution, which looks like a useful addition. I notice you're using the indented object syntax, which is a bit more verbose and can make nested arguments a little harder to read. Although both syntaxes are indeed correct for YML, since we're already using the more concise object syntax, this would make it more consistent and easier to maintain. Would you be able to update these? I've added a highlight to the code with an example. Thanks!

resources/imports/studentFutureAbsence.yml
```diff
@@ -0,0 +1,92 @@
...
+    args:
+      filter: string
+      required: true
+
+  reason:
+    name: "Reason"
+    desc: "From School Admin > Attendance Settings"
+    args:
+      filter: string
+      required: false
```
SKuipers: In this case, just one line does the trick, since false is the default for a boolean value.
```suggestion
    args: { filter: string }
```

resources/imports/studentFutureAbsence.yml
```diff
@@ -0,0 +1,92 @@
...
+      required: true
+    default: Future
+
+  direction:
+    name: "Direction"
+    desc: "Must be Out for absences"
+    args:
+      filter: string
+      required: true
+    default: Out
```
SKuipers: It is possible to hide a field and provide the value for it, for these cases. One of the many undocumented features of Gibbon 😅 
```suggestion
  direction:
    name: "Direction"
    desc: ""
    value: "Out"
    args: { filter: string, hidden: true }
```

resources/imports/studentFutureAbsence.yml
```diff
@@ -0,0 +1,92 @@
...
+  date:
+    name: "Date"
+    desc: "YYYY-MM-DD"
+    args:
+      filter: date
+      required: true
+
+  gibbonAttendanceCodeID:
+    name: "Attendance Code (short, e.g. PS, A, etc.)"
+    desc: "Looks up the code ID"
```
SKuipers: Wherever possible, try to re-use the same strings from Gibbon, as this prevents creating new translation strings (of which there are already several thousand 😅 ) Especially for the description fields, which some of them are automatically generated by the table and the args.

```suggestion
    name: "Attendance Status"
    desc: "Short Name"
```

SKuipers:
I've added some feedback, but it's nothing you could have known in advance, just some tips for consistency and reducing the strings used. I greatly appreciate the contribution. Thanks again!

MrCSanders:
I think I've cleaned everything up, thanks for the feedback. I've also broadened the inputs to username/email/studentID as I find myself more easily remembering studentIDs than usernames.

Guidance is appreciated, as this is my first contribution!

rossdotparker:
It's really nice to see @MrCSanders moving from certification to contribution, and getting to know @SKuipers (who does so much of the Gibbon heavy lifting these days : ). Thanks for your contribution Christopher!

rossdotparker:
PS, @MrCSanders we have a developer channel on Slack: let us know if you fancy joining us there.

MrCSanders:
> PS, @MrCSanders we have a developer channel on Slack: let us know if you fancy joining us there.

I have never used Slack but I am aware of its existence. I've created an account and would be happy to join and see how I can contribute.

SKuipers (approved):
Wonderful, thanks @MrCSanders I'm going to merge and test it out.

SKuipers:
I've tested it out and it works great 👍  I made a quick tweak that uses the attendance status field to fetch the gibbonAttendanceCodeID in the background, just some fancy wiring of the fields to require one less piece of data in the importer. Thanks again!

PR #2012, opened 2025-11-03
System Admin: added a new migration file to add the images to the image backup table

ali-ichk:
**Description**
Now that we have a gibbonPersonImage table, it would be good to populate it with all of the existing images in the gibbonPerson table (for any active users). This way, the current year is preserved going forward.


SKuipers (approved):
Wonderful, this looks great, I'll merge and test.

PR #2015, opened 2025-11-07
Library: Updated the search by ID in the lending page to fix issues r…

ali-ichk:
**Description**
Some records in the table were causing errors as they were not in the right JSON format. Updated the code to be able to work with these exceptions.


SKuipers (approved):
Thanks, well caught

PR #2016, opened 2025-11-10
Markbook: Updated the number template to increase the width 

ali-ichk:
**Description**

Fixed the issue of box being too narrow to see the marks being typed.


SKuipers:
Thanks Ali, can you add a CHANGELOG entry, then I'll merge.

PR #2018, opened 2025-11-10
User admin   increase the username field length

ali-ichk:
**Description**
Increase the username field length to support upto 40 characters


SKuipers (approved):
Looking great, thanks!

PR #2022, opened 2025-11-12
Calendar: Updated the Notify staff feature to a background process

ali-ichk:
**Description**
Updated the Notify Staff feature to be a background process


SKuipers (approved):
Wonderful, thanks Ali.

PR #2028, opened 2025-11-18
Calendar: Updated the code to use DataTable to view event details 

ali-ichk:
**Description**
Changed the code to use DataTable instead of form to view event details


SKuipers (approved):
Thanks!

PR #2031, opened 2025-11-23
Attendance: add attendance history ordering option

sgillott:
**Description**
This PR adds a new option to the Attendance module allowing schools to choose how student attendance history is ordered. 
A new setting in School Admin --> Attendance Settings, called **Attendance History Ordering**, lets schools display attendance history entries either:
- In the order attendance was taken ('timestamp order' - Gibbon default behaviour for v30.0.00, still set as default), or
- In timetable period order (new option to show attendance listed in the timetable period order)

The change includes updates to the Attendance History data loader, view, Twig template, School Admin settings, and the relevant gateway. A database migration and changelog entry are included.

**Motivation and Context**
The existing attendance history display _could_ be confusing when records were not shown in timetable order, for example, if teachers forgot to record attendance during the lesson and went back to submit the attendance register much later (or even on a different day - see screenshots below). A request was made from our school leadership to be able to see student attendance history in order for each lesson/period, which could also assist them in seeing where attendance registers had not been taken more efficiently. This PR introduces a configurable option so each school can select the approach that best fits their workflow (original, unchanged behaviour, or new period ordering).

This improves usability without altering any core attendance logic and preserves backwards compatibility (default behaviour remains as the timestamp order, as in Gibbon v30.0.00, so existing schools would need to consciously switch to the new view).

**How Has This Been Tested?**
Tested on a clean Gibbon v31.0.00 installation with demo data using Laragon (PHP 8.2, MariaDB), then again with a copy of our live school database on a test Ubuntu server running Apache2.0 (PHP 8.3, MariaDB)

Verified that:
- The new setting appears under School Admin --> People --> Attendance Settings.
- Changing the setting updates attendance ordering correctly in People --> Students --> Attendance History.
- The feature functions correctly across multiple school timetables (e.g. different lower and upper school timetables in our case) and with partial attendance data.
- Confirmed that the migration adds the new setting successfully on upgrade page.

**Screenshots**
Screenshot 2025-11-22 215437

New configuration option in School Admin --> People --> Attendance Settings.

Screenshot 2025-11-23 122130 Screenshot 2025-11-22 214806

Examples of original Gibbon view/behaviour with default "Ordered by time attendance taken"

Screenshot 2025-11-23 125221 Screenshot 2025-11-23 125323

Examples of new view/behaviour for "Ordered by timetable period" (for same days/data as above)

Maintains original colours, adds grey for registers not taken indicator.

### DIFF
--- a/src/Domain/Attendance/AttendanceLogPersonGateway.php
+++ b/src/Domain/Attendance/AttendanceLogPersonGateway.php
@@ -110,7 +110,14 @@ class AttendanceLogPersonGateway extends QueryableGateway
         return $this->runQuery($query, $criteria);
     }
 
-    public function selectAllAttendanceLogsByPerson($gibbonSchoolYearID, $gibbonPersonID, $attendanceOrdering = 'timestamp')
+	/**
+     * Select all attendance logs for a person within a school year.
+     *
+     * @param string $gibbonSchoolYearID
+     * @param string $gibbonPersonID
+     * @return \Gibbon\Domain\DataSet|array
+     */
+    public function selectAllAttendanceLogsByPerson($gibbonSchoolYearID, $gibbonPersonID)
     {
         $query = $this
             ->newSelect()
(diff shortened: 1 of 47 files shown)


src/Domain/Attendance/AttendanceLogPersonGateway.php
```diff
@@ -110,26 +110,85 @@ public function queryClassAttendanceByPersonAndDate(QueryCriteria $criteria, $gi
...
+				]);
+			} else {
+				$query->orderBy(['timestampTaken ASC']);
+			}
+
 
         return $this->runSelect($query);
     }
+	
+	public function selectTimetablePeriodsByPersonAndDate($gibbonSchoolYearID, $gibbonPersonID, $date)
```
SKuipers: This would be best added to the TimetableDayDateGateway, even though it means adding another gateway to the constructor, as the method itself is unrelated to attendance specifically, and it will prevent duplicate functions being added in different gateways. Luckily, its safe to change the constructor for these classes, the Container takes care of dependencies.

SKuipers (changes requested):
Hi Steve, thank you for your work adding this improvement and sharing a detailed PR with descriptions and screenshots. I've had a look at your code changes and it looks good, with one minor comment on the gateway class.

In the interest of preventing new settings and code complexity, I'd be happy for this to become the new default ordering for the student history page. The periods are left-joined, so in the absence of that data, it will still fallback to the timestamp. The time in brackets is nice and readable, and it makes this view more closely match how the same data is presented in the attendance log in Attendance by Person. 

Would you be okay to update this PR to remove the setting and set this as the new default? (no need to close and re-open it, just push new changes to the same branch). Thanks!

sgillott:
Thanks, Sandra! I hope these changes are okay. I've tested it again, but please let me know if you need me to do anything else.

SKuipers (approved):
Thanks @sgillott for the updates, it's looking great 👍  I'm going to merge it in and test it out.

SKuipers:
Hi @sgillott I've tested it out and things are looking good. I've made a couple post-merge tweaks:

- Fixed periods for classes students have left showing in the student history
- Fixed classes with attendance disabled showing in the student history
- Changed "Not Taken" to "Not Available", which is an existing string and suggests the data may yet be pending
- Don't display attendance that was not taken when a student has a school-wide absence, since attendance would not be expected for those classes

Have a look at the changes and let me know if this causes any issues with your intended use case. You can always string-replace "Not Available" back to "Not Taken" for your system if that makes more sense to your staff.

sgillott:
Thanks Sandra, that looks great, and sorry about forgetting those edge cases whilst testing.

SKuipers:
Thanks, no worries at all, there's always lots of edge cases to account for! It turns out we may need to add a setting after all. In testing it with my own school's data, we noticed that this now shows classes where attendance may not have been taken for all previous school days. While it's good to catch missing attendance on the day, it's not so great if suddenly parents can see any small gaps in class attendance in the past. We use school-wide attendance as the overall status, so occasionally class attendance may be missed for a variety of reasons. 

I wonder if, based on your recent experience, you'd have the interest in adding this? We'd be looking to add a setting like "Show incomplete attendance" and use that to determine whether to show or hide incomplete classes. What do you think?

sgillott:
Hi Sandra, yes - I can look at that. I can wrap some logic around the synthetic logs to not get produced if the show incomplete attendance setting is false.
Would you also want an additional setting to toggle show incomplete attendance for staff only or everyone (staff/students/parents), or just hard code it so that students/parents never see incomplete attendance even if 'show incomplete attendance' is true?

sgillott:
Currently looking like this:
https://github.com/sgillott/GibbonEduCore/commit/fdc8371bce25af7b9c7a9bcacb739e78c9f14f69

SKuipers:
Thanks Steve, this looks like a good approach and doesn't add too much complexity to your existing changes.

sgillott:
https://github.com/GibbonEdu/core/pull/2066 resolves this.

PR #2038, opened 2026-01-08
Messenger: fixed the issue of incorrect name being displayed for mailing list recipients

ali-ichk:
**Description**
Messenger: fixed the issue of incorrect name being displayed for mailing list users in Report Data


modules/Messenger/messenger_manage_report.php
```diff
@@ -334,19 +346,19 @@
...
+                $sql = "SELECT COALESCE(gibbonPerson.surname, gibbonMessengerMailingListRecipient.surname) AS surname, COALESCE(gibbonPerson.preferredName, gibbonMessengerMailingListRecipient.preferredName) AS preferredName, gibbonPerson.gibbonPersonID, gibbonMessenger.*, gibbonMessengerReceipt.*, COALESCE(gibbonRole.category, 'External') AS roleCategory
+                    FROM gibbonMessengerReceipt
+                    JOIN gibbonMessenger ON (gibbonMessengerReceipt.gibbonMessengerID=gibbonMessenger.gibbonMessengerID)
+                    LEFT JOIN gibbonPerson ON (gibbonMessengerReceipt.gibbonPersonID=gibbonPerson.gibbonPersonID)
+                    LEFT JOIN gibbonRole ON (gibbonRole.gibbonRoleID=gibbonPerson.gibbonRoleIDPrimary)
+                    LEFT JOIN gibbonMessengerMailingListRecipient ON (gibbonMessengerReceipt.targetType = 'Mailing List' AND ((gibbonMessengerReceipt.unsubscribeKey IS NOT NULL AND gibbonMessengerReceipt.unsubscribeKey <> '' AND gibbonMessengerMailingListRecipient.key = gibbonMessengerReceipt.unsubscribeKey) OR (gibbonMessengerMailingListRecipient.email = gibbonMessengerReceipt.contactDetail)))
```
SKuipers: Hmm, this is a pretty complex JOIN statement, which could cause some optimization issues down the road as the mailing list gets longer. From what I understand, the `ON` is evaluated before filtering data, and a WHERE is after, so a complicated `ON` can make queries much slower.

I wonder, would `gibbonMessengerMailingListRecipient.email = gibbonMessengerReceipt.contactDetail AND gibbonMessengerReceipt.targetType = 'Mailing List'` work in this case, since we only care about emails actually sent? Lets put the email first too, so that an INDEX could be added to gibbonMessengerMailingListRecipient as necessary in the future, to speed up queries.

modules/Messenger/messenger_manage_report.php
```diff
@@ -334,19 +346,19 @@
...
-                        FROM gibbonMessengerReceipt
-                        JOIN gibbonMessenger ON (gibbonMessengerReceipt.gibbonMessengerID=gibbonMessenger.gibbonMessengerID)
-                        LEFT JOIN gibbonPerson ON (gibbonMessengerReceipt.gibbonPersonID=gibbonPerson.gibbonPersonID)
-                        LEFT JOIN gibbonRole ON (gibbonRole.gibbonRoleID=gibbonPerson.gibbonRoleIDPrimary)
-                        WHERE gibbonMessengerReceipt.gibbonMessengerID=:gibbonMessengerID ORDER BY FIELD(confirmed, 'Y','N',NULL), confirmedTimestamp, surname, preferredName, contactType";
-                    $result = $connection2->prepare($sql);
-                    $result->execute($data);
+                
+                $data = ['gibbonMessengerID' => $gibbonMessengerID];
+                $sql = "SELECT COALESCE(gibbonPerson.surname, gibbonMessengerMailingListRecipient.surname) AS surname, COALESCE(gibbonPerson.preferredName, gibbonMessengerMailingListRecipient.preferredName) AS preferredName, gibbonPerson.gibbonPersonID, gibbonMessenger.*, gibbonMessengerReceipt.*, COALESCE(gibbonRole.category, 'External') AS roleCategory
```
SKuipers, line 343: Nice use of COALESCE 😄

modules/Messenger/messenger_manage_report.php
```diff
@@ -251,25 +267,21 @@
...
                         $parent1Name = Format::name('', $recipient['parent1preferredName'], $recipient['parent1surname'], 'Parent', true);
                         $parent2Name = Format::name('', $recipient['parent2preferredName'], $recipient['parent2surname'], 'Parent', true);
 
-                        //Tests for row completion, to set colour
-                        $studentComplete = isset($receipts[$recipient['gibbonPersonID']]) && $receipts[$recipient['gibbonPersonID']]['confirmed'] == "Y";
-                        $parent1Complete = (isset($receipts[$recipient['parent1gibbonPersonID']]) && $receipts[$recipient['parent1gibbonPersonID']]['confirmed'] == "Y");
-                        $parent2Complete = (isset($receipts[$recipient['parent2gibbonPersonID']]) && $receipts[$recipient['parent2gibbonPersonID']]['confirmed'] == "Y");
+                        // Tests for row completion, to set colour
+                        $studentComplete = (!empty($recipient['gibbonPersonID']) && isset($receipts[$recipient['gibbonPersonID']]) && $receipts[$recipient['gibbonPersonID']]['confirmed'] == "Y");
```
SKuipers: !empty includes isset, so when switching one to the other, we can remove the isset

modules/Messenger/messenger_manage_report.php
```diff
@@ -210,28 +228,26 @@
...
                     $recipients = array_filter($recipients, function($recipient) use (&$receipts) {
-                        if (array_key_exists($recipient['gibbonPersonID'], $receipts)) {
+                        if (!empty($recipient['gibbonPersonID']) && array_key_exists($recipient['gibbonPersonID'], $receipts)) {
                             return true;
                         }
 
-                        if (array_key_exists($recipient['parent1gibbonPersonID'], $receipts)
-                        && (is_null($receipts[$recipient['parent1gibbonPersonID']]['gibbonPersonIDListStudent']) || in_array($recipient['gibbonPersonID'], $receipts[$recipient['parent1gibbonPersonID']]['gibbonPersonIDListStudent']))) {
-                                return true;
+                        if (!empty($recipient['parent1gibbonPersonID']) && array_key_exists($recipient['parent1gibbonPersonID'], $receipts) && (is_null($receipts[$recipient['parent1gibbonPersonID']]['gibbonPersonIDListStudent']) || in_array($recipient['gibbonPersonID'], $receipts[$recipient['parent1gibbonPersonID']]['gibbonPersonIDListStudent']))) {
```
SKuipers: !empty includes an array_key_exists check, so we shouldn't need both 🤔

modules/Messenger/messenger_manage_report.php
```diff
@@ -184,17 +183,36 @@
...
+                $receipts = [];
+                $receiptsByReceiptID = [];
+                $externalReceipts = [];
+
+                foreach ($receipetRows as $receipetRow) {
+                    $receipetRow['gibbonPersonIDListStudent'] = empty($receipetRow['gibbonPersonIDListStudent'])
+                        ? null
+                        : explode(',', $receipetRow['gibbonPersonIDListStudent']);
+
+                    $receiptsByReceiptID[$receipetRow['gibbonMessengerReceiptID']] = $receipetRow;
```
SKuipers: This array is created, but I don't see where it is used? 🤔

SKuipers (changes requested):
Thanks Ali, this is looking good overall, especially since the messenger is a complex beast. I've added a few notes and spots to reduce complexity.

SKuipers (approved):
Thanks for the changes, this is looking ready to merge.

PR #2039, opened 2026-01-08
User Admin: fix the bug to display user's personal address if turned on

ali-ichk:
**Description**
Fix the bug to display user's personal address if turned on


SKuipers (approved):
Thanks Ali, well caught.

PR #2040, opened 2026-01-09
Library: updated automatic shelf lists so that they are queried dynamically

ali-ichk:
**Description**
Library: updated automatic shelf lists so that they are queried dynamically so that no books can be added manually to an automatic list.  No data table is displayed for automatic list.


SKuipers (approved):
Thanks Ali, this looks good, I'll merge and test.

The `->placeholder('Please select...')` isn't necessary, as the text is automatic when the field is required, but it looks like this came from the original code, so I'll fix it in both post-merge.

PR #2041, opened 2026-01-13
System admin: validate email templates and show error message

ali-ichk:
**Description**
System Admin: added a validation to check for error in email twig templates


SKuipers (approved):
Excellent, thanks Ali, looks like some useful error checking to have.

PR #2042, opened 2026-01-13
Staff: updated the staff absence message for partial day absence

ali-ichk:
**Description**
Staff: updated the staff absence message for partial day absence


SKuipers (approved):
Thanks Ali

PR #2044, opened 2026-01-14
Reports: added error detection while editing twig templates

ali-ichk:
**Description**
Reports: added error detection while editing twig templates


SKuipers (approved):
Excellent, thanks, great to have error checking for this page too.

PR #2045, opened 2026-01-16
Library: Updated the null value of gibbonLibraryTypeID in the table

ali-ichk:
**Description**
Library: Updated the null value of gibbonLibraryTypeID in the table


SKuipers (approved):
Thanks!

PR #2048, opened 2026-01-22
Students: view student attendance

ali-ichk:
**Description**
At the moment we need to go into attendance to view a student's attendance and location. Thus, added a "where is this student right now" indicator at the top of the profile, that cross-references the various places and reasons a student could be somewhere to give you a quick answer.


SKuipers (approved):
Thanks for putting this together Ali, and for learning even more about the timetable and attendance systems in the process 😄

PR #2051, opened 2026-01-26
Reports: improved the file path checking while managing archives

ali-ichk:
**Description**
Reports: improved the file path checking while managing archives and uploading bulk reports


SKuipers (approved):
Thanks 👍

PR #2055, opened 2026-01-29
Formal Assessment: added the feature for teachers to view unpublished results

ali-ichk:
**Description**
Formal Assessment: added the feature for teachers to view unpublished internal Assessments


SKuipers (approved):
Looks good, thanks Ali for the fix. This looks like an area to eventually refactor into a DataTable.

PR #2054, opened 2026-01-28
Add Extra Information to Off-Schedule Lessons in Timetable

fvlasie:
**Description**
Enables the display of teacher, class and location information for lessons that are not associated with a timetable slot. Currently off-schedule lessons are treated as second class citizens. This PR solves that frightful inequity! :D

**Motivation and Context**
The off-schedule lessons stood out in the timetable viewer as not having class or course information. 

**How Has This Been Tested?**
Tested in my local MAMP dev environment with Gibbon 31.

**Screenshots**
You can see that Lesson type entry has course and class codes, location, and teacher information.
Screenshot 2026-01-28 at 11 31 14 AM

**I had to modify the DB**
Sorry, I am not sure how to submit core SQL changes so I am putting the SQL here:

`ALTER TABLE gibbonPlannerEntry
    ADD COLUMN gibbonSpaceID INT(10) UNSIGNED NULL AFTER gibbonUnitID;

ALTER TABLE gibbonPlannerEntry
    ADD CONSTRAINT fk_planner_space
        FOREIGN KEY (gibbonSpaceID)
        REFERENCES gibbonSpace (gibbonSpaceID)
        ON DELETE SET NULL;
`

PR #2057, opened 2026-01-30
Revised Add Extra Information to Off-Schedule Lessons in Timetable

fvlasie:
New PR to replace #2054 

Description
Enables the display of teacher, class and location information for lessons that are not associated with a timetable slot. Currently off-schedule lessons are treated as second class citizens. This PR solves that frightful inequity! :D

Motivation and Context
The off-schedule lessons stood out in the timetable viewer as not having class or course information.

How Has This Been Tested?
Tested in my local MAMP dev environment with Gibbon 31.


fvlasie:
While attempting to update the PR to use  addSelectSpace() I discovered that the values would not save in the Planner module.

It seems the JS bindings are not active for addSelectSpace() in Planner.

    <option value=""></option>

This element is never updated so the submitted value is always empty.

SKuipers (approved):
Thanks FV for the updated PR, I'll merge it and have a look and see about the addSelectSpace, it should be working in v31. There have been code changes for those selects, I wonder if you may need to clear cache locally or via Cache Manager if you are running cutting edge code? 🤔

SKuipers:
Hi FV, I've made a couple [tweaks post-merge](https://github.com/GibbonEdu/core/commit/8d325b7cac974f8c06aedc047fff41beae7f2344). It turned out the `ZEROFILL` was missing from the `gibbonSpaceID` column, so it wasn't comparing to the correct value in the `addSelectSpace`. I've added this to the CHANGEDB and updated the selector.

I've also made some tweaks to the logic so that the Location option only shows up if the lesson is not tied to the timetable. Otherwise, this creates confusion in the existing logic for handling timetabled rooms, bookings and space changes. To do this, it now stores the `gibbonTTDayRowClassID` quietly in the background if it finds TT data for that lesson (something that would have made life much easier in the TT if it did from the start 😅 ). 

Have a look at the changes and let me know if you have any questions.

fvlasie:
Thanks Sandra! The tweaks are perfect.

PR #2058, opened 2026-01-30
Staff: View staff attendance status

ali-ichk:
**Description**
Added an indicator at the top of view staff profile to indicate the attendance status and current location location of the staff


SKuipers (approved):
Thanks Ali, it'll be really handy to be able to see where someone is at a glance 😃

PR #2060, opened 2026-02-04
CSV  import for Planner

fvlasie:
Added an Import button to Planner for importing of lesson plans via CSV.

**Description**
The CSV import feature modifies the Planner module to allow the import of lessons via CSV. The importer uses the form building logic from planner_add.php to produce a CSV template that matches the fields presented in the Planner UI. This provides CSV support for custom fields such as those inserted by 3rd party modules. It should also make the CSV robust enough to handle future changes to the Planner UI. 

Bonus: This PR also modifies the new Location field for lesson plans to act as an override for the timtetable data. There are cases where (in my experience) specific lessons need to move to different locations even if the class is usually taught in the same place. Hiding the Location field for these classes makes it impossible to set a location for off-timetable lessons.

Having the option act as an override makes its use clear to end users so there is no longer any confusion. If the field is set unnecessarily it does not introduce any issues.

**Motivation and Context**
For classes that do not fit well into Gibbon's timetable scheme the only option was to enter the lessons manually. This PR provides a semi-automatic way to create lessons with any timing needed.

**How Has This Been Tested?**
Local dev environment (MAMP)

**Screenshots**
Screenshot 2026-02-03 at 4 05 22 PM


rossdotparker:
Hi FV, thanks for the time and thought you've put into this new feature, which looks to have loads of promise. For any change of more than a few lines, one of the first questions we ask ourselves is how much will this add to Gibbon's maintenance burden (which is already significant!)? In this case, with Gibbon already having an importer, my first thought is how much overlap might there be between what you've added and what exists in Import From File. I wonder if there might be a way to draw more on that existing code? I know that @SKuipers will weigh in shortly, but for now that is my initial take. As always, your input is greatly appreciated : )

fvlasie:
Hello Ross! This is a slightly different animal from Import From File since it is designed to automate planner_add.php's functionality in a limited scope.

SKuipers:
Hi FV, I can appreciate that you're looking to add a front-end button for the import functionality. As Ross mentioned, any new piece set of code is something to not only add but maintain in the future. For open source, I use the analogy of a free puppy: they're cute, but you still have to feed and walk them for the rest of their life 😅 

In that regard, I notice there are some very unusual coding choices in this PR, which don't quite align with how Gibbon generally works. May I ask if this code was written by AI? Otherwise, while it is quite creative ... it should be possible to re-use the existing classes and methods for handling CSV data imports to reduce the duplication of functionality.

fvlasie:
Hello Sandra, I am using Claude to help me code because I don't know all the ins and outs of Gibbon and php, is that alright? Would you like me to move this feature to Import From File as Ross suggested? I am happy to make any changes that you require.

SKuipers:
Hi FV, thanks for the update, I had wondered about that when looking at the way the code was structured. AI generated code can be useful for simple cases, but it doesn't tend to generate code that is maintainable long term. In this case, the methods it's taken to process CSV data don't look ideal (it's piping PHP functions directly into regex patterns). 

Using the existing Import from File system would be helpful, since that importer can already handle different types of spreadsheet files (not just csv) and it can do field-by-field error checking, as well as step-by-step import with a dry run and live fun. Adding a new import is a matter of creating a new .yml file in the resources/imports folder, so the coding overhead is low. 

Import files also have an Access setting based on existing Gibbon permissions, which enables giving import access for teachers to use the importer without having access to the rest of the imports.

PR #2061, opened 2026-02-04
Staff: added the feature to send an automated email for staff applica…

ali-ichk:
**Description**
Staff: added the feature to send an automated email for staff application confirmation


SKuipers (approved):
Thanks Ali, it'll be handy to have templates for these emails.

PR #2066, opened 2026-02-17
Attendance show incomplete toggle

sgillott:
**Description**
Adds a new setting in School Admin --> Attendance Settings called **Show Incomplete Attendance**, which controls whether the “Not Available” class attendance logs appear in the Student History view.

When enabled, incomplete class logs are visible only to users in the **Staff** role category.

Students and Parents _never_ see incomplete logs, regardless of setting.

**Motivation and Context**
From https://github.com/GibbonEdu/core/pull/2031#issuecomment-3615322800 it was identified that displaying incomplete class attendance could expose historical gaps to parents.

Some schools rely on school-wide attendance as the authoritative record, and occasional missing class-level logs should not be visible to non-staff users.

This change:
- Adds a configurable setting.
- Restricts visibility of incomplete logs to Staff role category.
- Fails safe by hiding incomplete logs for all non-Staff categories.

**How Has This Been Tested?**
Tested in a production-like environment with real attendance data.

Scenarios tested:
- Admin (Staff category) → setting Y → incomplete logs visible.
- Admin switched to Parent role → incomplete logs hidden.
- Teacher role → visible.
- Student login → hidden.
- Parent login → hidden.
- Setting N → incomplete logs hidden for all roles.

Role switching via gibbonRoleIDCurrent was verified.

**Screenshots**
Screenshot 2026-02-17 200931
Screenshot 2026-02-17 201014


modules/Attendance/src/StudentHistoryData.php
```diff
@@ -73,6 +77,19 @@ public function getAttendanceData($gibbonSchoolYearID, $gibbonPersonID, $dateSta
...
         $countClassAsSchool = $this->settingGateway->getSettingByScope('Attendance', 'countClassAsSchool');
         $firstDayOfTheWeek = $this->settingGateway->getSettingByScope('System', 'firstDayOfTheWeek');
+		
+		// Get showIncompleteAttendance setting from gibbonSetting
+		$showIncomplete = $this->settingGateway->getSettingByScope('Attendance', 'showIncompleteAttendance');
+		
+		// Determine current user's role category
+		global $session;
+		$role = $this->roleGateway->getByID($session->get('gibbonRoleIDCurrent'));
+		$roleCategory = $role['category'] ?? '';
```
SKuipers: Luckily, the current user's role category is available in `$session->get('gibbonRoleIDCurrentCategory')`, so you don't need to query it manually :)

SKuipers (changes requested):
Hi @sgillott, thanks for making these changes, they look great. I've made one note where we can simplify the code, which prevents the need for an extra gateway call.

SKuipers (approved):
Thanks @sgillott for the updates, this looks good. I've noticed one quick thing, that we can pull the session in from the constructor rather than as a global, but I can make that tweak post-merge 😄

PR #2069, opened 2026-02-25
System: Added Peruvian Sol as a currency option

ali-ichk:
**Description**
System: Added Peruvian Sol as a currency option
It came from : https://ask.gibbonedu.org/t/if-anyone-want-to-add-your-own-currency-please-edit-this-file-and-it-will-appear-your-currency/641/35


SKuipers (approved):
Thanks!

PR #2075, opened 2026-03-12
Library: added a new Library API Key in Library settings to call the …

ali-ichk:
**Description**
Library: added a new Library API Key in Library settings to call the Google Books API


SKuipers (approved):
Thanks!

PR #2037, opened 2025-12-12
feat(docker): application build via docker

zaidmasri:
**Description**
This PR introduces a Docker-based local development environment for Gibbon Core. It modernizes the repository's tooling by providing a standardized containerized setup, eliminating "works on my machine" issues.

The setup is split into two modular configuration files:

  * **`docker-compose.yaml`**: The base configuration defining the core services (`app` and `db`), networks, and volumes suitable for production-like builds.
  * **`docker-compose.dev.yaml`**: A development override that exposes ports (e.g., `8080`) and sets environment variables specifically for local testing.

**Motivation and Context**
Currently, setting up a local Gibbon development environment requires manually configuring PHP, Apache, and MySQL, which can be error-prone and inconsistent across operating systems (macOS, Windows, Linux).

This change solves the following problems:

  * **Simplifies Onboarding:** New contributors can spin up a fully working instance with a single command.
  * **Dependency Management:** Ensures all contributors use the exact same PHP version (8.2) and extensions (`bcmath`, `intl`, `gd`, etc.) required by Gibbon v31+.
  * **Isolation:** Keeps the database and application files isolated from the host machine, preventing system clutter.

**How Has This Been Tested?**
I have tested these changes on macOS (Intel) using Docker Compose CLI and deployed to a production-like environment using Coolify.

**Test Scenarios:**

1.  **Fresh Install:** verified that running the dev command spins up containers, and the Gibbon Web Installer is accessible at `http://localhost:8080`.
2.  **Database Connection:** Verified that the `app` container can successfully communicate with the `db` container using the hostname `db` and default credentials.
3.  **Extension Compatibility:** Confirmed that critical PHP extensions (`bcmath`, `gd`, `zip`) are loaded and functional.
4.  **Deployment (Coolify):** Validated the configuration by deploying via Coolify, ensuring the image builds correctly and runs in a self-hosted production context.

**How to Run (Instructions for Reviewers)**

**Option 1: Local Development (Recommended)**
This command uses the override file to expose the app on port `8080`.

```bash
docker compose -f docker-compose.yaml -f docker-compose.dev.yaml up -d --build
```

  * **Access the app:** http://localhost:8080
  * **Default DB Host:** `db`
  * **Default DB User/Pass:** `gibbon` / `gibbonpass`

**Option 2: Production/Base Build**
To test the standalone base configuration (no ports exposed to host by default, mimics production internal networking):

```bash
docker compose up -d --build
```

**Shutdown:**
[...]

PR #2077, opened 2026-03-17
Docker: Application build via Docker for development

ali-ichk:
**Description**
This PR reviews and adds some changes to the the Dockerfile introduced by Zaidmasri in https://github.com/GibbonEdu/core/pull/2037, resolving some issues that prevented Gibbon from running correctly in a Docker environment.

**Motivation and Context**
Currently, setting up a local Gibbon development environment requires manually configuring PHP, Apache, and MySQL, which can be error-prone and inconsistent across operating systems (macOS, Windows, Linux).

This change solves the following problems:

Simplifies Onboarding: New contributors can spin up a fully working instance with a single command.
Dependency Management: Ensures all contributors use the exact same PHP version (8.2) and extensions (bcmath, intl, gd, etc.) required by Gibbon v31+.
Isolation: Keeps the database and application files isolated from the host machine, preventing system clutter.

**How Has This Been Tested?**
Tested locally on macOS (Apple Silicon) using Docker Desktop 29.2.1 with the following steps:

Built the image: docker compose -f docker-compose.yaml -f docker-compose.dev.yaml up -d --build
Verified all required PHP extensions are present: docker exec gibbon_app php -m
Verified all PHP ini values meet Gibbon's requirements via php -r "echo ini_get(...)"
Verified AllowOverride All is active: docker exec gibbon_app grep AllowOverride /etc/apache2/apache2.conf
Verified Apache config syntax: docker exec gibbon_app apachectl configtest → Syntax OK
Verified Composer dependencies are satisfied: docker exec gibbon_app composer check-platform-reqs
Verified sensitive files ([.htaccess], [composer.json], [phpstan.neon], [gibbon.sql]) return 403
Confirmed Gibbon web installer loads at http://localhost:8888 with all system requirement checks passing
Confirmed database connection succeeds from the installer using the Docker Compose credentials


up.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo "Error: Docker is not installed or not available on PATH."
+    exit 1
+fi
+
+if ! docker info >/dev/null 2>&1; then
+    echo "Error: Docker is not running. Start Docker Desktop and try again."
+    exit 1
+fi
+
+COMPOSE_FILES="--project-directory . -f ops/docker-compose.yaml -f ops/docker-compose.dev.yaml"
```
pli888: Declaring this variable in `.env` using `COMPOSE_FILE=ops/docker-compose.yaml:ops/docker-compose.dev.yaml` might be better. It then allows running commands like an alias of `docker compose ps -a` from the project root which is useful for debugging.

ops/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
+
+WORKDIR /var/www/html
+
+# 4. Copy application source from local context
+# Build context is the project root (see ops/docker-compose.yaml context: ..)
+COPY . .
+
+# 5. Install PHP Dependencies
+RUN composer install --no-dev --optimize-autoloader --no-interaction
```
pli888: Composer downloads and installs the same packages twice; the first time here to bake the PHP packages into the Docker image and the second time in docker-compose.dev.yaml. This feels unnecessary for a dev environment and it increases the waiting time before it is ready for use.

ops/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+# 6. Configure PHP for Gibbon
+# Path is relative to build context (project root), so ops/php/gibbon.ini
+# In dev, this file is volume-mounted so changes take effect without a rebuild.
+COPY ops/php/gibbon.ini /usr/local/etc/php/conf.d/gibbon.ini
+
+# 7. Permissions
+RUN chown -R www-data:www-data /var/www/html \
+    && find /var/www/html -type d -exec chmod 755 {} \; \
+    && find /var/www/html -type f -exec chmod 644 {} \;
```
pli888: This step takes a long time to run (~13 minutes) because all the files and directories of the PHP packages downloaded by composer will be searched too.

ops/Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+    && sed -i 's/AllowOverride None/AllowOverride All/g' /etc/apache2/apache2.conf
+
+# 3. Install Composer
+COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
+
+WORKDIR /var/www/html
+
+# 4. Copy application source from local context
+# Build context is the project root (see ops/docker-compose.yaml context: ..)
+COPY . .
```
pli888: We would need to do this for a production Docker image. But I wonder if this step is required for a dev environment if the source code from your local filesystem is mounted over it in docker-compose.dev.yaml.

up.sh
```diff
@@ -0,0 +1,62 @@
...
+
+case "${1:-up}" in
+    up)
+        echo "Starting Gibbon dev environment..."
+        docker compose $COMPOSE_FILES up -d --build
```
pli888: Having the `--build` flag forces a rebuild of the images every time before containers are started. This will mean waiting 15 mins for images to build before a developer can start work every time they run up.sh. However, rebuilding images are only required when changes have been made to the Dockerfile. Try replacing this line with:
```
docker compose build app db
docker compose up -d
docker compose exec -T app composer install
```

Also comment out steps 4 to 7 in Dockerfile and then run `up.sh`

It will take ~10 mins for up.sh to execute first time, but then it will take less than 10 secs for containers to spin up when up.sh is run the second time and thereafter. This is because built images and their layers are reused. Composer is run once to download and install PHP packages; thereafter Composer will just check to ensure what's listed in composer.lock has already been downloaded in vendor folder.

pli888:
Hi @ali-ichk 

Nice work, I've made some comments to try to speed up image building.

Looking ahead, we should think about how to restore the database with test data and also show how developers can run tests on this Docker-based Gibbon development environment. This can be done in separate PRs though.

ops/docker-compose.dev.yaml
```diff
@@ -0,0 +1,12 @@
...
+        ports:
+            - "8080:80"
+        volumes:
+            ## Mount project root so live edits are immediately reflected without rebuilding the image. This overrides the COPY in the Dockerfile for local development.
+            - .:/var/www/html
+            ## Mount PHP config so values can be changed without an image rebuild.
+            - ./ops/php/gibbon.ini:/usr/local/etc/php/conf.d/gibbon.ini:ro
+        ## In dev, the volume mount hides vendor baked into the image. This command installs dependencies at startup if vendor/ is missing.
+        command: >
+            bash -c "[ ! -d vendor ] && composer install; apache2-foreground"
```
pli888: When I first executed `up.sh`, I found Composer was still installing PHP packages when the shell script finished processing. I think this happens because the gibbon_app container runs on detached mode. However, a developer would assume Gibbon's local docker environment is ready for use when `up.sh` finishes. I think this composer install command would be better placed somewhere else, e.g. in `up.sh`.

ali-ichk:
Hi @pli888 Thank you for your suggestions. I have made some changes based on your comments. If you have a moment, could you have a quick look and see if this looks okay to be merged to the core? Thank you for your time!!

pli888 (approved):
Looks good to me. Happy to approve.

SKuipers (approved):
Thanks @ali-ichk and @pli888 for working on this PR, I'm happy to approve and merge.

PR #2079, opened 2026-03-20
Calendar   remove future absence when deleting participant or editing the date

ali-ichk:
**Description**
Currently, there are a few areas where an event can be changed but the future absences do not get cleaned up:

If the date of the event is changed, all future absences need removed (with a warning message letting the user know)

If a student is removed from the participants list, just their future absence needs removed

If the whole event is deleted, all future absences attached to it need removed


SKuipers (approved):
Thanks Ali, these changed are looking great, I'm going to merge them in now.

PR #2081, opened 2026-03-25
Attendance: added a foreign table field to attendance logs for record…

ali-ichk:
**Description**
 Attendance: added a foreign table field to attendance logs for recording target of future absences


SKuipers (approved):
Looks great, thanks Ali for the quick work!

PR #2083, opened 2026-03-30
Library: fixed the bug that was preventing exporting more than 50 rec…

ali-ichk:
**Description**
When we click “export” in Catalog Summary, not all records are included in the exported file. Only 50 records are recorded, even if we change the number of records to display in the table. When we click "Print" or "export", only the first 50 records are included not all the records


SKuipers (approved):
Well caught!

PR #2084, opened 2026-04-08
Admissions: fix applicant not linked to admissions account on acceptance

brianp-matrix:
## Summary
- The acceptance process in `applications_manage_acceptProcess.php` used `hasResult()`/`getResult()` to check for parent and family records created during acceptance
- However, `CreateParents.php` and `CreateFamily.php` store these values using `set()`, not `setResult()` — so the condition always failed silently
- Changed to `has()`/`get()` to match how the data is actually stored by the form builder processes

**Impact:** Without this fix, accepting an application creates the student, parents, and family correctly, but the admissions account's `gibbonPersonID` is never updated — the applicant (parent) remains unlinked.

## Test plan
- [ ] Create a new admissions application with parent details
- [ ] Accept the application
- [ ] Verify `gibbonAdmissionsAccount.gibbonPersonID` is set to the parent's person ID
- [ ] Verify `gibbonAdmissionsAccount.gibbonFamilyID` is set to the new family ID

### DIFF
--- a/modules/Admissions/applications_manage_acceptProcess.php
+++ b/modules/Admissions/applications_manage_acceptProcess.php
@@ -94,14 +94,14 @@ if (isActionAccessible($guid, $connection2, '/modules/Admissions/applications_ma
 
     // Link the admissions account to new parent and family, if they were created during this acceptance process
     if ($formData->getStatus() == 'Accepted') {
-        if ($formData->hasResult('familyCreated') && $formData->hasResult('gibbonFamilyID') && empty($account['gibbonFamilyID'])) {
+        if ($formData->has('familyCreated') && $formData->has('gibbonFamilyID') && empty($account['gibbonFamilyID'])) {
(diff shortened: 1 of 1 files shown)


SKuipers (commented):
Hi @brianp-matrix, I appreciate you taking time to share your code as a PR. Are you able to help explain the issue you've encountered in greater detail? I've tried to reproduce it on my end and the script seems to be working as expected, with the gibbonPersonID and gibbonFamilyID fields being populated on acceptance. 

The `applications_manage_acceptProcess.php` script runs a method called `setReadOnly`, which ensures that all data captured from the acceptance process goes into Results and not the original Data, this way it does not overwrite data the applicant submitted with generated data. 

So, the `getResult()` method should be returning the correct data at that point in the code. If you're encountering an issue and not seeing the same thing on your end, let me know what steps I can take to reproduce the issue, and I'm happy to investigate further.

PR #2087, opened 2026-04-14
Add archive identifier fallback and tag fix

ZhangTao08:
When building per-person archive entries, try an identifier-based lookup if no recent archive is found (providing backward compatibility for archives linked by identifier/name). Rename the mapped variable for clarity and ensure the closure captures the outer $report. Also avoid rendering an empty status tag by checking that archive.status is non-empty before outputting the status badge.

### DIFF
--- a/modules/Reports/reports_generate_single.php
+++ b/modules/Reports/reports_generate_single.php
@@ -58,10 +58,24 @@ if (isActionAccessible($guid, $connection2, '/modules/Reports/reports_generate_s
 
     $context = ContextFactory::create($templateData['context']);
 
-    $ids = $context->getIdentifiers($pdo, $report['gibbonReportID'], $contextData, true);
-    $ids = array_map(function ($report) use ($gibbonReportID, &$reportArchiveEntryGateway) {
-        $report['archive'] = $reportArchiveEntryGateway->getRecentArchiveEntryByReport($gibbonReportID, 'Single', $report['gibbonPersonID'], 'Staff', true, true);
-        return $report;
+$ids = $context->getIdentifiers($pdo, $report['gibbonReportID'], $contextData, true);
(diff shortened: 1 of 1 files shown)


SKuipers (commented):
Hi @ZhangTao08, thank you for your pull request and code changes, which look pretty straight forward. 

Are you able to help explain the bug that you encountered related to this fix? I ask because the reports_generate_single.php page requires a $gibbonReportID to function (it generates an error message if it is missing), so its curious to see that a fallback would be needed for cases where there is no $gibbonReportID. I would be curious to hear more about the issue you've encountered.

Thanks!

ZhangTao08:
So the problem happened on Sayfol's Gibbon when we were generating report cards for students but some grade level report cards dont have 'last created' and 'status' and Go into the grade level it didnt have the option to view the report card for the students. And i replicated the error on our test instance.

SKuipers:
Thanks @ZhangTao08 for your response. I wonder, were they using the Upload Reports option to add reports with the same identifier as an existing report? Unless perhaps the report was deleted and a new one was created with the same name? Otherwise, could you share the steps you used to replicate the issue, since from what I can see, generated reports should always be attached to a report ID 🤔

ZhangTao08:
I am sure that we were just generating report cards by running thru the report card template, not by uploading reports. We customized a report card block and had some issue with pulling the right data, then we fixed it. We were about to bulk generate for each grade level but some grade levels worked with a last time stamp and status, but some didnt. And if we go into each student individual profile and check report there, the report card was generated with the time stamp and a status. It is weird~

SKuipers (approved):
HI @ZhangTao08 Sorry for the delay, it's been a busy time of the school year. I'm going to merge this, should there be an edge case that it helps with, however given that the gibbonReportID is a required value, its still unusual that this would have caused an error.

PR #2093, opened 2026-04-22
Attendance: add Class as a target option for Set Future Absence

ali-ichk:
**Description**
Added Class as a target option for Set Future Absence

### DIFF
--- a/modules/Attendance/attendance_future_byPerson.php
+++ b/modules/Attendance/attendance_future_byPerson.php
@@ -67,6 +67,7 @@ if (isActionAccessible($guid, $connection2, '/modules/Attendance/attendance_futu
     $target = $_GET['target'] ?? '';
     $gibbonActivityID = $_GET['gibbonActivityID'] ?? '';
     $gibbonGroupID = $_GET['gibbonGroupID'] ?? '';
+    $gibbonCourseClassID = $_GET['gibbonCourseClassID'] ?? '';
     $absenceType = $_GET['absenceType'] ?? 'full';
     $date = $_GET['date'] ?? '';
     $dateStart = $_GET['dateStart'] ?? '';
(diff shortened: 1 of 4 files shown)


modules/Attendance/attendance_future_byPerson.php
```diff
@@ -192,7 +198,7 @@
 
     // Get list of students for selected target
     if (!empty($target)) {
-        $targetID = $target == 'Activity' ? $gibbonActivityID : ($target == 'Messenger' ? $gibbonGroupID : $gibbonPersonIDList);
+        $targetID = $target == 'Activity' ? $gibbonActivityID : ($target == 'Messenger' ? $gibbonGroupID : ($target == 'Class' ? $gibbonCourseClassID : $gibbonPersonIDList));
```
SKuipers: This conditional chain is getting long enough, a `switch` would help keep the code clean, especially if more conditions are ever added in the future.

modules/Attendance/attendance_future_byPersonProcess.php
```diff
@@ -88,9 +89,9 @@
         $comment = $_POST['comment'] ?? '';
         $courseList = $_POST['courses'] ?? '';
 
-        if (in_array($urlParams['target'], ['Activity', 'Messenger'])) {
-            $foreignTable = $urlParams['target'];
-            $foreignTableID = $urlParams['target'] === 'Activity' ? $urlParams['gibbonActivityID'] : $urlParams['gibbonGroupID'];
+        if (in_array($urlParams['target'], ['Activity', 'Messenger', 'Class'])) {
+            $foreignTable = $urlParams['target'] === 'Activity' ? 'gibbonActivity' : ($urlParams['target'] === 'Messenger' ? 'gibbonGroup' : 'gibbonCourseClass');
+            $foreignTableID = $urlParams['target'] === 'Activity' ? $urlParams['gibbonActivityID'] : ($urlParams['target'] === 'Messenger' ? $urlParams['gibbonGroupID'] : $urlParams['gibbonCourseClassID']);
```
SKuipers: With more than two conditions, this would likely be best as separate code blocks to keep things clear, either a switch or if-else ladder.

SKuipers (changes requested):
Thanks for putting together this improvement! I've added a couple small notes to improve the code clarity, since it's now moving from two conditions to three (or more in the future).

SKuipers (approved):
Thanks for the improvements, looking cleaner and easier to read. I'll merge this in

PR #2095, opened 2026-04-22
Fix code-type custom fields becoming uneditable in course context

tiekubd:
**Summary**
Fixes an issue where code-type custom fields could become uneditable in course context.

**What changed**

* Refactored ACE editor setup into a single async initialization block
* Load `ext-language_tools` only when autocomplete is enabled
* Added a guard to prevent duplicate editor initialization on the same DOM node
* Synced the hidden textarea with the editor value immediately on load
* Updated HTMX cleanup to safely destroy the editor before page swaps

**Why**
In course context, repeated initialization or HTMX swap behavior could leave the code editor in a broken or uneditable state. This change makes initialization and teardown more reliable.

**Testing**

* Opened a code-type custom field in course context
* Confirmed the editor loads correctly
* Confirmed content changes sync back to the hidden textarea
* Confirmed duplicate editor initialization is prevented
* Confirmed editor teardown occurs cleanly before HTMX navigation

**Related discussion**
https://ask.gibbonedu.org/t/code-type-custom-fields-become-uneditable-in-course-context/15539/4

### DIFF
--- a/src/Forms/Input/CodeEditor.template.php
+++ b/src/Forms/Input/CodeEditor.template.php
@@ -3,32 +3,48 @@
 <div id="editor<?= $id; ?>" class="w-full" style="height: <?= $height; ?>px;"><?= htmlentities($text ?? '', ENT_QUOTES, 'UTF-8'); ?></div>
 
 <script type="text/javascript">
-    function setupEditor () {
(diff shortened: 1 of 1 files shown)


SKuipers (approved):
Thanks @tiekubd, this looks good, I'll merge it in now and test it out.

PR #2096, opened 2026-04-24
Students: Added a checkbox to the Medical Data Summary to exclude stu…

ali-ichk:
**Description**
Students: Added a checkbox to the Medical Data Summary to exclude students with no medical conditions


SKuipers (approved):
Thanks Ali for the improvement, I'm going to merge this now and sneak it into production so we can start using it 👍

PR #2098, opened 2026-04-27
Markbook: QUICK fix to the visual bug in markbook column

ali-ichk:
**Description**
Quick UI Bug fix


SKuipers (approved):
Perfect, thanks for the quick fix.

PR #2099, opened 2026-04-27
System: Fixed the bug to display the correct list of students

ali-ichk:
**Description**
Fixed the bug to display the correct list of students in Trip Overview


SKuipers (approved):
Thanks for another quick fix

PR #2100, opened 2026-04-28
Messenger: creator of targeted message can now view it in Message Wall

ali-ichk:
**Description**
Message Wall: Creator of targeted message can view their own message on Gibbon Wall


SKuipers (approved):
Thanks Ali for the improvement.

PR #2101, opened 2026-04-28
Timetable Admin: fixed error preventing the copy of courses to next y…

ali-ichk:
**Description**
Timetable Admin: fixed error preventing the copy of courses to next year. The issue was shared here: https://ask.gibbonedu.org/t/file-not-found-error-on-course-copy-to-next-year-timetable-admin-v30/15557.

### DIFF
--- a/modules/Timetable Admin/course_manage_copyProcess.php
+++ b/modules/Timetable Admin/course_manage_copyProcess.php
@@ -33,14 +33,15 @@ $URL = $session->get('absoluteURL')."/index.php?q=/modules/Timetable Admin/cours
 if (isActionAccessible($guid, $connection2, '/modules/Timetable Admin/course_manage.php') == false) {
     $URL .= '&return=error0';
     header("Location: {$URL}");
+    exit;
 } else {
-    //Proceed!
-    
+    // Proceed!
     // Check if school years specified (current and next)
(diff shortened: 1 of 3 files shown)


SKuipers (changes requested):
Hi Ali, are you able to test the Copy All to Next Year button before applying the fix? It is working correctly for me in v30 and v31. The proposed change appears to break the button by removing the `->directLink()`, which would cause it to be treated as an interface page and not a process script.

ali-ichk:
Hi Sandra, thank you for reviewing the changes and catching the bug. I realised I introduced another bug while fix the earlier issue. I have resolved and tested it now and Copy All to Next Year button works with the directLink(). Thanks

modules/Timetable Admin/course_manage_copyProcess.php
```diff
@@ -22,7 +22,7 @@
 use Gibbon\Domain\Timetable\CourseGateway;
 use Gibbon\Domain\Timetable\CourseClassGateway;
 
-require_once '../../gibbon.php';
+require_once dirname(__DIR__, 2).'/gibbon.php';
```
SKuipers:
```suggestion
require_once '../../gibbon.php';
```

SKuipers (approved):

PR #2103, opened 2026-04-30
System Admin: implemented the new file system and the new fileHandler class

ali-ichk:
**Description**
Implemented the fileHandler class in the bulk upload process page (such as uploading multiple user profile pictures)


SKuipers (approved):
Thanks, looks good.

PR #2107, opened 2026-05-12
Behaviour: fixed the issue of student not getting included in a multiple incident behaviour record

ali-ichk:
**Description**
Behaviour: fixed the issue of student not getting included in a multi behaviour record


SKuipers (approved):
Thanks for the quick fix!

PR #2108, opened 2026-05-12
Behaviour  improving the script for the add multi process page

ali-ichk:
**Description**
Behaviour: Improved the multiadd process page code to reduce computation overhead and add further error checking and remove code being repeated in a loop (functionality remains the same)


SKuipers (approved):
Wonderful, thanks for the code improvements!

PR #2117, opened 2026-06-16
Individual Needs: Improved the code so that the file upload is consistent with other files

ali-ichk:


modules/Individual Needs/in_supportPlan_add.php
```diff
@@ -87,13 +87,13 @@
 
     // File upload
     $row = $form->addRow()->addClass('rowFile');
-        $row->addLabel('file', __('PDF File'));
-        $row->addFileUpload('file')->accepts('.pdf');
+        $row->addLabel('file', __('Upload File'));
```
SKuipers: Thanks @ali-ichk Instead of `Upload File` and `Upload Link`, can we update this to just use `File` and `Link`, as these are existing strings and will make more sense when used on both the add and edit page.

SKuipers (approved):

PR #2119, opened 2026-06-17
Messenger: Fixed the issue of Non-Parameterised input

ali-ichk:
**Description**
Fixed the issue of Non-Parameterised input

### DIFF
--- a/modules/Messenger/src/MessageTargets.php
+++ b/modules/Messenger/src/MessageTargets.php
@@ -1958,35 +1958,13 @@ class MessageTargets
                         if ($sms=="Y" AND $countryCode!="") {
                             if ($parents=="Y") {
                             try { //Get the familyIDs for each student logged
-                                $validStudentIDs = [];
-                                foreach ($selectedStudents as $studentID) {
-                                    if (is_numeric($studentID)) {
-                                        $validStudentIDs[] = $studentID;
-                                    }
-                                }
-
-                                if (!empty($validStudentIDs)) {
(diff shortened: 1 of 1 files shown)


SKuipers (changes requested):
@ali-ichk While the variables have been sanitized, they have not necessarily been added as parameters. For best practice, they need to be added to the data array, and then used as a parameter in the SQL eg ` IN (:studentIDList) `

modules/Messenger/src/MessageTargets.php
```diff
@@ -1953,17 +1953,40 @@ public function createMessageRecipientsFromTargets($gibbonMessengerID, $data, &$
...
+                                    foreach ($validStudentIDs as $studentID) {
+                                        $paramName = 'student' . $count;
+                                        $dataFamily[$paramName] = $studentID;
+                                        $placeholders[] = ':' . $paramName;
+                                        $count++;
+                                    }
+
+                                    $inClause = implode(',', $placeholders);
+                            
+                                    $sqlFamily="SELECT DISTINCT gibbonFamilyID FROM gibbonFamilyChild WHERE gibbonPersonID IN ($inClause)" ;
```
SKuipers: @ali-ichk You're making progress, but the goal is to ensure there are no variables inserted into the SQL statement, so the `$inClause` itself should be a parameter, and the imploded string is passed into the data array. Check your code complexity, there should be a bit cleaner way to implement this, thanks.

SKuipers (approved):

PR #2121, opened 2026-06-18
Admissions: added new options for how houses are automatically assigned when a student application is accepted

ali-ichk:
**Description**
 Improved the logic to automatically assign a house based on siblings or parents.
 When a new student is accepted, the system checks if their parents are staff member and assigns the same house as them. If not, the system checks if they have siblings and assigns the student to the same house as their siblings. If not, the system automatically assigns them to a house based on old logic.


SKuipers (changes requested):
Hi @ali-ichk, thanks for making these improvements. I notice that there are 5 or more separate database queries for this action (depending on the number of family adults), which adds a fair bit of nested logic. I wonder if the same result could be achieved with a single new query in the `HouseGateway` called `selectExistingHouseByFamilyID`, and in that query it could get the house data for a given family, ordered by staff parents first, and siblings second.

PR #2133, opened 2026-07-16
Reports: Reporting Cycle edit saves milestones field as literal "Arra…

ali-ichk:
**Description**
Fixed the issue where Reporting Cycle edit was saving milestones field as literal "Array" string instead of JSON. Issue was reported here: https://ask.gibbonedu.org/t/bug-reporting-cycle-edit-saves-milestones-field-as-literal-array-string-instead-of-json/15590.


modules/Reports/reporting_cycles_manage_editProcess.php
```diff
@@ -51,21 +52,32 @@
         'cycleNumber'           => $_POST['cycleNumber'] ?? '1',
         'cycleTotal'            => $_POST['cycleTotal'] ?? '1',
         'notes'                 => $_POST['notes'] ?? '',
-        'milestones'            => $_POST['milestones'] ?? [],
+        'milestones'            => $currentReportingCycle['milestones'] ?? null,
```
SKuipers, line 55: In testing the fix, this line can prevent the ability to remove milestones. If someone has milestones, and then deletes them, it becomes a null array, but doing so forces it back to the original value.

modules/Reports/reporting_cycles_manage_editProcess.php
```diff
@@ -51,21 +52,32 @@
...
+
+            $orderKeys = array_keys($_POST['order'] ?? []);
+            if (!empty($orderKeys) && count($orderKeys) === count($milestones)) {
+                $milestones = array_combine($orderKeys, array_values($milestones));
+                ksort($milestones);
+            }
+
+            $data['milestones'] = json_encode($milestones);
+        } else {
+            $data['milestones'] = null;
```
SKuipers, line 79: Upon further investigation, it may be better to store empty arrays as an empty json value rather than null, as passing null to json_decode is depricated, enforcing a json value prevents the need for null checking this value every single time.

PR #2135, opened 2026-07-16
System: Added next year and previous year buttons in the Year Switcher in the SIdebar

ali-ichk:
**Description**
System: Added next year and previous year buttons in the Year Switcher in the SIdebar


SKuipers (approved):
Thanks Ali, this looks like a nice usability improvement

PR #2142, opened 2026-08-04
System: added a new sanitizePersonName helper method in Validator class

ali-ichk:
**Description**
System: added a new sanitizePersonName helper method in Validator class to sanitize user input names
As defense in depth, applying output/parameter validation (e.g. rejecting or escaping single quotes) on point of write ( publicRegistrationProcess.php , firstName at the user_manage_addProcess.php , user_manage_editProcess.php )would reduce the number of downstream sinks this class of stored value can reach.


SKuipers (approved):
Thanks Ali, I'll merge this and make the changes we discussed post-merge.

PR #2143, opened 2026-08-04
Messenger   resolved the bug of non paramaterised input in messenger gateway

ali-ichk:
**Description**
Messenger: fixed the issue of non-parameterized input in Messenger Gateway

### DIFF
--- a/CHANGELOG.txt
+++ b/CHANGELOG.txt
@@ -81,7 +81,8 @@ v31.0.00
         Behaviour: fixed the issue of student not getting included in a multiple behaviour record
         System: fixed handling of array values in localeFriendlySort method in DatabaseFormFactory
         System: renamed ca_CA locale to ca_ES and updated locale files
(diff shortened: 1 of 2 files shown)


sabuhimammadov0 (commented):
Thanks for the fix — reviewed at the code level and this resolves the second-order SQL injection.

✅ The vulnerable concatenation is replaced, not duplicated (2 insertions(+), 1 deletion(-)), so no vulnerable UNION block remains.
 ✅ firstName is now bound as a named parameter (:attendanceFirstName via $dataPosts['attendanceFirstName']), and $dataPosts is passed to $resultPosts->execute() — so a ' in firstName is now literal data, not SQL. This matches the safe pattern already used for :attendanceType2 and the other UNION blocks.
✅ No other raw variable is concatenated into the query in getMessages(); $dateWhere is a static string using a bound :date.

SKuipers (approved):
Thanks @ali-ichk and @sabuhimammadov0

PR #2145, opened 2026-08-17
School Admin: fixed the issue of Form Group Location not being displayed in edit form group

ali-ichk:
**Description**
School Admin: fixed the issue of Form Group Location not being displayed in edit page


SKuipers (changes requested):
Thanks Ali for the quick fix. For consistency with the codebase, we tend to handle all string padding in PHP rather than MySQL, as this keeps the query cleaner and forwards-compatible with padding changes in upcoming MySQL versions.

SKuipers (approved):

PR #2149, opened 2026-09-02
System: Fixed the issue of search field not including school year parameter

ali-ichk:
**Description**
Fixed the issue of search field not including school year parameter in core modules


SKuipers (approved):
Thanks Ali for the fixes, looks good

PR #2150, opened 2026-09-03
Update require_once for gibbon.php to  __DIR__ 

fvlasie:
Description
Update every require_once / include of gibbon.php to resolve the file from __DIR__ instead of a cwd-relative path (./gibbon.php, ../gibbon.php, ../../gibbon.php). Depth is preserved (root scripts use __DIR__ . '/gibbon.php'; module scripts use __DIR__ . '/../../gibbon.php'). No other logic is changed.

Motivation and Context
Relative includes are resolved against the process working directory, not the script location. That fails when PHP is invoked from another directory (CLI, cron, some web/php-fpm setups, or wrapping scripts). __DIR__ always points at the file on disk, so bootstrap still loads in those cases. This PR fixes the fragility of require_once and the panic attacks at 2am when Gibbons escape and nobody knows why. I encountered escaped Gibbons in my prod environment when using core functions such as copy classes to next year.

How Has This Been Tested?
Change is mechanical (661 one-line include path updates).
Locally in MAMP from a clean install.
In production (the reason for this PR in the first place).


SKuipers (approved):
Thank you @fvlasie for your PR. While normally 663 files changed would be a bit too overwhelming to review 😅  I appreciate that you've made one change across the whole codebase, which helps improves the consistency of the code. This wide-rather-than-deep refactoring strategy is something we like to use too, and your contribution is much appreciated.

PR #2151, opened 2026-09-04
Timetable Admin: updated the course selector in Course Enrolment to use multi selector

ali-ichk:
**Description**
updated the course selector in Course Enrolment to use multi selector


ali-ichk:
Hi @SKuipers
The test script needs to be updated to support the multi selector pattern (the destination array will be empty at first which breaks the current test code). Thanks.

PR #2154, opened 2026-09-07
Student Alerts: Fixed the Gateway name to use the correct Gateway

ali-ichk:
**Description**
Student Alerts: Fixed the Gateway name to use the correct Gateway

### DIFF
--- a/modules/Student Alerts/studentAlerts_addProcess.php
+++ b/modules/Student Alerts/studentAlerts_addProcess.php
@@ -19,6 +19,8 @@ along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 use Gibbon\Comms\NotificationEvent;
 use Gibbon\Data\Validator;
+use Gibbon\Http\Url;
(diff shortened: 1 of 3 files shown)


SKuipers (commented):
Thanks Ali, can I check if you have an extension that is re-ordering the namespace includes? For consistency, they tend to be ordered by system includes then module includes, but your PR seems to be alphabetizing them.

ali-ichk:
Hi @SKuipers thanks for catching that. It is probably the auto behaviour of VS Code extension Intelephense, when I simply right-click the class name and choose 'import class". I'll note it and keep changes accordingly.

SKuipers:
> Hi @SKuipers thanks for catching that. It is probably the auto behaviour of VS Code extension Intelephense, when I simply right-click the class name and choose 'import class". I'll note it and keep changes accordingly.

Thanks Ali, check the settings for Intelephense, I believe you can turn the auto-sort option off.

PR #2161, opened 2026-09-21
Students : First Aid - added a delete option to delete a record.

ali-ichk:
**Description**
Students : First Aid - added a delete option to delete a record compatible with the new file management system


SKuipers (approved):
Thanks Ali for adding the action and the test coverage.

